PR #19, opened 2022-07-26
Feature/sophia

jncots:
Update of sophia.py module to include it in hierachy of the base classes MCEvent and MCRun.


impy/definitions.py
```diff
@@ -104,9 +104,16 @@
...
+    ],  
+    [
+        'SOPHIA20', 'SOPHIA', '2.0', -1, 'sophia', sophia.SophiaRun,
+        sophia.SophiaEvent, 'laboratory'
     ]
 ]
 
+# 'tag', 'name', 'version', 'crmc_id', 'library_name', 'RunClass',
+#     'EventClass', 'output_frame'
+
```
afedynitch, line 266: Is this a leftover comment? Seems useful. Maybe, it's better to keep it above the block

impy/models/sophia.py
```diff
@@ -5,158 +5,191 @@
...
+
+class SophiaEvent(MCEvent):
+    """Wrapper class around Sophia code"""
+
+    def __init__(self, lib, event_kinematics, event_frame):   
+
+        lib.toevt() # prepare hepevt common block     
+        np = lib.hepevt.nhep # number of particles in event
+        pem_arr = lib.s_plist.p[:np,:].T # array of particle momenta
+        vt_arr = lib.hepevt.vhep[:, 0:np] # array of verticies (x, y, z, t) - all zeros for sophia
```
afedynitch: Comment seems to be over the line length limit

impy/models/sophia.py
```diff
@@ -5,158 +5,191 @@
...
+    
+    @property
+    def decayed_parent(self):
+        """Returns the array of indices of the decayed parent particles"""
+        MCEvent.parents(self)
+        return self.lib.schg.decpar[0:self.npart]
+
+    @property
+    def parents(self):
+        """In SOPHIA parents are difficult to obtain. This function returns 0."""
```
afedynitch, line 66: Isn't this working? Then should be reflected in the docstring. Also for the children below.

impy/models/sophia.py
```diff
@@ -5,158 +5,191 @@
...
+        MCEvent.children(self)
+        return self.lib.hepevt.jdahep[0:self.npart]
+
+class SophiaRun(MCRun):
+    """Implements all abstract attributes of MCRun for the 
+    Sophia event generator.
+    """
+
+    def sigma_inel(self, *args, **kwargs):
+        """Inelastic cross section according to current
```
afedynitch, line 83: You may want to add that it will return the total cross section (for now).

impy/models/sophia.py
```diff
@@ -5,158 +5,191 @@
...
+        info(5, 'Setting event kinematics.')
+        info(10, event_kinematics)
+        k = event_kinematics
+        
+        if k.p1pdg != 22:
+            info(0, "The first particle must be a photon, but particle pdg = ", k.p1pdg)
+            raise Exception('Input error')
+        
+        if k.p2pdg not in [2212, 2112]:
+            info(0, "The second particle must be a proton or neutron, but particle pdg = ", k.p2pdg)
```
afedynitch: Please, run black

impy/models/sophia.py
```diff
@@ -5,158 +5,191 @@
...
+            raise Exception('Input error')
+        
+        
+        self.nucleon_code_number = self.lib.icon_pdg_sib(k.p2pdg)
+        self.energy_of_nucleon = k.pmass2
+        self.energy_of_photon = k.elab 
+        # Here we consider laboratory frame where photon moves along z axis
+        # and nucleon is at rest. The angle is counted from z axis.
+        # However, because of the definitions in "eventgen" subroutine of
+        # SOPHIA code (line "P_gam(3) = -EPS*COS(theta*pi/180.D0)")
```
afedynitch, line 138: Please, don't forget to add this as an issue after PR merged.

impy/models/sophia.py
```diff
@@ -5,158 +5,191 @@
...
+            info(5, 'pdgid/sid = {0}/{1} allowed to decay'.format(pdgid, sid))
+            idb[sid - 1] = np.abs(idb[sid - 1])
+
+    def generate_event(self):
+        # Generate event (the final particles and their parameters) 
+        # by underlying Fortran library 
+        self.lib.eventgen(self.nucleon_code_number, 
+                self.energy_of_nucleon, 
+                self.energy_of_photon, 
+                self.angle_between_nucleon_and_photon, 0)
```
afedynitch: The last 0 is actually a return parameter for which cross section has been selected. There should be an Cf2py intent(out) in eventget. You can pick up the return parameter as a class variable integer and make a property function that returns the string of what the interaction was, resonance, direct, multi-string etc. It certainly should not be an call argument.

src/sophia/eventgen.f
```diff
@@ -263,6 +268,8 @@ subroutine eventgen(L0,E0,eps,theta,Imode)
        endif
 
 c... consider only stable particles:
+       if (REMDEC .eq. 0) then !anpros 2022/07/22: Leave decayed
```
afedynitch, line 271: Don't use inline comments in F77 code. There make random problems with some f2py versions: F77 = comment character, free form via !. No need to break the standard if not crucially needed.

src/sophia/eventgen.f
```diff
@@ -286,6 +293,7 @@ subroutine eventgen(L0,E0,eps,theta,Imode)
          enddo
         endif
         NP = istable       
+       end if  ! anpros 2022/07/22  REMDEC condition end
```
afedynitch, line 296: There is a file called impy_config.yaml. There should be a definition of SOPHIA. The behavior of removing the decayed particles should be controlled from there. It's something that doesn't or shouldn't change in runtime. You can put the default to keep the particles, since most other models do it. The performance penalty for something fast as SOPHIA is totally negligible.

src/sophia/eventgen.f
```diff
@@ -44,6 +44,11 @@ subroutine eventgen(L0,E0,eps,theta,Imode)
       IMPLICIT DOUBLE PRECISION (A-H,O-Z)
       SAVE
 
+C** anpros 2022/07/22 added block begins
+       INTEGER REMDEC
+       DATA REMDEC / 0 /     !default value means removing 
+       COMMON /EG_IO/ REMDEC !controls removing decayed particles
+C** anpros 2022/07/22 added block ends
```
afedynitch, line 51: Missing the f2py intent(out) imode

src/sophia/utility.f
```diff
@@ -0,0 +1,69 @@
+      SUBROUTINE TOEVT()
+C************************************************************************
+C
```
afedynitch, line 3: The file should be called impy_sophia.f in line with what is there for all other generators EXCEPT SIBYLL :) It's an historical oversight and need to be corrected there, too.

src/sophia/utility.f
```diff
@@ -0,0 +1,69 @@
...
+C     standard HEPEVT common block format, 
+C     and S_PLIST1 and S_CHP to SCHG common block containing
+C     charge and parent decayed particle information
+C     
+C************************************************************************
+      IMPLICIT DOUBLE PRECISION (A-H,O-Z)
+      IMPLICIT INTEGER (I-N)
+
+      SAVE
+      INTEGER PDGCOD(49)
```
afedynitch, line 14: Where is this table from?

src/sophia/utility.f
```diff
@@ -0,0 +1,69 @@
...
+      PARAMETER (NMXHEP = 2000)
+      PARAMETER (NMXHP2 = NMXHEP * 2, NMXHP4 = NMXHEP * 4, 
+     & NMXHP5 = NMXHEP * 5)
+      COMMON /HEPEVT/ NEVHEP, NHEP, ISTHEP(NMXHEP), IDHEP(NMXHEP),
+     & JMOHEP(2, NMXHEP), JDAHEP(2, NMXHEP), 
+     & PHEP(5, NMXHEP), VHEP(4, NMXHEP)
+
+C     Initialization of HEVEVT variables
+      DATA NEVHEP / 0 /
+      DATA JMOHEP/ NMXHP2 * 0.0/, JDAHEP / NMXHP2 * 0.0 /
```
afedynitch, line 36: Usually not needed in the context since the blocks are never read out beyond nevhep. But it may trigger gfortran with -pedantic flags.. so.

src/sophia/utility.f
```diff
@@ -0,0 +1,69 @@
...
+
+      INTEGER CID  ! id of the current (in the loop) particle
+C     number added to decayed particle:
+      INTEGER DNUM
+      PARAMETER (DNUM = 10000)
+       
+      NEVHEP = NEVHEP + 1
+      NHEP = NP
+C     We do not transpose P, because we don't use PHEP         
+C      PHEP = TRANSPOSE(P)
```
afedynitch, line 53: Not using PHEP is a bit non standard for the rest of impy. Not sure I recommend really doing this because the HEPEVT is a standard common block and could be used as such by someone else with the corresponding interface.

How much penalty in ns,us, ms does it cost to fill PHEP?

In the end this must work with the HepMC3 interface, so whatever it requires. Also, consider that I expect most people to use python to analyse the output. So making numpy computations work on non-contiguous arrays might kill the benefit you gain here on the fortran side. However, I never tested this.

src/sophia/utility.f
```diff
@@ -0,0 +1,69 @@
...
+        
+      DO I = 1, NHEP
+        CID = LLIST(I)
+        IF (ABS(CID) .GT. DNUM) THEN
+            ISTHEP(I) = 2
+            CID = CID - ISIGN(DNUM, CID)
+        ELSE 
+            ISTHEP(I) = 1
+        END IF     
+        IDHEP(I) = ISIGN(1, CID) * PDGCOD(ABS(CID))
```
afedynitch, line 63: What does it do for neutral particles or K0 like 310, 130?

afedynitch (changes requested):
Please have a look at the comments and resubmit

impy/definitions.py
jncots: Yes, this is leftover from working with the code and it should be removed as it copy/paste from the code above

impy/models/sophia.py
jncots: Fixed

jncots: The comment is moved to the line above

impy/models/sophia.py
jncots: It returns an array of correct shape with all zeros. I added this info to the docstring.

impy/models/sophia.py
jncots: The docstring now reflects this

impy/models/sophia.py
jncots: Done

impy/models/sophia.py
jncots: I will add it

src/sophia/eventgen.f
jncots: The dictionary to sophia is added to impy_config.yaml. It mostly copies sibyll with relevant changes. However, the meaning of some of the entries is not clear to me, so this should be checked. keep_decayed_particles : True is added as default, to control keeping of decayed particles. Actually, in this case there is no performance penalty, because keeping particles do not require any operations as opposed to the removing.

src/sophia/eventgen.f
jncots: Added

src/sophia/utility.f
jncots: The file is renamed utility.f -> impy_sophia.f

src/sophia/utility.f
jncots: It is the table from FUNCTION ICON_PDG_SIB(ID) in eventgen.f, which is used to convert PDG particle codes to SIBYLL particle codes

src/sophia/utility.f
jncots: Here I am not sure that I understand a suggestion. Could you elaborate what is not needed?

src/sophia/utility.f
jncots: The operation PHEP = TRANSPOSE(P) adds approximately 10% to the execution time of bare event generation. Probably, it is still not critical, considering that subsequent operations in python will eat up  much more. I commented it out because it turns out, that the operation like
```python
px, py, pz, en, m = lib.s_plist.p[:np, :].T
```
produces contiguous arrays px, py, pz, en, m. And if we used transposed PHEP matrix, px, py, pz, en, m will not be contiguous in memory. This is because  lib.s_plist.p(2000, 5) in the memory goes like p(1, 1), p(2, 1), p(3, 1) and f2py do not change layout in memory. One can verify that px is contiguous by looking at px.__array_interface__ dictionary. 

So from this point of view we don't need PHEP at all. However, from point of view of interfaces, probably PHEP should be filled out using PHEP = TRANSPOSE(P) after all.

src/sophia/utility.f
jncots: For neutral particles sophia ID is positive, and we have effectively IDHEP(I) = PDGCOD(CID)

src/sophia/utility.f
afedynitch: As I mentioned, there is no need to transpose the entire array. NEVHEP is usually between 2 and 100. So just a for loop is enough

afedynitch (approved):
Looks fine to me

afedynitch:
You should be able to merge now

Issue #8, opened 2021-06-24
Make event_kinematics a property of a generator


HDembinski:
See #3 

Instead of calling `generator.init_generator(event_kin)` allow users to query and assign to a property with the name `event_kinematics`. Setting this internally calls `init_generator`, querying it returns the current event kinematics.

Issue #7, opened 2021-06-24
Simplify for loop over events from generator


HDembinski:
See #3 

This is a follow-up on #6 

* `for event in generator.event_generator(event_kinematics, nevents):` should be replaced by `for event in generator(nevents):`

Issue #6, opened 2021-06-24
Make models importable as classes


HDembinski:
See #3 

* allow `from impy.models import sibyll_23d` etc
* remove or hide `make_generator_instance`
* ctor of a model class gets the event_kinematics: `generator = sibyll_23d(event_kin)` and is then fully initialized and ready to draw events from
* since the models are singletons, follow https://stackoverflow.com/questions/6760685/creating-a-singleton-in-python

Issue #3, opened 2021-06-24
Design proposals


HDembinski:
One of the fundamental principles in good software design is to use the most basic language features to implement some feature. In other words, be minimalist when designing the interface. Always ask yourself: is there a simpler more basic way of providing this functionality? Another principle is: "Make interfaces easy to use correctly and hard or impossible to use incorrectly." This is obvious, but you have to really ask yourself this question every time you design something. How can I make it so that it is easy to use correctly and hard or impossible to use incorrectly?

Let's discuss the example code from that perspective.

```py
from impy.definitions import *
from impy.constants import *
from impy.kinematics import EventKinematics
from impy import impy_config

# Define the parameters of the collisions
event_kinematics = EventKinematics(
    ecm=13 * TeV, p1pdg=2212, p2pdg=2212)

# Create an instance of an event generator by passing
# the model name as a string
generator = make_generator_instance(
    interaction_model_by_tag['SIBYLL2.3D'])

# Initialize it
generator.init_generator(event_kinematics)

# Number of events to generate
nevents = 100

for event in generator.event_generator(event_kinematics, nevents):
    event.filter_final_state_charged()
    # do something with event.p_ids, event.eta, event.en, event.pt, etc.
    # these variables are numpy arrays, that can be histogrammed or counted like
    average_pt += 1/float(nevents)*np.mean(event.pt[np.abs(event.p_ids) == 211])

print('Average pT for charged pions {0:4.3f}'.format(average_pt))
```

* `EventKinematics` seems ok, it is probably useful to bundle this in a class. Providing units like `TeV` is good. Ideally, those are not just numbers but some class with `__mul__` and `__rmul__` operators defined, so that the `EventKinematics` object can check that the user actually supplied the unit and give an error when a plain number is passed to it. This is a bit fancy, so has low priority, but it would be the right way to implement this (make it impossible to use it incorrectly).

[...]

PR #20, opened 2022-08-03
Refactor/api changes

jncots:
According to the discussions in issues #3, #6, #7 and #8 the API is changed to:

```python
# Import is fixed, so autocompletion shows available submodules
import impy

ekin = impy.kinematics.EventKinematics(...)
# Event generator constructor (issue #6)
generator = impy.models.Sophia20(ekin)

# Callable generator (issue #7)
for event in generator(10000):
    # Chaining of filters
    event.filter_final().filter_charged()
    # One can "undo" last filter
    event.remove_last_filter()
    # Or completly remove all applied filters
    event.remove_all_filters()

    # One can get number of generated events (10000 in this case)
    print(generator.nevents)
    # generator.norm = 1/nevents, for example
    protons += generator.norm/widths*np.histogram(...)

    # One can change and get event_kinematics property (issue #8)
    ekin = generator.event_kinematics
    generator.event_kinematics = impy.kinematics.EventKinematics(...)
```

### DIFF
--- a/impy/common.py
+++ b/impy/common.py
@@ -335,6 +335,11 @@ class MCRun(six.with_metaclass(ABCMeta)):
 
         # FORTRAN LUN that keeps logfile handle
         self.output_lun = None
+        
+        # Number of events to generate
+        self.nevents = None
+        # Normalization 1/nevents for histograms
+        self.norm = None
 
     def __enter__(self):
         """TEMP: It would be good to actually use the with construct to
@@ -552,6 +557,8 @@ class MCRun(six.with_metaclass(ABCMeta)):
         # Initialize counters to prevent infinite loops in rejections
         ntrials = 0
         nremaining = nevents
+        self.nevents = nevents
+        self.norm = 1./float(nevents)
         while nremaining > 0:
             if self.generate_event() == 0:
                 yield self._event_class(self.lib, self._curr_event_kin,
--- a/impy/common.py
+++ b/impy/common.py
@@ -18,6 +18,28 @@ from impy import impy_config
 from impy.util import info
 
 
+class EventData:
+    """Data structure to keep views of filtered data"""
+    def __init__(self, npart, p_ids, status, charge,
+                 px, py, pz, en, m, vx, vy, vz, vt,
+                 pem_arr, vt_arr):
+        self.npart = npart
+        self.p_ids = p_ids
+        self.status = status
+        self.charge = charge
+        self.px = px
+        self.py = py
+        self.pz = pz
+        self.en = en
+        self.m = m
+        self.vx = vx
+        self.vy = vy
+        self.vz = vz
+        self.vt = vt
+        self.pem_arr = pem_arr
+        self.vt_arr = vt_arr
+    
+    
 class MCEvent(object, six.with_metaclass(ABCMeta)):
     """The basis of interaction between user and all the event generators.
 
@@ -47,7 +69,9 @@ class MCEvent(object, six.with_metaclass(ABCMeta)):
         vt (np.array)      : temporal order of vertex in ps, fs?
     """
     __sliced_params__ = [
-        'p_ids', 'status', 'px', 'py', 'pz', 'en', 'm', 'vx', 'vy', 'vz', 'vt',
+        'p_ids', 'status', 'charge',
+        'px', 'py', 'pz', 'en', 'm',
+        'vx', 'vy', 'vz', 'vt',
         'pem_arr', 'vt_arr'
     ]
 
@@ -81,18 +105,72 @@ class MCEvent(object, six.with_metaclass(ABCMeta)):
         if impy_config['pre_slice']:
             info(10, 'Pre-slice enabled.')
             self.selection = slice(None, self.npart)
+            self.charge = self._charge_init
             self._apply_slicing()
         else: 
             info(10, 'Pre-slice disabled.')
             self.selection = slice(None, None)
+            self.charge = self._charge_init
         
         # The default slice only cuts limits the view to the array to
         # to the current number of entries
         self._is_filtered = False
+        self._views_cache = []
 
         # Apply boosts into frame required by user
         self.kin.apply_boost(self, event_frame, impy_config["user_frame"])
         self.event_frame = impy_config["user_frame"]
+
+    
+    def filter_final(self):
+        self.selection = (self.status == 1)
+        self._cache_current_view()
+        self._apply_slicing()
+        return self
+      
+    def filter_charged(self):
+        self.selection = (self.charge != 0)
+        self._cache_current_view()
+        self._apply_slicing()
+        return self
+    
+    def _cache_current_view(self):
+        """Cache the current view of data"""
+        self._views_cache.append(EventData(
+            self._npart, self.p_ids, self.status, self.charge,
+            self.px, self.py, self.pz, self.en, self.m,
+            self.vx, self.vy, self.vz, self.vt,
+            self.pem_arr, self.vt_arr
+        ))
+    
+    def remove_last_filter(self):
+        """Remove the last applied filter"""
+        # if list empty then return
+        if not self._views_cache:
+            return
+        last = self._views_cache.pop()
+        self.npart = last.npart
+        self.p_ids = last.p_ids
+        self.status = last.status
+        self.charge = last.charge
+        self.px = last.px
+        self.py = last.py
+        self.pz = last.pz
+        self.en = last.en
+        self.m = last.m
+        self.vx = last.vx
+        self.vy = last.vy
+        self.vz = last.vz
+        self.vt = last.vt
+        self.pem_arr = last.pem_arr
+        self.vt_arr = last.vt_arr
+     
+    def remove_all_filters(self):
+        """Remove all filters (return to original data)"""
+        del self._views_cache[1:]
+        self.remove_last_filter()
+        self._is_filtered = False
+    
 
     def _apply_slicing(self):
         """Slices/copies the all varaibles according to filter criteria"""
@@ -126,7 +204,7 @@ class MCEvent(object, six.with_metaclass(ABCMeta)):
         pass
 
     @abstractproperty
-    def charge(self):
+    def _charge_init(self):
         """Electrical charge"""
         pass
 
--- a/impy/common.py
+++ b/impy/common.py
@@ -36,8 +36,8 @@ class EventData:
         self.vy = vy
         self.vz = vz
         self.vt = vt
-        self.pem_arr = pem_arr
-        self.vt_arr = vt_arr
+        self._pem_arr = pem_arr
+        self._vt_arr = vt_arr
     
     
 class MCEvent(object, six.with_metaclass(ABCMeta)):
@@ -72,7 +72,7 @@ class MCEvent(object, six.with_metaclass(ABCMeta)):
         'p_ids', 'status', 'charge',
         'px', 'py', 'pz', 'en', 'm',
         'vx', 'vy', 'vz', 'vt',
-        'pem_arr', 'vt_arr'
+        '_pem_arr', '_vt_arr'
     ]
 
     def __init__(self, lib, event_kinematics, event_frame, nevent, npart,
@@ -98,8 +98,8 @@ class MCEvent(object, six.with_metaclass(ABCMeta)):
         self.vt = vt
 
         # Full arrays of kinematical vectors
-        self.pem_arr = pem_arr  # (px, py, pz, E, m)
-        self.vt_arr = vt_arr  # (vx, vy, vz, t)
+        self._pem_arr = pem_arr  # (px, py, pz, E, m)
+        self._vt_arr = vt_arr  # (vx, vy, vz, t)
 
         # Initialize current selection to all entries up to npart
         if impy_config['pre_slice']:
@@ -140,7 +140,7 @@ class MCEvent(object, six.with_metaclass(ABCMeta)):
             self._npart, self.p_ids, self.status, self.charge,
             self.px, self.py, self.pz, self.en, self.m,
             self.vx, self.vy, self.vz, self.vt,
-            self.pem_arr, self.vt_arr
+            self._pem_arr, self._vt_arr
         ))
     
     def remove_last_filter(self):
@@ -162,8 +162,9 @@ class MCEvent(object, six.with_metaclass(ABCMeta)):
         self.vy = last.vy
         self.vz = last.vz
         self.vt = last.vt
-        self.pem_arr = last.pem_arr
-        self.vt_arr = last.vt_arr
+        self._pem_arr = last._pem_arr
+        self._vt_arr = last._vt_arr
+        
      
     def remove_all_filters(self):
         """Remove all filters (return to original data)"""
@@ -414,10 +415,8 @@ class MCRun(six.with_metaclass(ABCMeta)):
         # FORTRAN LUN that keeps logfile handle
         self.output_lun = None
         
-        # Number of events to generate
-        self.nevents = None
-        # Normalization 1/nevents for histograms
-        self.norm = None
+        # Number of generated events so far
+        self.nevents = 0
 
     def __enter__(self):
         """TEMP: It would be good to actually use the with construct to
@@ -456,7 +455,7 @@ class MCRun(six.with_metaclass(ABCMeta)):
         
         The maximal energy and particle masses from the event_kinematics
         object define the maximal range, i.e. the energy requested in subsequent
-        `set_event_kinematics` calls should not exceed the one provided here.
+        `_set_event_kinematics` calls should not exceed the one provided here.
 
         Args:
             event_kinematics (object): maximal energy and masses for subsequent runs
@@ -475,7 +474,7 @@ class MCRun(six.with_metaclass(ABCMeta)):
         pass
 
     @abstractmethod
-    def set_event_kinematics(self, evtkin):
+    def _set_event_kinematics(self, evtkin):
         """Set new combination of energy, momentum, projectile
         and target combination for next event.
 
@@ -493,7 +492,7 @@ class MCRun(six.with_metaclass(ABCMeta)):
     
     @event_kinematics.setter
     def event_kinematics(self, evtkin):
-        self.set_event_kinematics(evtkin)
+        self._set_event_kinematics(evtkin)
 
     @abstractmethod
     def set_stable(self, pdgid, stable=True):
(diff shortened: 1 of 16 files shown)


afedynitch:
Hi @HDembinski,

it would be nice if you can review the interface changes, since most of these have been proposed by you a while ago. Let us know if you can do it, or if it would cause delays.

HDembinski:
Ok will do, excited to see this.

HDembinski:
I am currently on a workshop and today is the excursion + dinner day, so I won't be able to make comments until tomorrow. The generator feature is great, but I wonder whether the filter can be made more general. Design-wise it is better give users a  general system to filter that's easy to understand than specific functions, but perhaps it has to be this way for reasons that I don't understand yet. Is this design inspired by Pythia?

afedynitch:
Very cool. Thanks!!
It's not something to be done within a day. My question was about days, or rather weeks...

For the filters, there is maybe no deep reason. I personally think that numpy boolean selections or np.where-s are a sufficient filtering framework. But preselecting charged and stable particles is so common that requiring always an additional expression `event.status == 1` in your selections is quite annoying every time. I think what is needed are just these two "pre-filters" (for final state and charge) and the rest is up to the user to select stuff, e.g. something like `np.where((event.p_ids == 211) & (np.abs(event.eta) < 2.5))` or way more complex expressions.

Regarding PYTHIA, this way of selecting particles is not specific to it. Selecting events in a fortran loop over the events does the same thing. I feel that Pythia8 works according to previous fortran logic but is just written in a different language.

impy/common.py
```diff
@@ -18,6 +18,28 @@
...
+        self.py = py
+        self.pz = pz
+        self.en = en
+        self.m = m
+        self.vx = vx
+        self.vy = vy
+        self.vz = vz
+        self.vt = vt
+        self.pem_arr = pem_arr
+        self.vt_arr = vt_arr
```
HDembinski, line 40: Should these be public? If not, use `self._pem_arr` and `self._vt_arr`. Please make everything private that is only for internal use. One can always later make private stuff public if necessary but vice versa is a breaking change.

impy/common.py
```diff
@@ -18,6 +18,28 @@
 from impy.util import info
 
 
+class EventData:
+    """Data structure to keep views of filtered data"""
+    def __init__(self, npart, p_ids, status, charge,
+                 px, py, pz, en, m, vx, vy, vz, vt,
+                 pem_arr, vt_arr):
+        self.npart = npart
```
HDembinski, line 26: This looks like a derived quantity that should be a property?

```py
@property
def npart(self):
     return len(self.px)
```

impy/common.py
```diff
@@ -18,6 +18,28 @@
...
+        self.charge = charge
+        self.px = px
+        self.py = py
+        self.pz = pz
+        self.en = en
+        self.m = m
+        self.vx = vx
+        self.vy = vy
+        self.vz = vz
+        self.vt = vt
```
HDembinski, line 38: Perhaps these arrays should be made read-only, so that users cannot accidentally override the values?

impy/common.py
```diff
@@ -81,19 +105,73 @@ def __init__(self, lib, event_kinematics, event_frame, nevent, npart,
...
+    def _cache_current_view(self):
+        """Cache the current view of data"""
+        self._views_cache.append(EventData(
+            self._npart, self.p_ids, self.status, self.charge,
+            self.px, self.py, self.pz, self.en, self.m,
+            self.vx, self.vy, self.vz, self.vt,
+            self.pem_arr, self.vt_arr
+        ))
+    
+    def remove_last_filter(self):
```
HDembinski, line 146: I do not understand the need to remove filters. Users familiar with numpy know that applying a mask is destructive (`b = a[mask]`). This is fine in numpy because numpy does not override the original array.

I think impy should either follow this approach and use masks to perform irreversible reductions like numpy, but then not do them in place. Or you should use [numpy's masked arrays](https://numpy.org/doc/stable/reference/maskedarray.html) to keep the in-place filtering non-destructive and reversible.

There shouldn't be a need to cache previous filtering results, since the only thing that needs to be modified is the array mask.

impy/common.py
```diff
@@ -335,6 +413,11 @@ def __init__(self, interaction_model_def, settings_dict=dict(), **kwargs):
 
         # FORTRAN LUN that keeps logfile handle
         self.output_lun = None
+        
+        # Number of events to generate
+        self.nevents = None
+        # Normalization 1/nevents for histograms
+        self.norm = None
```
HDembinski, line 420: I recommend to remove `norm` or make it a property. Keeping values in sync by hand is a maintenance nightmare. 

If `norm` only exists to calculating the histogram in the way it is described in the PR, then `norm` should be removed. It is both inefficient and inaccurate to calculate a histogram in this way. I will explain this elsewhere.

impy/common.py
```diff
@@ -404,6 +487,14 @@ def set_event_kinematics(self, evtkin):
         """
         pass
 
+    @property
+    def event_kinematics(self):
+        return self._curr_event_kin
+    
+    @event_kinematics.setter
+    def event_kinematics(self, evtkin):
+        self.set_event_kinematics(evtkin)
```
HDembinski, line 496: Make `set_event_kinematics` private, so that there is only "one obvious way" to set it (see Zen of Python), via the property.

impy/common.py
```diff
@@ -535,6 +626,32 @@ def _define_default_fs_particles(self):
...
+                yield self._event_class(self.lib, self._curr_event_kin,
+                                        self._output_frame)
+                nremaining -= 1
+                ntrials += 1
+            elif retry_on_rejection:
+                info(10, 'Rejection occured. Retrying..')
+                ntrials += 1
+                continue
+            elif ntrials > 2 * nevents:
+                raise Exception('Things run bad. Check your input.')
```
HDembinski, line 651: Perhaps better raise `RuntimeError`?

HDembinski: It is good to use standard exceptions, but use the ones which are as specific as possible to the type of error.

impy/common.py
```diff
@@ -535,6 +626,32 @@ def _define_default_fs_particles(self):
...
+        self.nevents = nevents
+        self.norm = 1./float(nevents)
+        while nremaining > 0:
+            if self.generate_event() == 0:
+                yield self._event_class(self.lib, self._curr_event_kin,
+                                        self._output_frame)
+                nremaining -= 1
+                ntrials += 1
+            elif retry_on_rejection:
+                info(10, 'Rejection occured. Retrying..')
```
HDembinski, line 647: I haven't seen yet how `info` is implemented, but if this a custom solution, I recommend to use one of the major logging libraries, because doing logging right is not trivial.

impy/models/__init__.py
```diff
@@ -0,0 +1,10 @@
+from impy.models.sophia import Sophia20
+from impy.models.sibyll import Sibyll21, Sibyll23, Sibyll23c, Sibyll23c00, Sibyll23c01,\
+    Sibyll23c02, Sibyll23c03, Sibyll23c04, Sibyll23d
+from impy.models.dpmjetIII import DpmjetIII191, DpmjetIII192, DpmjetIII306
+from impy.models.epos import EposLHC
+from impy.models.phojet import Phojet112, Phojet191
+from impy.models.urqmd import UrQMD34
+from impy.models.qgsjet import QGSJet01c, QGSJetII03, QGSJetII04
+from impy.models.pythia6 import Pyphia6
+from impy.models.pythia8 import Pyphia8
```
HDembinski, line 10: Is there some notable initialization cost when these models are imported? If yes, then `__init__.py` should not all import them. For any given run, I will mostly only need one of these models, but the way it is written here, I always have to pay for the initiation cost of all of them.

impy/models/dpmjetIII.py
```diff
@@ -272,3 +272,25 @@ def generate_event(self):
         reject = self.lib.dt_kkinc(*self._dpmjet_tup(), kkmat=-1)
         self.lib.dtevno.nevent += 1
         return reject
+
+class DpmjetIII191(DpmjetIIIRun):
+    def __init__(self, event_kinematics, seed="random", logfname=None):
+        from impy.definitions import interaction_model_by_tag as models_dict
+        interaction_model_def = models_dict["DPMJETIII191"]       
+        super(DpmjetIIIRun, self).__init__(interaction_model_def)
```
HDembinski, line 280: Since Python3, `super().__init__` is sufficient here. It makes no sense to still support Python2, which has reached end of life. Numpy dropped Python3.6 support, so the baseline to program for at this point for a scientific library is Python 3.7.

impy/models/sibyll.py
```diff
@@ -210,3 +210,67 @@ def generate_event(self):
...
         self.conv_hepevt()
         return 0  # SIBYLL never rejects
+
+class Sibyll21(SIBYLLRun):
+    def __init__(self, event_kinematics, seed="random", logfname=None):
+        from impy.definitions import interaction_model_by_tag as models_dict
+        interaction_model_def = models_dict["SIBYLL21"]       
+        super(SIBYLLRun, self).__init__(interaction_model_def)
+        self.init_generator(event_kinematics, seed, logfname)
+
```
HDembinski, line 220: If you are not appalled by a bit of "Python metaprogramming", you could generate all these classes in a single loop, by generating Python source code as a string, which is then passed to `exec`. If you find this appalling, I think this is fine too. It is a lot to type once and repetitive, but very clear once written and you don't have to change this often.

HDembinski: I believe there is also a way to generate classes programmatically. That would still be "metaprogramming" but not via generating Python source code as a string.

HDembinski (commented):
Thanks for entrusting me with a review. I think the generator feature is great. I have added some comments on the code that I have reviewed, which is mostly not related to the new generator feature. I am not happy with the way how filtering works, but this is not strictly a topic of this PR.

So, I summary, I think the PR should be accepted to move forward, but it would be nice to discuss my comment a bit and potentially to follow up with new issues on those points that I raised.

HDembinski:
I strongly advice against this way of histogramming for three reasons:
```py
    # generator.norm = 1/nevents, for example
    protons += generator.norm/widths*np.histogram(...)
```
1) np.histogram returns integers. Adding integers is exact (until you overflow, but that is practically impossible to "achieve" when histogramming data), but you are converting your nice exact integers here into floating point numbers and thus loose precision, since multiplying and adding floating point numbers suffers from roundoff error.
2) This is not efficient. Instead of y = a n1 + a n2 + a n3 + ..., it is equivalent and faster to do `y = a (n1 + n2 + n3 + ...)`. In other words, sum up the integers first once the loop is complete, multiply with `norm/widths`.
3) Use boost-histogram for this instead of np.histogram. boost-histogram was designed for this use case, where you fill chunks of data repeatedly in a loop.

impy/common.py
HDembinski: These are views into the FORTRAN common block, right? What is the memory layout of this block? Is it
px1 px2 px3 px4 ... py1 py2 py3 ... etc or is it px1 py1 ... px2 py2 ... ? If it is the latter, the view could be a single record array. This would make the masking/filtering easier.

HDembinski:
@afedynitch I moved the discussion on filtering to #21, because it is a separate topic from this PR. Please let's continue the discussion there. Once I am able to actually build impy on my computer, I could test my ideas. But before I can do any work, we need a few preparatory steps.
1) I need to figure out how to build impy on my system, potentially with a docker container. 
2) We need to add pre-commit to the repository and apply a standard formatter (black) and linter (flake8) and fix all the issues that these find.

impy/models/sibyll.py
HDembinski: For an example of (admittedly ugly) metaprogramming, see https://github.com/HDembinski/numba-stats/blob/main/src/numba_stats/_util.py#L57

impy/common.py
jncots: Yes, these arrays should be private for now. It will be changed.

impy/common.py
jncots: It can be a property if all other `px`, `py`, .. are properties as well. See below.

impy/common.py
jncots: class EventData was needed only for the cache of views. If `px`, `py`, `pz`, ... are properties then EventData will be removed

impy/common.py
jncots: The reason why the implementation of filtering uses cache of previous filtering results and not mask is because currently filtering of views `px`, `py`,... is made in-place. The alternative is to change `px`, `py`, ... to properties, where mask will be applied to the initial view to Fortran data. Another consideration was that properties are usually slower than direct access to arrays + overhead of creation of new masked view at every invocation. However, it is not critical if every filtered view to `px`,... is accessed not so many times, because we create it anyway at least once. Moreover, at every filtering we apply filters to all views in _apply_slicing() method instead of one property we need.

Considering all above I can change `px`, `py`, ... to properties. Then it will be wise to convert  npart to property:

```python
@property
def npart(self):
     return len(self.p_ids)
```
while keeping initial `npart` as private `_npart`. In this case the properties for momenta will look like :

```python
@property
def px(self):
     return self._pem_arr[0, mask]
``` 

or even

```python
@property
def px(self):
     return self.lib.phep[0, mask]
``` 

where self.lib.phep is the fortran array. The layout of `phep` array is px1 py1, pz1,  ... px2 py2, pz1, so the numpy array will be a view with strides.

About remove filters: if we use `mask` then to undone some filtering we should keep a cache of `mask`s. Or do you mean that we don't need the 'undone'/'clear' filtering functions at all?

impy/common.py
jncots: Ok, it will be removed. Also, nevents could be changed to give the number of generated events and not the number of events passed to generator as it is now.

impy/common.py
jncots: Ok

impy/common.py
jncots: I agree `RuntimeError` will be more specific.

impy/common.py
jncots: I cannot answer much to this comment. Because the `info` function is scattered around the code it is not so easy to change logging system now. Moreover, one should decide what to use instead. I think it deserves its own issue.

impy/models/__init__.py
jncots: All that imports are needed only because it allows to have autocompletion like:

```python
import impy
impy.models. # <- here the autocompletion gives the list of available models
```

The initialisation time is around 1 - 1.5 sec. Interestingly, itt doesn't depend on number of imports, only on the fact that there is an import in __init__.py file. The cost of initialisation like:
```python
from impy.models.sophia import Sophia20
```
is around 0.01 sec. 

I don't know whether 1 sec worth it to sacrifice the convenience of the autocompletion.

impy/models/dpmjetIII.py
jncots: Ok, it will be changed to `super().__init__`. Althought the intent was to call `__init__` of parent of super() (i.e. parent of `DpmjetIIIRun` in this case). However, as Anatoli noticed, it is better to call parent's `__init__` in case it will appear in the future.

impy/models/sibyll.py
jncots: Yes, I was thinking about metaclass creation, but, again, there will be no autocompletion in this case.

jncots:
> 

Yes, norm will be removed

impy/common.py
HDembinski: Yes, that's what I had in mind. If we know the layout of the underlying memory block, we can apply the mask very efficiently only once.

impy/common.py
HDembinski: Thank you for the discussion. I would like to experiment a bit with ideas on filtering in a separate PR. I think for this PR the code is fine as it is.

impy/common.py
HDembinski: I would like to experiment a bit with filtering in a different PR. I think for this PR, the code is fine.

impy/common.py
HDembinski: Why? I don't see a reason to make everything a property here. You can use a property just for npart and leave the rest as is.

HDembinski: Anyway, it is fine like this for this PR, I would like to experiment a bit with this class in a separate PR.

impy/common.py
HDembinski: You are right, for this PR there is no change needed.

impy/models/__init__.py
HDembinski: Ok, fine!

impy/common.py
afedynitch: Info is implemented in util and prints out the caller name (class and method) along with the message, if the level is above the one defined by config['debug_level'], otherwise it just returns without.

Not sure we really need a super extensive logging, but if, it should be a drop in replacement anyways, and it should print the caller name. Also it hopefully is a light dependency without involving too many crazy requirements.

impy/models/sibyll.py
HDembinski: What I was proposing creates the exact same result with autocompletion and all. But no need to change this, this code is repetitive but easy to understand.

impy/common.py
HDembinski: I think that would be a very important change indeed! I was not aware that the two numbers can be different.

impy/models/dpmjetIII.py
HDembinski: Yes. Calling `__init__` from a base class higher up the hierarchy is going against principles of OOP and likely to cause bugs in the future.

HDembinski:
@jncots @afedynitch I resolved the discussions which were mostly for my education. I left those open where we agreed on changes. @jncots Could you implement the changes and then mark the discussions as resolved? Would be great to merge this PR soon, so that #22 can follow.

jncots:
The changes have been applied. I didn't change anything related to filtering system, because it is now a separete issue. However, because of pre-commit.ci checks were failed.

HDembinski:
@jncots Please try to create an empty .pre-commit-config.yaml file in the project directory, that should "fix" the pre-commit error. Or perhaps the owner (@afedynitch ?) can simply merge anyway, ignoring the failing test.

HDembinski (approved):
looks all good

jncots:
Ok. It seems that adding the empty .pre-commit-config.yaml doesn't fix a problem. So do I merge?

afedynitch:
@HDembinski Since you mentioned this previously, I opened a new issue on the build issues since I encountered some on a system I regularly work on. See #23

PR #22, opened 2022-08-04
enable pre-commit with black and yaml check

### LINKED ISSUE #16: Add pre-commit and use black formatter
pre-commit is a tool to enforce coding standards on a repository. It runs a few quick tests whenever a developer locally attempts to `git commit`.

pre-commit automatically ensures that new code that is committed is formatted according to the guidelines of the project. When several people are working the same project, this is extremely useful. For example, my editor is configured to use apply the `black` formatter automatically to Python code.

This project is currently not black-formatted, so when I try to patch something, the patch gets blown up by whitespace changes. Rather than disabling black individually for this project, I propose to follow the masses (all scikit-hep projects use black, as do many many other projects) and also auto-format with black.

I will set up a pre-commit configuration. It also automatically applies the flake8 linter, which works very well. It does not generate false positives, but finds a lot of annoyances and clutter, like variables that are imported but not used.

HDembinski:
Closes #16 

This is a minimal config for pre-commit, which just does some essential checks and auto-formats with black. Eventually, we should also enable flake8, but at the moment it finds too many issues, so I want to defer enabling this to a later PR.

I did not perform any code changes in Python*, all changes are made by black.

*) I did turn two left-overs from Python2 `print something` into `print(something)` in two tests, because black does not support Python2 anymore.

The yaml-checker already found a minor issue, a duplicated key in one of the configurations. 👍 

PS: I apparently succeeded in enabling pre-commit also as a Github service, because this PR now runs the pre-commit check.

PPS: You probably want to merge #20 first. This will create merge conflicts here, but I can resolve them. No other PRs should be started in the meantime until this is merged.


impy/impy_config.yaml
```diff
@@ -104,7 +104,6 @@ urqmd:
     # at plab < 4.9 use euation of state and Lorentz contraction of projectile
     # and target. This is not set necessarily in CORSIKA
     LE_EOS       : True
-    debug_level  : 0
```
afedynitch, line 107: This one sets a separate [debug level for URQMD in FORTRAN](https://github.com/impy-project/impy/blob/44e548d69c240c9c37c1d8a6e2314e1dba578187/impy/models/urqmd.py#L178). It's in a subdictionary, so shouldn't generate a name clash. It might be better to simply change this to urqmd_debug_level or so to get a clean yaml check.

afedynitch (approved):
Looks all good except at one place in the yaml file. The checker deleted "debug_level", which is a subkey for UrQMD. I renamed the key to `urqmd_dbglvl`.

I also agree that merging #20 should go first. So please wait for it before merging this one.

impy/impy_config.yaml
HDembinski: There are two debug_level defined in the same dict, this is an error, please look again.

HDembinski: The linters I recommend to enable do not generate false positives in my experience, so it is safe to apply them. There are some linters which do, and we should keep away from those. They are causing more harm than good.

afedynitch: OK! Gotcha. Oversight. Actually all generators have that flag ;)
I reverted my change. The PR is good to go once #20 is merged.

afedynitch (approved):
Good to go after #20

PR #24, opened 2022-08-05
Update README.md

HDembinski:
You have to install impy first before running make, since otherwise you may not have numpy installed (try the commands in a fresh venv and you will get an error otherwise, been there, done that)


afedynitch (approved):
OK!

PR #26, opened 2022-08-05
move to pyhepmc, removed some unused variables in setup.py

### LINKED ISSUE #15: Upgrade to pyhepmc
The package pyhepmc-ng has been replaced with pyhepmc, so please install pyhepmc now instead of pyhepmc_ng and rename the module everywhere to pyhepmc. From what I have seen, this switch should be painless apart from the renaming.

I am happy to do this, if you give me permissions to create branches.

HDembinski:
Closes #15 

This updates from pyhepmc-ng, which was only provided as a source package, to pyhepmc, which is released as binary wheels, so the user does not have to compile anything.

I also commented out/removed a few unused variables that flake8 was complaining about in setup.py.

I checked that test_hepmc_writer.py passes, but I also noticed that 5 tests fail. I am sure that these errors are not related to my changes and therefore this should not delay merging of this PR.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,3 @@
 [flake8]
 max-line-length = 90
+extend-ignore = E203, F403, E402
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,6 @@ https://github.com/pybind/cmake_example/blob/master/setup.py
 """
 
 import os
-import re
 import sys
 import platform
 import subprocess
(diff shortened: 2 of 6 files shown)


setup.cfg
```diff
@@ -1,2 +1,3 @@
 [flake8]
 max-line-length = 90
+extend-ignore = E203, F403, E402
```
afedynitch, line 3: Is this a permanent proposal to ignore those?

setup.py
```diff
@@ -67,7 +69,7 @@ def build_extension(self, ext):
                 "LEXT=" + libext,
             ]
         else:
-            make_command = "make"
+            # make_command = "make"
```
afedynitch, line 72: Right now it doesn't seem to work anyways. And I never use this actually.

afedynitch (approved):
Good to go. 2 comments for curiousity.

setup.cfg
HDembinski: These ignores are temporary for now.

PR #29, opened 2022-08-05
install with docker

HDembinski:
Instructions on how to install impy in a docker instance that is isolated from the host OS. This should always work and never fail. Could a reviewer please try these instructions to see that I haven't forgotten anything?


jncots:
I've tried to follow the instructions ans successfully installed `impy` with slight changes related to the locations of python distributions in `quay.io/pypa/manylinux2010_x86_64` image:

```bash
# discover python distributions in the container
ls /opt/python
# With python36:
/opt/python/cp36-cp36m/bin/python -m pip install -e .
PYTHON_EXE=/opt/python/cp36-cp36m/bin/python make -j8
```

The same works for `python/cp37-cp37m`. With other distributions the installation failed because of absence of wheel for `scipy`. The build for `scipy` from sources failes with the message `../../scipy/meson.build:120:0: ERROR: Dependency "openblas" not found, tried pkgconfig and cmake`.

HDembinski:
Thank you, I forgot an important step, to create a venv inside the docker image. This is better than setting PYTHON_EXE.

HDembinski:
Another change is to use `pip install --prefer-binary` since otherwise it will try to compile a recent scipy from source for the older Python versions instead of using the last scipy wheel available.

HDembinski:
@jncots Thank you for testing, I updated the instruction, does it now work?

jncots:
Yes, it works now, except `source` command should be added, and I would add (matter of taste) a version to venv name something like

```bash
python3.8 -m venv venv3.8
source ./venv3.8/bin/activate

# to exit virtual environment
deactivate
```

HDembinski:
> Yes, it works now, except `source` command should be added, and I would add (matter of taste) a version to venv name something like
> 
> ```shell
> python3.8 -m venv venv3.8
> source ./venv3.8/bin/activate
> 
> # to exit virtual environment
> deactivate
> ```

Oops, yes, thanks

HDembinski:
@jncots If it is ok now, please merge.

afedynitch (commented):
I tried running the commands on a Apple Silicon mac and Docker Desktop on aarch64.

It worked actually well and very fast. I suggest to move to the newer image that has an aarch64 version. So the corresponding lines become:
```
# download linux image for x86_64
    docker pull quay.io/pypa/manylinux2014_x86_64
# or aarch64 if on this architecture or Apple Silicon
    docker pull quay.io/pypa/manylinux2014_aarch64
```

except that there is a deeper reason behind the GCC-8? toolchain of manylinux2010.

HDembinski:
> I suggest to move to the newer image that has an aarch64 version. So the corresponding lines become:
> 
> ```
> # download linux image for x86_64
>     docker pull quay.io/pypa/manylinux2014_x86_64
> # or aarch64 if on this architecture or Apple Silicon
>     docker pull quay.io/pypa/manylinux2014_aarch64
> ```

I am by no means an expert on this, @henryiii is, but that manylinux is old is a feature and not a bug. With the newer image we lose support for older Python versions, but that is fine, I think.

Could you please commit this fix directly onto the branch covered by this PR?

> except that there is a deeper reason behind the GCC-8? toolchain of manylinux2010.

There is no deep reason behind the gcc version, but the point of manylinux is to provide a sufficiently old version of libc to which all newer versions are backward compatible.

gcc-8 is more than enough for this code. You are not using modern versions of C++, so you don't have to worry about the compiler version anyway. I started writing Boost Histogram for gcc-5, which is able to compile C++14 just fine. gcc-8 supports C++17.

afedynitch (approved):
Works

afedynitch:
The number of bugs fixed in gfortran shouldn't be underestimated either. We have a collection of codes spanning >30 years of styles. So, there is some challenge to the compiler beyond just API compatibility. For aarch64 or Apple Silicon one needs gcc > 11 anyways. Merging...

HDembinski:
> The number of bugs fixed in gfortran shouldn't be underestimated either. We have a collection of codes spanning >30 years of styles. So, there is some challenge to the compiler beyond just API compatibility. For aarch64 or Apple Silicon one needs gcc > 11 anyways. Merging...

On the contrary. Newer compilers will choke on errors that the code has which older compilers silently ignored. For old FORTRAN code as in this library, you probably want to use an old compiler.

It is safe to use the manylinux image that numpy uses. That's what I suggested. For aarch you need a more recent image, but not for x86_64. These choices are not made arbitrary. But anyway, manylinux2014 is fine since we don't want to support old Python versions.

PR #30, opened 2022-08-06
Enable flake8

HDembinski:
I enabled flake8 and fixed the issues it found. While doing that, I converted some of the manual tests into automated tests. This is done very crudely at the moment, but better than nothing. We urgently need automated testing, because I now see that impy is largely in a broken state. Many of the models are not working in my Linux docker instance, in particular all SIBYLL versions just shut down the Python interpreter without an error message.

It turns out that the switch from pyhepmc-ng to pyhepmc broke the HepMCWriter. I did not notice this, because the corresponding test used SIBYLL and as mentioned before, SIBYLL just kills the interpreter before I could see that the test was not working.

Another general thing I noticed: unit tests are not documentation and are not examples. Tests should not explain how to use some interface, that should be part of the docstring of that interface. Examples go into a separate folder, they are not tests.

- I sprinkled the code with TODO and FIXME. Anatoly, can you please have a look at those? I also found several hardcoded paths.
- Can we please make `particle` a dependency for `impy`, so I can use well-defined particle names in the library instead of PDG ids like 211, 2212?
- I corrected a typo in the class names for Pythia, which was spelled Pyphia
- I removed `misc_test_func.py` which contained `plot_hist`, this is not necessary anymore since we now have `matplotlib.pyplot.stairs`; the function was also not used anywhere
- I removed `from __future__ import print_function`, because we have to drop Python2 support anyway


afedynitch (approved):
OK, thanks for the tedious work

PR #33, opened 2022-08-08
Fix/init generator

### LINKED ISSUE #27: Example in README is broken now
```py
from impy.definitions import *
from impy.constants import *
from impy.kinematics import EventKinematics
from impy import impy_config

# Define the parameters of the collisions
event_kinematics = EventKinematics(
    ecm=13 * TeV, p1pdg=2212, p2pdg=2212)

# Create an instance of an event generator by passing
# the model name as a string
generator = make_generator_instance(
    interaction_model_by_tag['SIBYLL2.3D'])

# Initialize it
generator.init_generator(event_kinematics)

# Number of events to generate
nevents = 100

for event in generator.event_generator(event_kinematics, nevents):
    event.filter_final_state_charged()
    # do something with event.p_ids, event.eta, event.en, event.pt, etc.
    # these variables are numpy arrays, that can be histogrammed or counted like
    average_pt += 1/float(nevents)*np.mean(event.pt[np.abs(event.p_ids) == 211])

print('Average pT for charged pions {0:4.3f}'.format(average_pt))
```
After merging #7 the example in the README does not work anymore and needs to be updated.

jncots:
The example in README is changed to the working version. In the process it turned out that  _set_event kinematics call in init_generator is at a wrong position and calls  library functions before proper initialization. This caused issues that was fixed by the reodering. Closes #27.


HDembinski:
If `_set_event_kinematics` has these pre-requisites, could you please add comments in the code to express that?

In addition and if it is possible, there should be an assert inside `_set_event_kinematics` that fails if the generator was not initialized.

Even better is to design the API in such a way that certain functions can only be called in the correct order. I did this in the past for the IceCube Event Browser, but I don't understand the code well enough to suggest how it can be implemented here. The pattern is to make `_set_event_kinematics` depend on a variable in its signature that is returned by the initialization code. This is way, the dependencies in the code are made explicit in the API.

HDembinski (commented):
Please add comments about this surprising aspect ~and then it is ready to merge~.

Edit: Please have a look at the failures that I found below.

HDembinski:
@jncots Can you try running the test `test_cms_all.py`? I just updated and fixed this test. Now the test itself works and produces useful output.

After installing impy in the docker image, please install pytest `pip install pytest` and then
`python -m pytest impy/tests/test_cms_all.py -s -vv`

I am seeing failures for a range of models, some are probably related to the API change:

E     +  'DpmjetIII192',
E     +  'Phojet112',
E     +  'Phojet191',
E     +  'Pythia6',
E     +  'Pythia8',
E     +  'QGSJetII03',
E     +  'QGSJetII04',
E     +  'Sophia20',

The failure of QGSJetII may be related to the fact that it has to generate large tables on first run, and I am not patient enough at the moment to let QGSJetII run for a few hours to generate these tables. @afedynitch Can we somehow supply the tables? Github has large file support, I have never used it, but it is for this sort of thing.

The other models seem to fail because of errors in Python.

DpmjetIII192 init
DpmjetIII192 abort
  File "/app/impy/tests/test_all_cms.py", line 35, in run_generator
    generator = model(event_kinematics, logfname=log.name)
  File "/app/impy/models/dpmjetIII.py", line 300, in __init__
    super().__init__(interaction_model_def)
  File "/app/impy/common.py", line 450, in __init__
    self.lib = importlib.import_module(interaction_model_def.library_name)
  File "/opt/_internal/cpython-3.8.13/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 973, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'dpmjetIII192'

Phojet112 init
Phojet112 abort
  File "/app/impy/tests/test_all_cms.py", line 35, in run_generator
    generator = model(event_kinematics, logfname=log.name)
  File "/app/impy/models/phojet.py", line 281, in __init__
    self.init_generator(event_kinematics, seed, logfname)
  File "/app/impy/models/phojet.py", line 242, in init_generator
    if self.lib.pho_event(-1, self.p1, self.p2)[1]:
AttributeError: 'Phojet112' object has no attribute 'p1'

Phojet191 init
Phojet191 abort
  File "/app/impy/tests/test_all_cms.py", line 35, in run_generator
    generator = model(event_kinematics, logfname=log.name)
  File "/app/impy/models/phojet.py", line 290, in __init__
    self.init_generator(event_kinematics, seed, logfname)
  File "/app/impy/models/phojet.py", line 242, in init_generator
    if self.lib.pho_event(-1, self.p1, self.p2)[1]:
AttributeError: 'Phojet191' object has no attribute 'p1'

Pythia8 init
Pythia8 abort
  File "/app/impy/tests/test_all_cms.py", line 35, in run_generator
    generator = model(event_kinematics, logfname=log.name)
  File "/app/impy/models/pythia8.py", line 296, in __init__
    super().__init__(interaction_model_def)
  File "/app/impy/common.py", line 450, in __init__
    self.lib = importlib.import_module(interaction_model_def.library_name)
  File "/opt/_internal/cpython-3.8.13/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 973, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'pythia8'

Sophia20 init
Sophia20 abort
  File "/app/impy/tests/test_all_cms.py", line 35, in run_generator
    generator = model(event_kinematics, logfname=log.name)
  File "/app/impy/models/sophia.py", line 249, in __init__
    self.init_generator(event_kinematics, seed, logfname)
  File "/app/impy/models/sophia.py", line 195, in init_generator
    self.nucleon_code_number
AttributeError: 'Sophia20' object has no attribute 'nucleon_code_number'

Since Pythia8 is not really supported, I suggest to not import it in `impy.models` until it works. It will then also disappear in this test.

afedynitch:
Indeed, it seems that most fails are due to API changes.

The tables are on LFS and I’m even paying for traffic. Some recent change must have broken this.

jncots:
There were following problems:

* `Phojet191` and `Phojet112` did not pass tests indeed because of changes in API. It was a problem with `_set_event_kinematics` call in the wrong place. For different generators `_set_event_kinematics` should be placed in different order.
* `DpmjetIII192` and `Pythia8` did not find library files because of the absence of these files. In case of `Pythia8` it had not been generated. In case of `DpmjetIII192` `dpmjetIII193` is generated because of version change. I changed the name and the version information for `DpmjetIII192` to `DpmjetIII193`
* 'QGSJetII04', 'DpmjetIII191' failed because of long run. I changed the line in the file with tests `r = job.get(timeout=30)` to 30 sec and now they pass
* With `Pythia6` I could not manage to understand what is the problem. It just freezes. And it seems not a problem with API change.
* `Sophia20` don't pass tests because it accepts only photon as first argument and currently only lab frame.

HDembinski:
> 'QGSJetII04', 'DpmjetIII191' failed because of long run. I changed the line in the file with tests r = job.get(timeout=30) to 30 sec and now they pass

This is strange, because I waited much longer than that and they did not succeed. I will try again with this version of the PR.

HDembinski:
> There were following problems:
> 
>     * `Phojet191` and `Phojet112` did not pass tests indeed because of changes in API. It was a problem with `_set_event_kinematics` call in the wrong place. For different generators `_set_event_kinematics` should be placed in different order.
> 
>     * `DpmjetIII192` and `Pythia8` did not find library files because of the absence of these files. In case of `Pythia8` it had not been generated. In case of `DpmjetIII192` `dpmjetIII193` is generated because of version change. I changed the name and the version information for `DpmjetIII192` to `DpmjetIII193`
> 
>     * 'QGSJetII04', 'DpmjetIII191' failed because of long run. I changed the line in the file with tests `r = job.get(timeout=30)` to 30 sec and now they pass
> 
>     * With `Pythia6` I could not manage to understand what is the problem. It just freezes. And it seems not a problem with API change.
> 
>     * `Sophia20` don't pass tests because it accepts only photon as first argument and currently only lab frame.

@jncots Thank you for the update. I tried again on my machine and I am still seeing QGSJetII failing.

```
E         +  'DpmjetIII191',
E         +  'DpmjetIII193',
E         +  'Phojet191',
E         +  'QGSJetII03',
E         +  'QGSJetII04',
E         +  'Sophia20',
```
The QGSJetII errors are probably related to this error message
```
At line 463 of file qgsjet-II-04.f (unit = 1, file = '/app/impy/iamdata/qgsjet/qgsdat-II-04')
Fortran runtime error: Bad real number in item 1 of list input
At line 363 of file qgsjet-II-03.f (unit = 1, file = '/app/impy/iamdata/qgsjet/qgsdat-II-03')
Fortran runtime error: Unformatted file structure has been corrupted
```
I am also getting compilation errors now. For `src/dpmjetIII-19.1` the installation seems to fail and `src/sophia` is recompiling every time, although it seems to succeed.
```

It is fair to skip Sophia20 in the test, since it cannot succeed on these inputs.

**Comment by HDembinski** [comment]
@jncots I looked into /app/impy/iamdata/qgsjet/qgsdat-II-03 and it turns out that this is only a link (?) on my computer.

```
version https://git-lfs.github.com/spec/v1
oid sha256:3e5dde3b17c7a2b36243de199d545ee1c5ed5bc89216515d0466f84c37866899
size 131514464
```

What are the instructions to actually download the real file from git-lfs? These need to be added to the README.md.

**Comment by jncots** [comment]
Ok.  I leave the instructions for manylinux here:

```bash
yum install wget
wget https://github.com/git-lfs/git-lfs/releases/download/v3.2.0/git-lfs-linux-amd64-v3.2.0.tar.gz
tar -xvf git-lfs-linux-amd64-v3.2.0.tar.gz
cd git-lfs-3.2.0/
./install.sh
```

Later it will be added to README.md

afedynitch:
The problem with large non-versioned binaries is solved for MCEq by attaching these files as binaries along github releases.
See [the mceq_db file here](https://github.com/afedynitch/MCEq/releases/tag/release_1_2_0). The files get permanent URLs. On first run, a piece of code in mceq_config checks the SHA and downloads them onto the system.
@HDembinski @jncots, what about using the same method instead of lfs? These binaries never change so there is no point in version track them.

HDembinski:
Sounds great, let's do that and download them on demand.

HDembinski:
@jncots Did you make progress on fixing the other models? I checked Sophia20 in #39 and it seems to work. This leaves the following models in a non-working state (at least for me):

'DpmjetIII191'
'DpmjetIII193'
'Phojet191'

HDembinski:
Phojet112 fails because it never sets the parameter file paths.

The two if conditions in phojet.py:205 and phojet.py:212 are skipped, because the dpmjet306 module neither contains the functions `parfn` nor `datdir`.

jncots:
`python -m pytest impy/tests/test_all_cms.py -s -vv` test in `quay.io/pypa/manylinux2010_x86_64` and `quay.io/pypa/manylinux2014_x86_64` has the same problems:

*  'QGSJetII04' and 'QGSJetII03' have the problem:

```bash
ImportError: /app_old/impy/lib/qgsII04.cpython-38-x86_64-linux-gnu.so: undefined symbol: _ZGVbN2v_log
```

All `Sibyll21` - `Sibyll23c03`  and 'QGSJet01c' have the problem:

```bash
ImportError: /app_old/impy/lib/*libname*.cpython-38-x86_64-linux-gnu.so: undefined symbol: _ZGVbN2v_exp
```

* After adding in src/Makefile `SUBDIRS = ...  dpmjetIII-19.3 ...`
DpmjetIII193 works

* After the changes in `/src/dpmjetIII-19.1/common/dpmjetIII191.pyf` in lines

```fortran
double precision dimension(24.0,22.0) :: rcbfbk
double precision dimension(24.0,22.0) :: exjpfb
....
integer dimension(23.0,21.0,2) :: ipsind
integer dimension(23.0,21.0,2) :: ifbind
```
from real to integer dimensions DpmjetIII191 also works.


* Phojet112, Phojet191, UrQMD34 also pass the test.

* Pythia6 has a problem with random generator which I am fixing currently.

HDembinski:
> `python -m pytest impy/tests/test_all_cms.py -s -vv` test in `quay.io/pypa/manylinux2010_x86_64` and `quay.io/pypa/manylinux2014_x86_64` has the same problems:
> 
>     * 'QGSJetII04' and 'QGSJetII03' have the problem:
> 
> 
> ```shell
> ImportError: /app_old/impy/lib/qgsII04.cpython-38-x86_64-linux-gnu.so: undefined symbol: _ZGVbN2v_log
> ```
> 
>     * All `Sibyll21` - `Sibyll23c03`  and 'QGSJet01c' have the problem:
> 
> 
> ```shell
> ImportError: /app_old/impy/lib/*libname*.cpython-38-x86_64-linux-gnu.so: undefined symbol: _ZGVbN2v_exp
> ```

This sounds like we need to link to `libm`, which we do not.

jncots:
It turned out that the problem was with leftover files from past compilations. Now all tests pass. The pull request could be closed after the conflict with `src/Makefile` is resolved.

jncots:
The conflict with src/Makefile is resolved.

README.md
```diff
@@ -5,33 +5,32 @@ This package implements are generic user interface to popular event generators u
...
-    average_pt += 1/float(nevents)*np.mean(event.pt[np.abs(event.p_ids) == 211])
-
-print('Average pT for charged pions {0:4.3f}'.format(average_pt))
+    pt = event.pt[np.abs(event.p_ids) == 211]
+    # The list could be empty
+    if len(pt) > 0:
+        nevents += 1
+        average_pt += np.mean(pt)
+
+average_pt = 1 / float(nevents) * average_pt
```
HDembinski, line 32: While this line is faster than `average_pt / nevents`, I would keep the code as simple as possible in a pedagogical example, because the difference in speed is imperceptible. I suggest to change the line to `average_pt = average_pt / nevents`. This will work as intended, no `float` needed.

This does not work: `average_pt /= nevents`, because it is an in-place operation. The numpy devs decided that this should not change the dtype of the array.

HDembinski (approved):
Looks good.

HDembinski:
In addition to the review, I confirm that test_all_cms.py passes on my computer with python3.9 in manylinux2010_x86_64.

README.md
jncots: Changed.

PR #35, opened 2022-08-09
move impy to src/impy, move tests and examples to proper places

HDembinski:
This moves `impy` to `src/impy`. This is commonly done so that `import impy` in a shell inside the project directory does not import `impy` from the project folder, but the installed `impy` version.

I further moved the tests from `impy/tests` to `tests`. This is more convenient and standard in scikit-hep projects. I moved the examples which are placed in the tests folder into `examples`.


HDembinski:
Can we please merge this? It shouldn't be controversial and #39 expects this directory layout.

afedynitch (changes requested):
Tests for all models should pass first. The pace of changes is too high for me personally to test all of them and the code is more sensible than it seems on the first sight.

- [ ] The paths to the files are relative to the base path defined in impy/__init__.py. That has to be updated correctly everywhere.
- [ ] I don't like the new structure that it mixes impy python stuff with the Fortran sources and piles up in one folder, even if it technically works. So either the Fortran folders need a subfolder or impy needs to move somewhere else.

HDembinski:
> I don't like the new structure that it mixes impy python stuff with the Fortran sources and piles up in one folder, even if it technically works. So either the Fortran folders need a subfolder or impy needs to move somewhere else.

In modern packaging, everything that the package needs, data and modules, need to be under one common path. That's why `impy` needs to go under `src`. This is the standard, see https://setuptools.pypa.io/en/latest/userguide/datafiles.html.

HDembinski:
Anyway, I am closing this, as it is superseded by #39

Issue #11, opened 2021-10-25
Compatibility issues with numpy > 1.20


afedynitch:
For some reason, with numpy 1.21.2 only a part of symbols are exposed in python by the Floyd wrapped libraries.
To reproduce, compile with numpy 1.21.2 and check if `dir(sib23d)` contains the functions `sibini` or `isib_pdg2pid`.

For a temporal workaround downgrade numpy. With numpy 1.19.x this problem doesn’t occur.

HDembinski:
I found the related bug in numpy https://github.com/numpy/numpy/issues/22648

HDembinski:
I found a workaround. You can copy the folder f2py from numpy-1.19.5 into the corresponding folder of your local numpy installation. chromo will compile then.

jncots:
Solved in #127

PR #39, opened 2022-08-10
Replacing make with cmake, adding continuous integration

HDembinski:
This is a draft. I have a working prototype for the new cmake-based build system. You just need to do `pip install --prefer-binary -e .` and that's it.

It only works for eposlhc for now, I need to add the other models. Figuring this out was way harder than expected, since the examples on the numpy website do not work for numpy-1.19 and since we want to avoid compiling with f2py. That was causing the problems on my machine. I can build with cmake just fine, no need for docker.

In the cmake build, f2py is only used to generate the wrapper code, the compilation is run by cmake.

To-do:
- [x] Add all models to cmake build
- [x] All models with `test_new_interface.py` pass on CI (except Pythia6 and URQMD34)
- [x] Fortran files are moved to src/fortran
- [x] Use pre-generated f2py files from src/f2py


afedynitch:
Thanks!! 
So does it build a correct library with numpy > 1.20?

HDembinski:
> Thanks!! So does it build a correct library with numpy > 1.20?

Probably, I am not at the point where I want to test this.

HDembinski:
> > Thanks!! So does it build a correct library with numpy > 1.20?
> 
> Probably, I am not at the point where I want to test this.

Seems to work.

HDembinski:
@afedynitch Can you please help me? I added another model, Sibyll21 to the prototype. Compilation succeeds, but the test fails. I cannot figure out why the compiled module is missing a function.

```
self = <impy.models.sibyll.Sibyll21 object at 0x10f8f45b0>
event_kinematics = Event kinematics:
	ecm      : 7000.00000
	pcm      : 3499.99992
	elab     : 26093809.42942
	plab     : 26093809.42942
... : 3727.68720
	bgamm_cm : 3727.68706
	pdgid 1  :       -211
	nucprop 1: 1/-1.0
	pdgid 2  :       2212
	nucprop 2: 12/6


    def _set_event_kinematics(self, event_kinematics):
        """Set new combination of energy, momentum, projectile
        and target combination for next event."""

        info(5, "Setting event kinematics.")
        info(10, event_kinematics)
        k = event_kinematics
        if k.p1_is_nucleus:
            raise Exception("Projectile nuclei not natively supported in SIBYLL")
        elif k.p2_is_nucleus and k.A2 > 20:
            print(k.p2_is_nucleus, k.A2)
            raise Exception("Target nuclei with A>20 not supported in SIBYLL")
>       self._sibproj = self.lib.isib_pdg2pid(k.p1pdg)
E       AttributeError: module 'sib21' has no attribute 'isib_pdg2pid'

src/impy/models/sibyll.py:157: AttributeError
```

Steps to reproduce. In a fresh virtual environment, do:
```sh
pip install --prefer-binary --no-build-isolation -v -e .'[test]' ; python -m pytest tests/test_new_interface.py
```

The f2py logs are `build/temp.<something>/sib21.0.log` and  `build/temp.<something>/sib21.1.log`. The latter looks ok, but the former has lots of lines like this: 
```
         analyzeline: implied-DO list "(sigqcd(k,2),k=    1,   61)" is not supported. Skipping.
```
and
```
Line #8146 in /Users/hdembinski/Extern/impy/src/sibyll/sibyll_21.f:"      CALL NUC_GEOM_INI                 ! nucleus profiles"
        analyzeline: No name/args pattern found for line.
```
That does not seem right.

afedynitch:
The f2py errors or warnings are usually not serious. It's only a problem if they occur in block data statements, because f2py needs to write wrappers for their initialization.

The error comes from the numpy issue. The function isib_pdg2pid doesn't show up in the c wrapper for some unknown reason. This is the numpy > 1.19 issue. For the rest, the tests just hang for me. Please collaborate with @jncots on this issue, since my responses will be slow.

afedynitch:
Sorry,  it as one of the people who has to work with this as a research code, I want a clear separation between the Fortran and python sources. The Fortran sources can be in a sub directory of the impy package if that’s the only solution.

Another important req is that incremental builds should work fast and robust if I change a single Fortran file without cleaning, rebuilding, or reinstalling.

HDembinski:
> Sorry, it as one of the people who has to work with this as a research code, I want a clear separation between the Fortran and python sources. The Fortran sources can be in a sub directory of the impy package if that’s the only solution.

Ok, what kind of layout do you want? I only need everything to sit inside `src`, but I am flexible regarding the substructure. How about this?

```
src/
    impy/
       <Python stuff>
    fortran/
       <fortran stuff>
```

> Another important req is that incremental builds should work fast and robust if I change a single Fortran file without cleaning, rebuilding, or reinstalling.

This is ensured. You can do `pip install --no-build-isolation -e .` during development. I use that too. CMake knows which targets depend on other targets and only re-compiles/re-links what's necessary.

If you work only with the Fortran code, you can also do
```
mkdir build
cd build
cmake ..
make -j8
```
This will compile all the fortran modules, but it will not put them in the proper places. Nevertheless, in that build folder, you can start a python shell and do
```
import eposlhc
```
and call functions of that module.

HDembinski:
# Status

I have successfully ported EPOS-LHC, SIBYLL-2.1, and SIBYLL-2.3d now. I think this is a sufficient demonstration that the cmake-based build works. I had to make many little technical changes to make impy installable as a package in site-packages, just like other packages, and so that installation works with pip, which uses an isolated build environment by default. All that seems to work now.

The CI test for OSX succeeds with these, but it currently only runs `tests/test_new_interface.py` which calls exactly those three models.

HDembinski:
I need to figure out how to use the large files in the tests on the CI runners. It is possible to cache large files, this could be a solution.

afedynitch:
@HDembinski thanks for taking this initiative.

So I checked parallel and incremental builds. The incremental builds are kinda ok and can be made a bit faster by `--no-deps` flag. Parallel builds are significantly slower than before because each module is built in parallel, but not multiple modules in parallel. The recursive makefiles built all subfolders in parallel and were very significantly faster. If there is any chance to make this work, it would simplify the farewell from makefiles.

For DPMJET, the Makefiles need to be maintained because of FLUKA. I think it would be simpler to just redo the makes in CMake leaving DPMJET untouched.

UrQMD, although I fixed it, can not be described in adequate words. The include files have the ending .f and were confused with sources. If you look at what happens there then it becomes clear how much underneath make is going through unnoticed. There also some magic things happening I really cannot even explain at that moment. There are 4 different types files with the random number generators subroutine defined. It seems to find the correct one due to the `mk` folder, somehow. For cmake these files need to be excluded. It probably also doesn't work properly and needs to be tested more carefully (as all the other generators).

One more point is about the debug compiler flags. How can these be introduced in the build?

For the source folder, I prefer either src/impy/models/fortran_sources or src/fortran_src or src/model_src. Something in this direction. The rangen.f and logging.f should be also there.

HDembinski:
> So I checked parallel and incremental builds. The incremental builds are kinda ok and can be made a bit faster by `--no-deps` flag. Parallel builds are significantly slower than before because each module is built in parallel, but not multiple modules in parallel. The recursive makefiles built all subfolders in parallel and were very significantly faster. If there is any chance to make this work, it would simplify the farewell from makefiles.

What's the delay in relative numbers? 5 %, 10 % or more?

Everything is possible, but what you want breaks the structure imposed by Python's extension system. A compiled Python extension expects that the result of a compilation is a single compiled Python module. We generate a separate compiled module for each model, so I have to make an extension for each. Each is calling separately into cmake to do the actual compilation, that's why all the compilations together do not all happen in parallel, although cmake can of course do that. If you run cmake by hand, it will compile everything in parallel.

Thank you for looking into DPMJet and URQMD.

> One more point is about the debug compiler flags. How can these be introduced in the build?

You can do `DEBUG=1 pip install ...` or if you build by hand `cmake -DCMAKE_BUILD_TYPE=Debug ..` in cmake's setup phase. By default, cmake uses something reasonable, `-O0 and -g`, I think, but we can customize this. I haven't ported the debug options from the makefiles yes, but I will look into this.

HDembinski:
> For the source folder, I prefer either src/impy/models/fortran_sources or src/fortran_src or src/model_src. Something in this direction. The rangen.f and logging.f should be also there.

Ok, I will fix this.

HDembinski:
> Everything is possible, but what you want breaks the structure imposed by Python's extension system. A compiled Python extension expects that the result of a compilation is a single compiled Python module. We generate a separate compiled module for each model, so I have to make an extension for each. Each is calling separately into cmake to do the actual compilation, that's why all the compilations together do not all happen in parallel, although cmake can of course do that. If you run cmake by hand, it will compile everything in parallel.

I think I could cheat the system by compiling everything on the first call to generate a CMakeExtension. The subsequent calls then would just check whether the expected output already exists and run cmake only if not. Since the Extensions are compiled sequentially and not in parallel, this should work. I don't like hacks like these, though, so if the problem is not too large, I would prefer to not go there.

afedynitch:
> What's the delay in relative numbers? 5 %, 10 % or more?

New CMake:
```
real    3m41.394s
user    4m27.187s
sys     1m27.568s
```
Makefile (same limited selection of models):
```
real    0m51.459s
user    4m19.278s
sys     0m33.630s
```
Makefile (all models including 3 DPMJET versions with by far the most amount of source files and length):
```
real    0m54.608s
user    7m10.107s
sys     1m4.354s
```

In any case, linker really slows everything down. I hope one can replace it with clang ldd or mold.

HDembinski:
Do I read this correctly, the old makefile was four times faster? That's indeed huge. How did you run the build? With pip or the manual way that I described? With the latter it should be very close to make (I hope).

On the other hand, you are not going to rebuild everything during development, so this comparison is a bit misleading.

HDembinski:
@afedynitch Thanks for fixing the urqmd build. I did not read the makefile carefully enough and missed that only a selected fortran files are included. The way you implemented it was closer to how I had it initially, but I think it is best if I faithfully reproduce what you have in the makefiles, so I changed the build to match the makefile. I hope you also like this better.

HDembinski:
@afedynitch You commented this out, why?

```
        # if model not in (models.Sibyll21, models.Sibyll23d):
        #     # Sibyll fails these, is this expected?
        #     assert event.impact_parameter > 0
        #     assert event.n_wounded_A == 1
        # if model not in (
        #     models.Sibyll21,
        #     models.Sibyll23d,
        #     models.QGSJetII03,
        #     models.QGSJetII04,
        # ):
        #     # Sibyll and QGSJetII fail this, is this expected?
        #     assert event.n_wounded_B > 0
        # # assert event.n_NN_interactions > 0
```

afedynitch:
> Do I read this correctly, the old makefile was four times faster? That's indeed huge. How did you run the build? With pip or the manual way that I described? With the latter it should be very close to make (I hope).
> 
> On the other hand, you are not going to rebuild everything during development, so this comparison is a bit misleading.

It is indeed 4 times faster. I ran just the build through pip without checking dependencies after deleting the build directory, vs a make -j a makefile branch. I didn't run cmake not through pip, yet.

I still think that it's something worth improving if possible, or at least documenting how to speed up. On a fresh clone or inside CI there will quite a delay of 10 minutes or so, which is not really great. From experience, I have built impy more often from scratch rather than incrementally.

How do I run the cmake part only?

afedynitch:
> @afedynitch You commented this out, why?

Because it's not robustly implemented for all the models. This is the reason why QGSJET01 throws an error because it doesn't define these within the same arrays. I think it's enough for this test to just make sure that the models produce some particles. If the impact parameter is <10 is also an arbitrary choice. This test won't run on phojet or pythia, because they don't support nuclei. But, these lines could be something for a dedicated test on nuclei among models that support nuclei that's why it's commented out.

HDembinski:
Okay, I understand and I agree that this particular test should just check whether any output is generated.

If not all models support all event properties, then we should discuss elsewhere how to handle this. I think the attributes should not exist if the model does not provide that information or they should return `None` to indicate that this info is not available.

HDembinski:
@afedynitch I added the hack that I mentioned to make the build faster and I fixed an issue to allow Ninja as a build backend. With Ninja the parallel compilation is much faster. Could you try again with Ninja please?

HDembinski:
On my computer Ninja is not able to accelerate the build significantly, although subjectively it felt faster.
```
CMAKE_GENERATOR=Unix pip install --no-deps --no-build-isolation -v -e .  711.69s user 28.53s system 485% cpu 2:32.46 total
CMAKE_GENERATOR=Ninja pip install --no-deps --no-build-isolation -v -e .  678.51s user 29.58s system 476% cpu 2:28.65 total
```

afedynitch:
Thanks for trying to improve this. On WSL2/linux non docker, I get:
For Ninja: `real 1m18.359s user    4m29.253s sys     0m36.317`
For Unix:  `real 1m22.969s user    0m1.636s sys     0m0.109s`

But there are issues;
1. I need to pass CMAKE_GENERATOR="Unix Makefiles" for the second command to work
2. The Makefile backend has exactly the same speed. The problem is that it gets stuck in "maybe infinite loop" and runs the `cmake` exec over and over again, until I kill it by hand.
3. I had to comment out the `_sib21` target because it gets passed the wrong rangen version with the double precision version of the random number generator. So somehow the cpp flag SIBYLL_SP is not passed around correctly after transitioning to *.fpp

HDembinski:
> Thanks for trying to improve this. On WSL2/linux non docker, I get: For Ninja: `real 1m18.359s user 4m29.253s sys 0m36.317` For Unix: `real 1m22.969s user 0m1.636s sys 0m0.109s`

I will remove the extra code to use detect and use Ninja as the default if it is available, since it does not really speed up things and just makes the implementation more complex.

> But there are issues:
> 
>     1. I need to pass CMAKE_GENERATOR="Unix Makefiles" for the second command to work

Ok, noted, that is in agreement with the cmake documentation. I looked up the value in the CMakeCache.txt, where it just said "Unix".

>     2. The Makefile backend has exactly the same speed. The problem is that it gets stuck in "maybe infinite loop" and runs the `cmake` exec over and over again, until I kill it by hand.

I had a bug in some intermediate version where that happened, but it should be fixed. If I cannot fix this, then we should revert to the original slower but correct version.
 
>     3. I had to comment out the `_sib21` target because it gets passed the wrong rangen version with the double precision version of the random number generator. So somehow the cpp flag SIBYLL_SP is not passed around correctly after transitioning to *.fpp

I thought I can simplify the code here. Ninja and make are inconsistent here. cmake+Ninja always adds -cpp to all gfortran calls just like you did in the original Makefiles. cmake+Make does not add -cpp to all gfortran calls, but according to the gfortran docs this is done automatically when the file ending .fpp, so I thought for this particular file rangen.f, it is better to indicate that it needs preprocessing via the extension. I assume that you use a different gfortran version or an entirely different Fortran compiler which does not follow this rule. Previously, I had expressed this rule in cmake code and did not rely on the Fortran compiler. I will revert to the old solution.
 
> I have now more issues on manylinux2014 aarch64:
> 
>     1. I verified the previously tested version compiles.
> 
>     2. After checking out the latest, I can not make builds work any longer, neither for Ninja or Makefile backends.
> 
> 
> ```
>     [74/322] Linking Fortran shared module /app/build/lib.linux-aarch64-3.9/impy/models/_sib23c00.cpython-39-aarch64-linux-gnu.so
>     FAILED: /app/build/lib.linux-aarch64-3.9/impy/models/_sib23c00.cpython-39-aarch64-linux-gnu.so
>     : && /opt/rh/devtoolset-10/root/usr/bin/gfortran -fPIC -O3   -shared  -o /app/build/lib.linux-aarch64-3.9/impy/models/_sib23c00.cpython-39-aarch64-linux-gnu.so CMakeFiles/_sib23c00.dir/mnlxmpb/lib/python3.9/site-packages/numpy/f2py/src/fortranobject.c.o CMakeFiles/_sib23c00.dir/_sib23c00module.c.o CMakeFiles/_sib23c00.dir/_sib23c00-f2pywrappers.f.o CMakeFiles/_sib23c00.dir/src/sibyll/sibyll2.3c00.f.o CMakeFiles/_sib23c00.dir/src/sibyll/sibyll_init.f.o CMakeFiles/_sib23c00.dir/src/logging.f.o CMakeFiles/_sib23c00.dir/src/rangen.fpp.o  -lpython3.9 && :
>     /opt/rh/devtoolset-10/root/usr/libexec/gcc/aarch64-redhat-linux/10/ld: cannot find -lpython3.9
> ```

It is lacking the `-L<Python_libs_path>`. I think I understand the issue here and I can fix that. We need to add a runner for aarch64 to make sure this stuff does not break, but at the moment there are no images for this architecture and it is only possible to test this in emulation on GitHub, which is super slow. 

> On Apple Silicon mac, with numpy 1.23 parsing urqmd34 to pyf fails. Numpy < 1.20 is a bit of a problem because of the lack of a universal binary. So it doesn't matter at this moment. The build is quite fast (sans `urqmd34`) 173.64s user 24.35s system 370% cpu 53.500 total and SIBYLL2.1 builds as well.

The restriction Numpy < 1.20 is a problem in general, but this is something we cannot solve in this PR.

HDembinski:
> How do I run the cmake part only?

Inside the project directory, do
```
mkdir build
cd build
cmake ..
make -j8
```

HDembinski:
> For the source folder, I prefer either src/impy/models/fortran_sources or src/fortran_src or src/model_src. Something in this direction. The rangen.f and logging.f should be also there.

Can we settle for `src/fortran` ? `src/fortran_sources` is redundant and it is always good to avoid underscores as much as possible.

afedynitch:
> Can we settle for `src/fortran` ? `src/fortran_sources` is redundant and it is always good to avoid underscores as much as possible.

Fine with me.

afedynitch:
> I thought I can simplify the code here. Ninja and make are inconsistent here. cmake+Ninja always adds -cpp to all gfortran calls just like you did in the original Makefiles. cmake+Make does not add -cpp to all gfortran calls, but according to the gfortran docs this is done automatically when the file ending .fpp, so I thought for this particular file rangen.f, it is better to indicate that it needs preprocessing via the extension. I assume that you use a different gfortran version or an entirely different Fortran compiler which does not follow this rule. Previously, I had expressed this rule in cmake code and did not rely on the Fortran compiler. I will revert to the old solution.

It would be better not to use this special feature and ensure the previous cop development. DPMJET has cpp directives in many file because of specific features for FLUKA and CORSIKA. It would be hard to transition it to the same style. Currently the pyf is built on a cpp-ed single file of all sources because f2py is/was not reliable for parsing the directives correctly.

HDembinski:
> > I thought I can simplify the code here. Ninja and make are inconsistent here. cmake+Ninja always adds -cpp to all gfortran calls just like you did in the original Makefiles. cmake+Make does not add -cpp to all gfortran calls, but according to the gfortran docs this is done automatically when the file ending .fpp, so I thought for this particular file rangen.f, it is better to indicate that it needs preprocessing via the extension. I assume that you use a different gfortran version or an entirely different Fortran compiler which does not follow this rule. Previously, I had expressed this rule in cmake code and did not rely on the Fortran compiler. I will revert to the old solution.
> 
> It would be better not to use this special feature and ensure the previous cpp behavior. DPMJET has cpp directives in many file because of specific features for FLUKA and CORSIKA. It would be hard to transition it to the same style. Currently the pyf is built on a cpp-ed single file of all sources because f2py is/was not reliable for parsing the directives correctly.

I checked which files contain cpp instructions when I made these changes. We should only enable -cpp for those files where it is needed, not for all files. Running all files through the preprocessor even if they do not contain preprocessor commands is slowing things down and it can lead to weird compilation errors, like in case of SIBYLL which used a variable called `NDEBUG`, which cmake replaced with "" (since its adds the flag `-DNDEBUG` by default to release builds, which is correct for C/C++ files). Speed was important to you, so you need to make a decision.

Since I discovered and fixed this, I can add `-cpp` to all Fortran calls now, but I do not like to do this for the aforementioned reasons. As far as my limited understanding of Fortran source conventions go, Fortran files should be called .F or .fpp when they contain preprocessor instructions. Adding those to normal .f source is going against conventions.

HDembinski:
> It is lacking the `-L<Python_libs_path>`. I think I understand the issue here and I can fix that.

I looked into FindPythonLibsNew.cmake, which we need to use because the default FindPython.cmake shipped with cmake does not work for us. The difference between the two is that FindPythonLibsNew.cmake uses the Python interpreter to discover all the platform settings, which is what we need to reliably support building in virtual environments.

However, it looks like your platform does not set the variables that this script expects. Could you please run the following in your Python interpreter and tell me the output?

```
import sysconfig as s
print(s.get_config_var('LIBDIR'))
print(s.get_python_lib(plat_specific=True))
```

HDembinski:
> On Apple Silicon mac, with numpy 1.23 parsing urqmd34 to pyf fails. Numpy < 1.20 is a bit of a problem because of the lack of a universal binary. So it doesn't matter at this moment. The build is quite fast (sans urqmd34) 173.64s user 24.35s system 370% cpu 53.500 total and SIBYLL2.1 builds as well.

I merged the master, which updates/adds pyf files for all models. I use them in the build now, but unfortunately, that does not work for urqmd. I don't get why.
```
  # Using pre-generated source results in this error
  #  Undefined symbols for architecture x86_64:
  #  "_f2pyinitterms_", referenced from:
  #  _f2py_init_terms in _urqmd34module.c.o
```
This works for all other models that I already implemented.

HDembinski:
Since we are reviewing anyway, I update the state of this PR to "ready for review" although the PR is not finished yet

afedynitch:
> ```
>   # Using pre-generated source results in this error
>   #  Undefined symbols for architecture x86_64:
>   #  "_f2pyinitterms_", referenced from:
>   #  _f2py_init_terms in _urqmd34module.c.o
> ```
> 
> 
>     
>   
> 
> This works for all other models that I already implemented.
> 
> 
> This particular function is not even mentioned in the pyf file.
> 
> 
> **Update** Diff'ing the two pyf (the one in src and the generated), I find that the module `terms`
> is not included in `src/urqmd-3.4/urqmd34.pyf`

f2py sometimes has problems with modules. There is only a single module TERMS in urqmd. I replaced that module with a common block in my previous commit, so the old pyf won't work and you need a more recent one.

afedynitch:
> However, it looks like your platform does not set the variables that this script expects. Could you please run the following in your Python interpreter and tell me the output?
> 
> ```
> import sysconfig as s
> print(s.get_config_var('LIBDIR'))
> print(s.get_python_lib(plat_specific=True))
> ```

```
[root@b7db87ad1935 app]# python3.9
Python 3.9.13 (main, Aug  9 2022, 10:52:54) 
[GCC 10.2.1 20210130 (Red Hat 10.2.1-11)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sysconfig as s
>>> print(s.get_config_var('LIBDIR'))
/opt/_internal/cpython-3.9.13/lib
>>> print(s.get_python_lib(plat_specific=True))
  File "<stdin>", line 1, in <module>
AttributeError: module 'sysconfig' has no attribute 'get_python_lib'
```

Indeed. So it's the manylinux2014_aarch64 that doesn't do it for some reason.

HDembinski:
> Indeed. So it's the manylinux2014_aarch64 that doesn't do it for some reason.

It should be fixed now, try again with the latest version. It turns out that python3.9 on manylinux2010 and possibly other platforms do not have a python3.9.a library.

> f2py sometimes has problems with modules. There is only a single module TERMS in urqmd. I replaced that module with a common block in my previous commit, so the old pyf won't work and you need a more recent one.

Can then please for consistency also submit the correct .pyf? I changed all the other build steps to use the existing .pyf file.

afedynitch:
> Can you then please for consistency also submit the correct .pyf? I changed all the other build steps to use the existing .pyf file.

OK, the file is updated for consistency.

afedynitch:
Actually,... 
For typical builds, we actually can drop all the f2py dependencies. The fortran models, once properly implemented, won't change. So, we don't need to build the cmodule and fortranwrapper files everytime. Why not making another subfolder in `src/f2py`, and put all the pyf and fortranwrapper files into it. The build system should of course know how to build them.

afedynitch:
@HDembinski: Thanks for implementing the dpmjetIII191/3 builds. The problem with 191 is there is something CMake doesn't like about the parentheses. For 193 the build works fine because it has a different set of includes. The parentheses files are from FLUKA, so it can not be changed for the old version retrospectively. Is there maybe some kind of macro or replacement going on?

I removed the static lib, since for impy there is no need to generate it.

The build on mac CI has some problem with undefined symbols but I left my mac at home and cannot check what's going on. On linux/WSL the build passes and 15 of the tested models! What's going on with the linux runners? Are your other projects experiencing similar issues?

HDembinski:
> Actually,... For typical builds, we actually can drop all the f2py dependencies.

I think that's the way to go and this may also solve the numpy < 1.20 issue.

HDembinski:
> @HDembinski: Thanks for implementing the dpmjetIII191/3 builds. The problem with 191 is there is something CMake doesn't like about the parentheses.

Thanks for narrowing that down, I suspected this but I didn't get around to test my theory. I don't know why this is happening, this is the first time I build Fortran source at all. A workaround is to let CMake generate modified versions of the sources where the parentheses are removed in the build folder and then build using those. This will work for sure but it is not super elegant. More elegant would be to figure out why CMake does not like the parentheses. Perhaps the filenames just need to be escaped properly somewhere. But anyway, there is a way to go forward and we can always make it nicer later. 

> I removed the static lib, since for impy there is no need to generate it.

Sure, I just had to make it easier to debug the build problem.

> The build on mac CI has some problem with undefined symbols but I left my mac at home and cannot check what's going on.

Please send me the build logs when you can get them with the VERBOSE=1 flag, ie. `VERBOSE=1 pip install ...`

> On linux/WSL the build passes and 15 of the tested models!

Yes, almost done!

> What's going on with the linux runners? Are your other projects experiencing similar issues?

I have no idea, we should contact GitHub support. I never had such a persistent issue. I googled for this yesterday and this appears to happen rarely to some people, which then contacted support and they fixed it.

afedynitch:
> Please send me the build logs when you can get them with the VERBOSE=1 flag, ie. `VERBOSE=1 pip install ...`

You can see errors in the logs of the macos runner.

afedynitch:
> I have no idea, we should contact GitHub support. I never had such a persistent issue. I googled for this yesterday and this appears to happen rarely to some people, which then contacted support and they fixed it.

Will contact support.

afedynitch:
OK, so ubuntu-latest has fixed the issues with the linux runners. The test don't pass because the timeout of 15 seconds is too short and because the large files are not downloaded via git-lfs, see also #41

afedynitch:
Cool. Thanks for moving forward. I tried to test DpmjetIII193, but the test fails because of the IO common block. In fact, many symbols (all common blocks) are missing from the library.

HDembinski:
> Cool. Thanks for moving forward. I tried to test DpmjetIII193, but the test fails because of the IO common block. In fact, many symbols (all common blocks) of `_dpmjetIII193.so` are not exposed in Python.

Yes, I can confirm. I am using the pre-generated pyf file from `src/fortran/dpmjetIII-19.3/common/dpmjetIII193.pyf`, I only renamed the module to add the underscore prefix. If it works for you @afedynitch with this pyf file, then it is a problem with the generated -f2pywrappers.f and module.c. Could you please commit those files from your build that works to src/f2py?

To fix this properly, I have to imitate what you do in the makefile in cmake: first generate all the preprocessed sources and then let run f2py on those.

HDembinski:
@afedynitch I am very proud to say that dpmjetIII191 now also compiles and passes the test, too! However, dpmjetII193 is still failing the test and I don't know why.

I found a nice recipe to load/cache the lfs files for our tests in CI, but it does not seem to work. `git lfs pull` is executed but it looks like the files are still the links.

HDembinski:
## Status

Everything works, except the following:

DpmjetIII193: Unknown DPMJET version, IO common block not detected.
Phojet112: fails (probably linked to DpmjetIII193)
UrQMD34: fails
Pythia6: fails

HDembinski:
I started to understand the urqmd issue. urqmd contains f90 files, and in this case f2py generates f2py-wrappers.f and f2py-wrappers2.f90. The latter was always generated but since I did not expect it, this source file was not added to the compilation of the module.

Another issue is that I had added `-std=legacy` to the compilation of any Fortran source file, including f90 files. This does not work. I disabled this and now get more warnings. I am getting a new error now:
```
[ 44%] Building Fortran object CMakeFiles/_urqmd34.dir/_urqmd34-f2pywrappers2.f90.o
/usr/local/bin/gfortran -DNPY_NO_DEPRECATED_API=NPY_1_7_API_VERSION -D_urqmd34_EXPORTS -I/usr/local/Cellar/python@3.8/3.8.13_1/Frameworks/Python.framework/Versions/3.8/include/python3.8 -I/Users/hdembinski/Extern/impy/venv/lib/python3.8/site-packages/numpy/core/include -I/Users/hdembinski/Extern/impy/venv/lib/python3.8/site-packages/numpy/f2py/src -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX11.3.sdk -fPIC -Wno-uninitialized -fno-second-underscore -fallow-argument-mismatch -c /Users/hdembinski/Extern/impy/build/_urqmd34-f2pywrappers2.f90 -o CMakeFiles/_urqmd34.dir/_urqmd34-f2pywrappers2.f90.o
/Users/hdembinski/Extern/impy/build/_urqmd34-f2pywrappers2.f90:7:11:

    7 |       use terms, only : mass
      |           1
Fatal Error: Cannot open module file 'terms.mod' for reading at (1): No such file or directory
compilation terminated.
```
I noticed that there is a terms.mod in src/fortran/urqmd-3.4/sources, so I tried including the source directory. This seems to work indeed, `terms.mod` is found, but I am still getting a different error:
```
[ 45%] Building Fortran object CMakeFiles/_urqmd34.dir/_urqmd34-f2pywrappers2.f90.o
/Users/hdembinski/Extern/impy/build/_urqmd34-f2pywrappers2.f90:7:11:

    7 |       use terms, only : mass
      |           1
Fatal Error: Cannot read module file '/Users/hdembinski/Extern/impy/src/fortran/urqmd-3.4/sources/terms.mod' opened at (1), because it was created by a different version of GNU Fortran
```

afedynitch:
OK, so. I fixed most of the stuff. On my aarch64 mac all tests pass on Darwin and in manylinux2014, and now we can also support the recent numpy. So, although the issue #11 is mitigated, it is not entirely solved because to rebuild the interface or to add a new model, an old numpy version is still needed.

For UrQMD there were still confusions of some sources. As mentioned above, I replaced the TERMS modeule with a common block because it was used without meaning. Then without modules, there is no F90 wrappers. In the previous makefiles, there was a different set of FLAGS for f90. 

One issue with the data files and why the test don't pass is the length of the paths, which is limited in some of the codes to 256 or so. That might be part of the problem. Probably the best thing would be to symlink these files or move them to a shorter path.

afedynitch:
```
tests/test_new_interface.py::test_new_interface[DpmjetIII191] PASSED       [  5%]
tests/test_new_interface.py::test_new_interface[DpmjetIII193] PASSED       [ 10%]
tests/test_new_interface.py::test_new_interface[Phojet191] PASSED          [ 15%]
tests/test_new_interface.py::test_new_interface[Sibyll21] PASSED           [ 21%]
tests/test_new_interface.py::test_new_interface[DpmjetIII306] PASSED       [ 26%]
tests/test_new_interface.py::test_new_interface[EposLHC] PASSED            [ 31%]
tests/test_new_interface.py::test_new_interface[Sibyll23d] PASSED          [ 36%]
tests/test_new_interface.py::test_new_interface[Sibyll23c02] PASSED        [ 42%]
tests/test_new_interface.py::test_new_interface[Sibyll23c00] PASSED        [ 47%]
tests/test_new_interface.py::test_new_interface[QGSJetII04] PASSED         [ 52%]
tests/test_new_interface.py::test_new_interface[Sibyll23] PASSED           [ 57%]
tests/test_new_interface.py::test_new_interface[QGSJet01c] PASSED          [ 63%]
tests/test_new_interface.py::test_new_interface[Phojet112] PASSED          [ 68%]
tests/test_new_interface.py::test_new_interface[Sibyll23c04] PASSED        [ 73%]
tests/test_new_interface.py::test_new_interface[Sibyll23c03] PASSED        [ 78%]
tests/test_new_interface.py::test_new_interface[Sophia20] PASSED           [ 84%]
tests/test_new_interface.py::test_new_interface[Sibyll23c01] PASSED        [ 89%]
tests/test_new_interface.py::test_new_interface[Sibyll23c] PASSED          [ 94%]
tests/test_new_interface.py::test_new_interface[QGSJetII03] PASSED         [100%]

============================== 19 passed in 20.53s ===============================

just for your entertainment...
```

jncots:
I am currently resolving conflicts

HDembinski:
> 



> One issue with the data files and why the test don't pass is the length of the paths, which is limited in some of the codes to 256 or so. That might be part of the problem. Probably the best thing would be to symlink these files or move them to a shorter path.

No, I would draw a line here. We can try to accommodate some idiosyncrasies of old code, but if the buffers for path lengths are too short, the buffers need to be enlarged by modifying the source code. This can be done by applying source patches in the cmake build, similar to what I am doing for dpmjetIII191.

HDembinski:
> For UrQMD there were still confusions of some sources. As mentioned above, I replaced the TERMS modeule with a common block because it was used without meaning. Then without modules, there is no F90 wrappers. In the previous makefiles, there was a different set of FLAGS for f90.

I don't understand, I copied the urqmd34.pyf from the working manylinux build using @jncots branch, and it had the bit about terms.

HDembinski:
@afedynitch The new src/f2py *module.c files that you committed do not work anymore with older numpy versions. This indicates that *module.c should be generated on the fly and we cannot use static files.

HDembinski:
I confirm that `test_new_interfaces.py` passes on my machine: Intel Mac, cpython38, numpy 1.23.2, no manylinux, just a normal venv. Congratulations @afedynitch @jncots !

Now, if we only could get that git lfs working on CI, we would be able to see this success in CI as well.

HDembinski:
> How old? It runs on 1.19.3 on x64 as well. Just checked this

numpy-1.18.5

HDembinski:
Some implementation notes:

I added an underscore to the names of the fortran wrapper modules for two reasons. These modules logically belong in `impy/models`, but then the names of some Python and Fortran modules collide (e.g. sophia.py and sophia<stuff>.so). Secondly, it is good practice to start everything that is an implementation detail and my change in the future with `_`, this communicates to the user that these objects are private and their API is not stable.

afedynitch:
> > How old? It runs on 1.19.3 on x64 as well. Just checked this
> 
> numpy-1.18.5

Do we need to support this? I mean the final `*.so` don't link numpy libs. On the same note, I think that CMAKE should know how to build the interface files. For the commit above I used the makefiles to build them because I didn't dig into the cmake F2Py function to do this. It looked to me that it should work.

HDembinski:
> Do we need to support this? I mean the final `*.so` don't link numpy libs.

This experiment shows that *module.c depend on the numpy version. It may work for now but break in the future. I only went along with your suggestion to not generate these files because we believed they are stable, but they are apparently not. This means we should generate them to be future-proof. The pyf files are truly stable, so it is fine to keep them around in src/f2py, but apparently not the *module.c files. We can still keep those *module.c files as a workaround where that's needed.

> On the same note, I think that CMAKE should know how to build the interface files.

I agree and it mostly generates the interface files just fine. Some more work is needed to fix those cases where it does not work. I assume that the errors go away when I imitate what the makefile does, first CPP preprocess all sources and then apply numpy.f2py.

> For the commit above I used the makefiles to build them because I didn't dig into the cmake F2Py function to do this. It looked to me that it should work.

afedynitch:
> I agree and it mostly generates the interface files just fine. Some more work is needed to fix those cases where it does not work. I assume that the errors go away when I imitate what the makefile does, first CPP preprocess all sources and then apply numpy.f2py.

To generate the interface files you need only the pyf file, nothing else `python -m numpy.f2py XXX.pyf`. So it should be easy to do this.

HDembinski:
> To generate the interface files you need only the pyf file, nothing else python -m numpy.f2py XXX.pyf. So it should be easy to do this.

Ok, that's true, but apparently when I run `python -m numpy.f2py XXX.pyf` I get a different result than you do in some cases with the old makefiles, no?

HDembinski:
To summarize, I am aware of the following differences between the old and new build system:
- The old system applied different compiler flags for .f and f90 files. This only affects urqmd, but it may be important there. I can easily fix that, if this turns out to be important.
- The old system first CPP preprocessed some sources before running numpy.f2py on them to generate the module source. I am not doing this at the moment. As Anatoli pointed out, this should only affect the generation of the .pyf files, which we currently ship with the code. So in principle the build should work also without pre-generated module.c and f2pywrapper.f.

afedynitch:
> Ok, that's true, but apparently when I run `python -m numpy.f2py XXX.pyf` I get a different result than you do in some cases with the old makefiles, no?

I think it really depends on the exact version and platform. So there might be minor differences but I wouldn't worry about these except if symbols are missing. I hope that this is not the step that kills the export of symbols with numpy>1.20 versions.

afedynitch:
> @afedynitch @jncots Do you want to try to fix Pythia6 and UrQMD34 before merging this branch, or shall we merge and address those other models in a separate PR?

Let’s do one attempt tomorrow morning (ours) to get Pythia on board. UrQMD was actually working, I believed. Haven’t checked if the test runs and if not why not.

HDembinski:
Yes, let's try to fix the other models then. I commented urqmd34 back in, but it still fails. I suppose that's related to the differences between .f and .f90. I think URQMD is the one one where we need to compile f90 files.

```
tests/test_new_interface.py:61: TimeoutError
------------------------------------------------------------------------------------------ Captured stdout call ------------------------------------------------------------------------------------------
 no collision problem in UrQMD
```

HDembinski:
I don't get why git lfs is not working in CI, I am giving up on that for now.

afedynitch:
I fixed Pythia and UrQMD to pass the tests. Pythia needs more work in replacing the random number generator. @jncots seems to be on it.  I think this fix can be part of #44.

UrQMD has a problem with pion projectiles. But it is new for me that it doesn't work at all. #45 will take care of this. 

> I don't get why git lfs is not working in CI, I am giving up on that for now.

Let's merge without this after a last review. Although it's a kinda sad that it's the only thing why tests don't pass :)

jncots:
I changed random generator in pythia6 for `make` build system. I will try to do this in `cmake`

src/fortran/urqmd-3.4/sources/init.f
```diff
@@ -342,7 +342,6 @@ subroutine init
 c    &           bdist=nucrad(Ap)+nucrad(At)+2*CTParam(30)
 c hjd1
  215        bimp=sqrt(ranf(0))*bdist
-cdh         if (bimp<bmin) goto 215
             if (bimp .lt.  bmin) goto 215
```
HDembinski, line 345: wtf?!

src/impy/models/pythia6.py
```diff
@@ -66,9 +66,9 @@ def children(self):
     @property
     def _charge_init(self):
         if self.charge_vec is None:
-            self.charge_vec = [
+            self.charge_vec = np.asarray([
                 self.lib.pychge(self.lib.pyjets.k[i, 1]) / 3 for i in range(self.npart)
-            ]
+            ])
```
HDembinski: Perhaps use np.fromiter to avoid creating a temporary array.

src/fortran/urqmd-3.4/sources/init.f
HDembinski: No suggestion to change anything, just a comment of my surprise to see this kind of code.

tests/test_new_interface.py
```diff
@@ -0,0 +1,66 @@
...
+        ecm=7 * TeV,
+        p1pdg=p1pdg,
+        p2pdg=p2pdg,
+    )
+
+    # Some models need to initialize same fortran code,
+    # which can only be initialized once, therefore run
+    # in separate thread
+    # maxtasksperchild=1 is needed to prevent multiproc to recycle workers and take for each task a fresh interpreter.
+    with Pool(1, maxtasksperchild=1) as p:
```
HDembinski, line 55: This is not necessary, so I am inclined to revert this.
1) The pool only exists to create one job and then destructs. So there is only ever one task executed per child.
2) It was working fine without.

afedynitch: Right. That's different in how Pool is used test_all_cms. I'm fine with reverting..

src/impy/models/pythia6.py
afedynitch: OK! Thanks

tests/test_new_interface.py
HDembinski: One reason to leave it is as a reminder that  maxtasksperchild=1 is necessary in general for Pool, to prevent copy/paste errors from this test in the future.

HDembinski: Ok, let me just add that as a comment and then leave it as is

HDembinski:
I am a bit surprised that Pythia6 does not produce an error. The test was still using pi- + proton and Pythia6 does not support pi- as beam particles AFAIK.

afedynitch:
> I am a bit surprised that Pythia6 does not produce an error. The test was still using pi- + proton and Pythia6 does not support pi- as beam particles AFAIK.

Pythia6 can run pi- proton:

```
****************** PYINIT: initialization of PYTHIA routines *****************

 ==============================================================================
 I                                                                            I
 I             PYTHIA will be initialized for a pi- on p+ collider            I
 I                  at   7000.000 GeV center-of-mass energy                   I
 I                                                                            I
 ==============================================================================
```

What is your error?

afedynitch:
> Let's merge without this after a last review. Although it's a kinda sad that it's the only thing why tests don't pass :)

Why I said this was that for getting binary downloads from github releases, we need to make a release first to have the links to the file(s).

HDembinski:
> What is your error?

There is no error, I was surprised that it can do pi + X. I recently talked to Marius Utheim who is working on integrating generic hadron-hadron collisions in Pythia8, so I didn't expect Pythia 6 to support this.

HDembinski:
> > Let's merge without this after a last review. Although it's a kinda sad that it's the only thing why tests don't pass :)
> 
> Why I said this was that for getting binary downloads from github releases, we need to make a release first to have the links to the file(s).

Yes, makes sense. However, I read that LFS files included in the release will count towards your LFS bill. In CI we can cache the large files a little bit to reduce the number of downloads, but this will create some traffic.

afedynitch:
> Yes, makes sense. However, I read that LFS files included in the release will count towards your LFS bill. In CI we can cache the large files a little bit to reduce the number of downloads, but this will create some traffic.

I know, but the I thought the plan is to take them out from lfs and just use a binary unversioned arxiv that's attached to a release and gets a permanent link. Then the problem is gone and I can stop paying for lfs. Caching some files would still be good than asking the CI to download and unpack 0.5GB of data every time.

afedynitch (approved):
The PR has been extensively review by all members. It can be merged despite CI failing due to binary file downloads from `git lfs`. We'll fix it in the next PR.

It was a tough one. Thanks!

Issue #32, opened 2022-08-06
EventKinematics


HDembinski:
The `__init__` is not well designed. It is too flexible and thus confusing for the user.

A better design is to have several subclasses, which have EventKinematics as the common case class. The models use the interface of the base class, while the user uses one of the derived classes to set the parameters. There is one class for CenterOfMass and one for FixedTarget, and possibly a generic one. Users should not be allowed to create instances of EventKinematics directly. This can be achieved by calling an abstract method in the `__init__.py` that is only implemented in the derived classes.

```py
class CenterOfMass(EventKinematics):
     def __init__(self, energy, p1, p2): # energy in CM frame
             ...

class FixedTarget(EventKinematics):
    def __init__(self, energy, p1, p2): # energy in lab frame
```

This is self-documenting and should also set the frame for the output.

`p1` and `p2` should be very flexible and allow the user to specify a particle via 
- pdg id (if int)
- name (if string, for lookup please use `particle` library)
- as a nucleus with (A, Z) (if tuple).

Issue #31, opened 2022-08-06
impy_config needs to go


HDembinski:
impy_config is used as a modifiable global state that is used in several places. This is bad, because changing impy_config in some place can have unexpected side-effects elsewhere. It introduces a hidden coupling of all classes to each other, which is impossible for an outsider to understand. This bad pattern is why ROOT is such a pain to work with, and goes against the principles of OOP.

Instead of a global config that is shared between all models, each model should accept its own configuration in the `__init__`. It is then up the user whether they want to use the same config instance for two models or different independent instances.

HDembinski:
Python has keyword arguments, which allows one to pass configuration and set default values in an elegant way. This is how professional libraries like matplotlib do it.

Pseudo-code
```py
class Sibyll:
    def __init__(self, event_kinematics, *, debug_level=0, foo_or_bar="foo", ...):
```

afedynitch:
The motivation behind a global state is that if one tries to debug something physics wise, one needs to be able to switch on and off debugging info for some specific events or change some parameters globally, like the condition for stable particles. So a global shared state is kinda intended.

For the generator specific settings there is no need for a global shared state, but then each specific setting needs a doc or a setter/getter. These are probably very few items, though.

We can consider moving this forward but I recommend going for this issue after the updated API is stable and some more urgent technical issues are fixed.

HDembinski:
Yes, let's postpone this, but I want to clarify one point. I am just against having a **single global dict**. I am ok with each model having its own local dict, which ideally should be attached to the model instance, but since we can only generate a single instance of each model anyway, it can also be attached to the class.

 I think it is fine that each generator has a flexible dict with configurations, which is for advanced usage and there is no API provided for changing this dict. My concern is only the coupling which this global state creates between all models in the library.

afedynitch:
I agree that it has to go. There now already a design issue with a global config when looking at #43 where it would actually result in faulty behavior if one would initialize a model with `FixedTarget` and then a different one with `CenterOfMass`. The first frame would be overwritten by center of mass.

In fact there is only a very small list of arguments that would need some sort of global setting, if any. It needs some thoughts though.

HDembinski:
I plan to work on this, because I have a very clear plan in my mind how to remove impy_config without restricting the opportunities it offers.

Briefly: the configuration options in impy_config will become fields of the MCRun base class, so that they are specific to the generator instance. We further make MCRun a singleton so that only one instance can exist per process.

PR #43, opened 2022-08-16
Feature/comptarget integration

jncots:
Currently  the `CompositeTarget` class was implemented but haven't been integrated into the `EventKinematics`.  Here the subclass `CompositeTargetKinematics` is created to work with composite targets like:

```python
import impy
from impy.kinematics import CompositeTarget
from impy.kinematics import EventKinematics
from impy.constants import TeV, GeV


air = CompositeTarget("Air")

air.add_component("Nitrogen", 14, 7, 2 * 0.78084)
air.add_component("Oxygen", 16, 8, 2 * 0.20946)
#air.add_component("Argon", 40, 18, 0.00934) # Sibyll doesn't work with nuclei A > 20 
air.add_component("Oxygen(Vapor)", 16, 8, 0.004)
air.add_component("Hydrogen(Vapor)", 1, 1, 2 * 0.004)

# Print information
print(air)

impy.impy_config["user_frame"] = "laboratory"

ekin = impy.kinematics.CompositeTargetKinematics(elab = 13 * TeV, p1pdg = 2212,
                                                 composite_target = air)
generator = impy.models.Sibyll23d(ekin)

for event in generator(1000):
    print("Nevent = ", generator.nevents)
```

Currently the example works only with Sibyll. Some other modules, e.g. DpmjetIII191, should be fixed by changing the inialization way, because they use event kinematics parameters for the ranges in table calculations.


jncots:
Now the models which use `CompositeTargetKinematics` are initialized with nuclei with maximum A, and `margin_factor = 2` is added for passed energy to have reserve in energy. `DpmjetIII191` works now.

HDembinski:
This needs a redesign. EventKinematics mixes two orthogonal concepts: specifying the frame of the interaction, cms or fixed target, and specifying the types of colliding particles. Orthogonal concepts need to be handled by orthogonal components. See #32 for more information.

HDembinski:
In other words, I propose a design which allows one do these things:

```py
from impy.eventkinematics import CompositeTarget, CenterOfMass, FixedTarget

# particles can be specified by PDG ID, name (based on particle library), or as a tuple (Nucleons, Protons)
# CenterOfMass sets the output frame to center-of-mass
ekin1 = CenterOfMass(13 * TeV, 2212, 2212)
ekin2 = CenterOfMass(13 * TeV, "proton", "proton")
ekin3 = CenterOfMass(13 * TeV, "pi-", "N")
ekin3 = CenterOfMass(13 * TeV, "pi-", (12, 6))

# FixedTarget supports the same flexible particle definitions
ekin4 = FixedTarget(10 * PeV, "proton", (12, 6))

air = CompositeTarget() # no need for a name, should be optional
air.add_component("Nitrogen", 14, 7, 2 * 0.78084)
air.add_component("Oxygen", 16, 8, 2 * 0.20946)
air.add_component("Oxygen(Vapor)", 16, 8, 0.004)
air.add_component("Hydrogen(Vapor)", 1, 1, 2 * 0.004)

# Both CenterOfMass and FixedTarget should support composite targets
ekin5 = FixedTarget(10 * PeV, "proton", air)
```

afedynitch:
> ```python
> ekin = impy.kinematics.CompositeTargetKinematics(elab = 13 * TeV, p1pdg = 2212,
>                                                  composite_target = air)
> ```

I also think that this is not ok. It should be passed as a normal parameter and not via subclassing.

Subclassing the output frame would be not bad and avoid errors by changing the frame accidentally in the global config (see #31 ). 

If that can be done by passing the output frame config option to init and check if it has changed during run? We have to test right now to catch mistakes from boosts and it would be tough to work through this without tests. We (mostly @kotania actually) took ages to figure out the correct settings for UrQMD, so the flags set are `sensitive parameter choices`.

jncots:
Now `EventKinematics` is  the abstract class with two subclasses `CenterOfMass` and `FixedTarget`, which can accept `pdg`, `string`, `tuple` and `composite_target`. `string` handling is not implemented yet.

HDembinski:
> Now `EventKinematics` is the abstract class with two subclasses `CenterOfMass` and `FixedTarget`, which can accept `pdg`, `string`, `tuple` and `composite_target`. `string` handling is not implemented yet.

Cool! For string handling, I recommend `particle`, which is part of/maintained by the Scikit-HEP project. The following code does a name lookup:
```py
from particle import literals

pdgid = getattr(literals, name).pdgid
```
There are other ways, but using the literals is convenient because they contain the commonly referred to particles.

HDembinski:
Thinking more about this:
```py
air = CompositeTarget() # no need for a name, should be optional
# also here, names should be optional, particle species can be defined exactly as in CenterOfMass and FixedTarget
air.add_component((14, 7), 2 * 0.78084)
air.add_component("O", 2 * 0.20946) 
air.add_component((16, 8), 0.004, name = "Oxygen(Vapor)")
air.add_component((1, 1), 2 * 0.004, name = "Hydrogen(Vapor)")
```
Another important design rule for classes is to avoid partial initialization. Once created, a class instance should be in a fully valid state. This is not the case for CompositeTarget in this design, which starts as an invalid empty target. Also, in this design the object has to compute the relative fractions again repeatedly whenever a component is added instead of doing it once.

So, based on these insights, we see that the composite should be completely defined in the constructor:
```
air = CompositeTarget([('N', 0.8), ('O', 0.2)])
```
The class should either check that the fractions add up to one and normalize them if they don't. Users can still create the collection to pass to the CompositeTarget class incrementally if they want, by starting with an empty list and then appending to it. Why is this better than the design before? A list is just a collection of items, it does not have to maintain the special invariant that the CompositeTarget has to maintain.

HDembinski:
Is it important to name the components of the CompositeTarget? I think there is no difference between oxygen and oxygen vapor as far as impy is concerned, or am I missing something?

afedynitch:
> Is it important to name the components of the CompositeTarget? I think there is no difference between oxygen and oxygen vapor as far as impy is concerned, or am I missing something?

No need for names. But there is a print function to list a table, so it should be at least optional. Vapor or not has no impact, indeed.

jncots:
Just a note: currently handling of strings is so that if the name of the particle is not in `common_particles` list the lookup could be time consuming (0.2 sec per lookup). I will try to fix this using `literals` as suggested by [HDembinski](https://github.com/HDembinski).

HDembinski:
> Just a note: currently handling of strings is so that if the name of the particle is not in `common_particles` list the lookup could be time consuming (0.2 sec per lookup). I will try to fix this using `literals` as suggested by [HDembinski](https://github.com/HDembinski).

I doubt that using particles is faster, but speed is not the issue. We do this lookup only once, so it does not matter how fast it is.

 If you have a mapping in impy from common particles names to pdgids that is also fine with me! It does not have to be particles. The advantage of particle is that we do not have to maintain this and if particle becomes widely used (to which we can contribute), the names established by particle will become standard names that people expect.

So it is more a maintenance/usability argument to use particle.

jncots:
> I doubt that using particles is faster, but speed is not the issue.

Yes, indeed. I found that the main problem with speed because of `Particle.from_string()` or `Particle.from_name()` if the particle name is not in `common_particles`.  `literals` doesn't help with it. So I will leave the current implementation unless there is other way how to search pdg_id via string name.

jncots:
> No need for names. But there is a print function to list a table, so it should be at least optional. Vapor or not has no impact, indeed.

In principle, oxygen from both component (air and vapor) could be merged. It just requires additional change in the code. Probably it will not affect performance.  On the other hand, the current implementation allows you to see the components when printed.

jncots:
Currently the interface is following:

```python
import impy
from impy.kinematics import CompositeTarget
from impy.kinematics import FixedTarget, CenterOfMass
from impy.constants import TeV, GeV

air = CompositeTarget(
    [
        ("N14", 2 * 0.78084, "Nitrogen"), # by name
        ((16, 8), 2 * 0.20946, "Oxygen"),   # by tuple
        ("Ar40", 0.00934),                # without name string
        ("O16", 0.004, "Oxygen(Vapor)"),
        ("proton", 2 * 0.004, "Hydrogen(Vapor)")
    ], 
    "Air" #optional name for target (for table printing, e.g. print(air))
)

# Works as well
air.add_component("N14", 2 * 0.78084, "Nitrogen")

generator = impy.models.Sibyll23d(CenterOfMass(13 * TeV, "proton", air))

for event in generator(1000):
    ...
```

Now lookup for particle names is fast, because instead of `common_particles`  dictionary defined in `particle` library, I define and use the  dictionary `all_pdgs` containing all particle names.

impy/kinematics.py
```diff
@@ -81,22 +166,27 @@ def sort_list(k, idcs):
...
         self.component_name = sort_list(self.component_name, idcs)
 
+    def get_maximum_AZ(self):
+        a_val = self.component_A
+        max_ind = a_val.index(max(a_val))
+        return self.component_A[max_ind], self.component_Z[max_ind]
+
     def get_random_AZ(self):
         """Return randomly an (A, Z) tuple according to the component fraction."""
         from numpy.random import choice
```
HDembinski, line 177: Should these be public or private methods? I think they are private, no? By default, we make everything private.

impy/kinematics.py
```diff
@@ -144,17 +251,21 @@ def __init__(
         beam=None,
         nuc1_prop=None,
         nuc2_prop=None,
+        particle1=None,
+        particle2=None,
     ):
```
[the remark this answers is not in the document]
HDembinski: Sorry about changing my mind, I suggested the `_init_guard`, but this does not produce a nice error message that tells the user what to do.

impy/kinematics.py
```diff
@@ -109,7 +199,20 @@ def __repr__(self):
         return ostr
 
 
-class EventKinematics(object):
+def get_particle_input_type(arg):
+    if isinstance(arg, int):
```
HDembinski, line 203: I believe this is also a private function so should start with `_`

impy/kinematics.py
```diff
@@ -183,6 +305,19 @@ def __init__(
...
+        if particle2:
+            particle_input_type = get_particle_input_type(particle2)
+            if particle_input_type == "pdg_id":
+                p2pdg = particle2
+            elif particle_input_type == "tuple":
+                nuc2_prop = particle2
+            elif particle_input_type == "string":
+                p2pdg = FromParticleName.get_pdg(particle2)
+            elif particle_input_type == "composite_target":
+                nuc2_prop = particle2.get_maximum_AZ()
```
HDembinski, line 317: There is a lot of code duplication here. I would introduce a function `_normalize_particle` (or similar name) which takes the allowed kinds of inputs and returns either a pdgid or a CompositeTarget. Nuclei can be mapped to pdgids. This simplifies the handling downstream and its reduces the code here to just two calls to this function.

HDembinski (commented):
This looks great! My comments are to improve the implementation and style.

impy/kinematics.py
```diff
@@ -144,17 +251,21 @@ def __init__(
...
+        assert np.sum(np.asarray([p1pdg, nuc1_prop, particle1], dtype="bool")) == 1, (
+            "Define either particle id, "
+            + "nuclear properties or particle1 for side 1."
         )
-        assert p2pdg or nuc2_prop, (
-            "Define either particle id or " + "nuclear properties for side 2."
+        assert np.sum(np.asarray([p2pdg, nuc2_prop, particle2], dtype="bool")), (
+            "Define either particle id, "
+            + "nuclear properties or particle2 for side 2."
         )
```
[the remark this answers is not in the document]
HDembinski: It took me a while to understand this myself, but `assert` is more an executable comment. It documents that the class has certain invariants (that's the jargon) which are upheld. For example, if you made a class that represents only positive numbers, then an assert at the end of the `__init__` could document this.
```
class PositiveNumber:
    def __init__(self, value):
          if value <= 0: raise ValueError("only positive numbers allowed")
          self._value = value
          assert self._value > 0
```
The assert looks superfluous, but if you think of it like a comment to another programmer, it makes sense. Especially in a non-trivial piece of code. If for some reason the input validation does not work as intended (a bug), you will trigger the assert instead of getting some other confusing error downstream in the code.

impy/kinematics.py
```diff
@@ -17,8 +17,71 @@
...
+        {p.programmatic_name: int(p.pdgid) for p in particle.Particle.findall()}
+    )
+    all_pdgs.update(
+        photon=22,
+        Higgs=25,
+        proton=2212,
+        antiproton=-2212,
+        neutron=2112,
+        antineutron=-2112,
+    )
```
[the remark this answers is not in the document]
jncots: Thank you. I checked: `pi-` and `pi+` are already in the dictionary.

impy/kinematics.py
```diff
@@ -17,8 +17,71 @@
 
 import six
```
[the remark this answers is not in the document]
jncots: After the merge with master it is removed.

impy/kinematics.py
```diff
@@ -17,8 +17,71 @@
...
+        return True
+    else:
+        return False
+
+
+class FromParticleName:
+    all_pdgs = {p.name: int(p.pdgid) for p in particle.Particle.findall()}
+    all_pdgs.update(
+        {p.programmatic_name: int(p.pdgid) for p in particle.Particle.findall()}
+    )
```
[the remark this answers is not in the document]
jncots: It is added

impy/kinematics.py
```diff
@@ -17,8 +17,71 @@
...
 import six
 import numpy as np
-from impy import pdata
+from impy import pdata, impy_config
 from impy.util import info
+import abc
+import particle
+
+
+def is_AZ_tuple(arg):
```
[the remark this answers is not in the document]
jncots: Changed

impy/kinematics.py
```diff
@@ -36,26 +99,37 @@ def __init__(self, label=""):
...
+                self._add_component(component[0], component[1], component[2])
+            else:
+                raise ValueError(
+                    "'CompositeTarget': wrong component length = {0} for {1}".format(
+                        len(component), component
+                    )
+                )
+        self._normalize()
+
+    def _add_component(self, AZ, fraction, name=""):
```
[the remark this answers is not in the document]
jncots: Changed to `az`

impy/kinematics.py
jncots: These methods are public because they are for use outside of `_FromParticleName` class

impy/kinematics.py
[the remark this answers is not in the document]
jncots: It is done as suggested. ` ValueError` returns clear message.

impy/kinematics.py
jncots: Yes. Changed.

impy/kinematics.py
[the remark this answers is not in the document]
jncots: All `assert`s are changed to exceptions.

impy/kinematics.py
jncots: I agree. It is changed to use `_normalize_particle` to remove duplication.

impy/kinematics.py
HDembinski: Ok, but they should still be private, I think. In Python, "private" means a different thing to C++. The "private" parts of a library at not something the user should call, but they can still be "public"ly used within the library. So even if these methods are called by another impy code, they can still start with an underscore. The question is really whether it makes sense for users to call these methods. To me it looks like it is not something we want to expose to the users, because these methods are probably used to randomly sample from the CompositeTarget, so it is an implementation detail.

jncots: I see. User should not use these methods. They are changed to private

jncots:
There were probably a typo at line 334 of `kinematics.py`. `self.p1pdg = 2212`  to `self.p1pdg = 2112` has been changed.

src/impy/models/sophia.py
```diff
@@ -142,22 +142,28 @@ def _set_event_kinematics(self, event_kinematics):
...
+                + "but received nucleus = ({0}, {1})".format(k.A2, k.Z2)
             )
-            raise Exception("Input error")
 
         self.nucleon_code_number = self.lib.icon_pdg_sib(k.p2pdg)
         self.lib.initial(
             self.nucleon_code_number
         )  # setting parameters for cross-section
-        self.energy_of_nucleon = k.pmass2
+        self.energy_of_nucleon = np.float32(k.pmass2)  # fix roundoff error
```
HDembinski, line 166: I don't understand, can you elaborate what is the problem? np.float32 has lower resolution than the standard Python `float` which is `np.float64`?

HDembinski (approved):

HDembinski:
@jncots If you like, you could merge in master now to get the new Event and the CI fixes. This should make the unit tests work in principle, although they will still fail since you made API changes. If you adapt the failing tests to the new API (just replace EventKinematics with CenterOfMass), the test should pass.

src/impy/models/sophia.py
afedynitch: This workaround is required because there is a subtraction of a small number in sophia and roundoff errors matter. Sophia initializes a double precision mass array with single precision literals, i.e. `/0.9378/' instead of `/0.9378D0/`. So it compares a true DP representation of a number with that of a DBLE cast. To not change stuff in sophia inconsistently we just pass the argument in a way that is doing the same. I preferred this solution compared to hacking sophia.

jncots:
I merged master and replaced EventKinematics with CenterOfMass. It is ready to merge with master.

afedynitch:
Is there a test?

afedynitch (commented):
There two issues remaining from my side:

1. The composite target as it is can not be tested deterministically. It needs to either use the random generator of the model, or use a python random number generator with a fixed seed. 
2. There should be a unit test that goes through all models and marks those that are expected to fail with xfail or so, so the it also tests that the Exception is raised correctly.

src/impy/models/sophia.py
HDembinski: Ok, that makes sense! Could you please add a more elaborate comment to explain to future readers of the code? This is a rather subtle issue and with your longer explanation I can actually understand why it has to be like this. You basically want to make sure that the calculation inside Sophia and in Python give bit-identical results, even though this change reduces the accuracy of self.energy_of_nucleon, right?

HDembinski:
>     1. The composite target as it is can not be tested deterministically. It needs to either use the random generator of the model, or use a python random number generator with a user-defined seed.

I haven't reviewed the API in that detail, but you can test something that using pseudorandom numbers deterministically by fixing the seed of the RNG. This will work until the implementation of the RNG changes, which happens rarely. So instead of having a test fails randomly (bad on CI) you have something that possibly fails only when the RNG changes. I usually go for that option and it works fine.

If you don't want to risk that, then the other option is to mock the RNG. Instead of the real RNG you put in a fake RNG that produces a sequence that you know. In that case, your test will not depend on external changes in RNGs. Mocking components that would otherwise produce unreliable inputs is standard in unit testing. Typically, this is used in code that talks to servers. The mock up then fakes server responses. This is important also to test scenarios which do not occur normally, i.e. to see whether the code can properly react to a server timeout or a server that sends incorrect data.
 
>     2. There should be a unit test that goes through all models and marks those that are expected to fail with xfail or so, so the it also tests that the Exception is raised correctly.

To test exceptions, use
```py
with pytest.raises(MyException):
      my_function_that_raises()
```
You can also test whether warnings are generated in a similar way.

jncots:
The tests for `_get_pdg`, `_get_AZ` from strings  and `get_random_AZ` from CompositeTarget are in `test_get_particles.py`. CompositeTarget uses now `numpy.random.default_rng()` which allows to pass a seed and run tests for the seed.

The tests for generators with CompositeTarget are in `test_comp_target.py`. They are practically identical with tests in `test_generators.py` with substitution of CompositeTarget. 

To make testing complete it is needed to record the outcome of `pid` array for specific seed for each models and test them agains it.

I add, for a time being to `kinematics.py`, the `lorentz_boost` function which transforms arrays of 4-vectors to arbitrary frame. I think it could be useful somewhere.

The pull request could be closed, if there is no other suggestions.

src/impy/kinematics.py
```diff
@@ -26,7 +89,10 @@ class CompositeTarget(object):
     Examples of such composite targets are Air, CO_2, HCl, C_2H_60.
     """
 
-    def __init__(self, label=""):
+    from numpy.random import default_rng
+
+    def __init__(self, component_list, label=""):
+        self.rng = self.default_rng()
```
HDembinski, line 95: Please add an optional keyword `random_state` which allows the user to pass their own random number generator. If `random_state` is None, use `np.random.default_rng()`. If the user provides an int, use that to seed the `np.random.default_rng`. See https://github.com/scikit-hep/resample/blob/develop/src/resample/bootstrap.py#L34 for an example. This is how scipy does it.

HDembinski: It is important to be able to set the seed, because we want deterministic randomness. This is even more flexible, because it allows users to pass their preferred random number generator instead of the numpy default. Moreover, it allows one to pass a mocked RNG for unit tests.

src/impy/kinematics.py
```diff
@@ -80,22 +173,28 @@ def sort_list(k, idcs):
         self.component_Z = sort_list(self.component_Z, idcs)
         self.component_name = sort_list(self.component_name, idcs)
 
+    def get_maximum_AZ(self):
+        a_val = self.component_A
+        max_ind = a_val.index(max(a_val))
+        return self.component_A[max_ind], self.component_Z[max_ind]
+
+    def set_rng_seed(self, seed):
+        self.rng = self.default_rng(seed)
```
HDembinski, line 182: The seed and the RNG should be only set in the constructor. Please remove.

src/impy/kinematics.py
```diff
@@ -16,8 +16,71 @@
...
+
+    @staticmethod
+    def _get_pdg(pname):
+        try:
+            return _FromParticleName.all_pdgs[pname]
+        except KeyError:
+            raise ValueError("Particle with name = {0} is not found".format(pname))
+
+    @staticmethod
+    def _get_AZ(arg):
```
HDembinski, line 70: No change necessary, because it is not important, but since the whole class is already marked as private, you don't have to also mark its methods as private.

src/impy/kinematics.py
```diff
@@ -80,22 +173,28 @@ def sort_list(k, idcs):
         self.component_Z = sort_list(self.component_Z, idcs)
         self.component_name = sort_list(self.component_name, idcs)
 
+    def get_maximum_AZ(self):
```
HDembinski, line 176: Shouldn't this be private, too?

src/impy/kinematics.py
```diff
@@ -80,22 +173,28 @@ def sort_list(k, idcs):
...
+        a_val = self.component_A
+        max_ind = a_val.index(max(a_val))
+        return self.component_A[max_ind], self.component_Z[max_ind]
+
+    def set_rng_seed(self, seed):
+        self.rng = self.default_rng(seed)
+
     def get_random_AZ(self):
```
HDembinski, line 184: Also this should be private, I think

jncots:
[Fix typo in self.pcm](https://github.com/impy-project/impy/pull/43/commits/536ee90ba0c862d9904e93bcc5c2f6352e16aabc) fixes the issue #58

src/impy/kinematics.py
jncots: Good idea. Added.

src/impy/kinematics.py
jncots: Yes, it is not needed. Removed

src/impy/kinematics.py
jncots: I see. Nothing changed

src/impy/kinematics.py
jncots: Fixed to private

src/impy/kinematics.py
jncots: Fixed to private

jncots:
[8a9dc03](https://github.com/impy-project/impy/pull/43/commits/8a9dc0351c0c1d15535126ae32371d697ccd011d) adds `random_state` to CompositeTarget constructor as suggested by [HDembinski](https://github.com/HDembinski) and change some functions to private.

afedynitch (approved):
Looks good to me! The random state comment is implemented as described by @HDembinski and it's cool to have these options. Did you mean this way of implementation of some kind of different numpy thing?

Otherwise great, good to see the tests. As a comment, there is usually no need to check such long sequences to find divergence. But since it's there, it's ok!

Thanks.

jncots:
Add functions `from_cms` and `from_laboratory` which transform matrix of 4-vectors in hep format `p=(px, py, pz, en)` from the `cms` and `laboratory` of two 4-vectors `p1` and `p2`.

HDembinski:
> Add functions `from_cms` and `from_laboratory` which transform matrix of 4-vectors in hep format `p=(px, py, pz, en)` from the `cms` and `laboratory` of two 4-vectors `p1` and `p2`. `laboratory` is `p2` rest frame.

This is interesting, but isn't this out of scope for impy?

For simplicity, we should only generate events where the beams are aligned with the z-axis. This is because impy should correspond to the lowest common denominator of all models. I believe at least some models cannot generate events with beams that have a relative angle. If the beams are always aligned, then it makes no sense to provide frames in which the beams are not aligned with the z-axis, this just makes the impy API more complicated. If the beams are always aligned with z, then we only need the simple 2D boosts that are already implemented in EventKinematics.

@afedynitch What do you think?

afedynitch:
I believe it is a good generic addition. As you correctly say, most models assume everything to be aligned on the z-axis, except phojet. In phojet, one specifies the beam and target as 4-vectors. Internally, the model rotates the frame to be aligned with z and then does the back-transformation before returning the particle stack.

The addition doesn't loose on generality and has the advantage that now impy can be used in arbitrary MC simulations of astrophysical sources, toy cascades or whatsoever by simply feeding the 4-vectors of projectile and target. IMO, this is not out of scope and actually part of what the models internally do anyways. 

My two concerns are:
1. There are some tests missing for these functions
2. There should be a similar derived class like CenterOfMass or Laboratory, that instead takes two 4-vectors and 2 particle or nuclear ids and does the boost fully transparent to the user (so the interface is not more complicated than now)

jncots:
Tests for the transformations are added. Doubles have relative accuracy around 1e-16.  To pass the tests for sure accuracy is lowered to 1e-15. It is noticed that under transformations relative accuracy lowers to 1e-15*(gamma*E/m)**2, where gamma is Lorentz factor of tranformation and E and m are energy and mass of the particle (see `tolerance_level` in tests).

HDembinski:
> The addition doesn't loose on generality and has the advantage that now impy can be used in arbitrary MC simulations of astrophysical sources, toy cascades or whatsoever by simply feeding the 4-vectors of projectile and target. IMO, this is not out of scope and actually part of what the models internally do anyways.

Adding features that other libraries already provide just increases the entropy of the universe. Instead of adding non-essential functionality, we should refer the user in the docs to a specialized library which provides this instead of implementing it ourselves. Also, adding anything requires not only to maintain the code, but also to write tests and docs.

There are specialized libraries that provide Lorentz vectors with generic boosts in Python. However, even HepMC3 can do generic boosts and we now provide the conversion from impy event to HepMC3 event in my open PR. So that functionality is already there. End users which need generic boosts will already have their own routines for this.

In summary: generic boosts are out of scope for this PR and out of scope for this library. It is easy to add non-essential features to a library, bloating it up. A bloated library takes longer to learn, is harder to maintain, and more difficult for potential contributors. Therefore, we need to strife to write minimal libraries.

afedynitch:
I disagree with the notion that proving 4-vector input among center of mass or laboratory frames is “bloated”. We are doing this service anyways by converting user input to the frame that a generator understands. I also disagree that thinking about the correct boosts and rotations is something that trivial to just leave it to the user. So I disagree on not proving the 4-vector input because I’m gonna use it for sure, and others as well.

What I agree on is that if some functionality is well implemented somewhere and doesn’t add too many dependencies, this functionality should not be duplicated. If you prefer to use a different package (preferably one that’s already part of our dependencies, or pure python numpy) to do the boosts and rotations, fine, I totally wouldn’t mind.

This should remind all of us (@jncots :) ) to discuss design proposals in a bit in advance to collect opinions.

jncots:
> This should remind all of us (@jncots :) ) to discuss design proposals in a bit in advance to collect opinions.

My bad. Although it is late, the proposal is following. Currently impy has 2 classes: `CenterOfMass` which simulates events in center of mass frame and returns momenta of produced particles in the same frame. The same is with `FixedTarget`: it returns momenta of produces particles in the rest frame of target. `EventKinematics` has the possibility to receive arbitrary 4 - momenta and calculate corresponding quantities for center of mass frame and rest frame. The momenta of produced particles can be returned in  center of mass frame and rest frame depending on `impy_config["user_frame"]`.  

What is missing is return of produced momenta in **initial** frame. Depending on "output_frame"=("center-of-mass", "laboratory") of model one could transform event.px, ..., event.en to initial frame using `from_cms(p1, p2, [event.px, event.py, event.pz, event.en])` or `from_laboratory(p1, p2, [event.px, event.py, event.pz, event.en])`. 

I think it's convenient to have this functionality. The implementation of the transformations does not play a role. If it is needed, I rewrite it using other libraries.

jncots:
I removed functions related to frame transformations.

PR #48, opened 2022-08-18
replace event.p_ids with event.id

HDembinski:
Underscores should be avoided in short names that are typed often. Therefore, this replaces event.p_ids with event.id. While `id` is a builtin function in Python, it is ok to use `id` as a class attribute.

Why `id`?
- It is as short as possible, `pid` just adds another letter without much benefit.
- No plural. p_ids stood out before, since it was a plural form, while all other short arrays are singular (px, not pxs; m, not ms)


jncots:
Since `p_ids` are used to keep `pdg`, why not to rename `p_ids` -> `pdg`? With `id` the code `event.xlab[event.id == 2212]` looks like it is `id` of event. While `event.xlab[event.pdg == 2212]` says directly what is the filtering conditions.

afedynitch:
I second the opinion on `event.pdg`.

HDembinski:
> I second the opinion on `event.pdg`.

We will never expose any other id, so `id` is best. It is unique, always will be, and it is the least to type. There is no need to make the name reflect what kind of id it is, this has to be documented.

We also do not document that the momentum is given in units of gev/c, we are happy to use .px instead of .px_in_gev. Why are we not doing that? Because it is too much too type and if you want to know what kind of unit the momentum has, you have to look into the documentation of the class.

HDembinski:
`event.xlab[event.id == 2212]` looks like it is `id` of event.

Good then that there is no event id and the concept of an event id makes no sense. We are currently using the field `nevent` in the Event class, this also needs to go, it is superfluous.

HDembinski:
I am willing to compromise on `event.pid`, but not `event.pdg`, sorry. The latter does not even have `id` in the name.

HDembinski:
Note that the hepevent record which we use, also has the field `idhep`, where `hep` is a common suffix, since Fortran does not have namespaces. It is not `dpgidhep`.

jncots:
`pid` is fine with me.

afedynitch:
`event.pid` is ok!

HDembinski:
Ok, `pid` it is, I will adapt my corresponding PR.

HDembinski:
Closing this, because it is superseded by #52

Issue #21, opened 2022-08-04
Possible re-design of event filters


HDembinski:
From the discussion of #20 
> For the filters, there is maybe no deep reason. I personally think that numpy boolean selections or np.where-s are a sufficient filtering framework. But preselecting charged and stable particles is so common that requiring always an additional expression `event.status == 1` in your selections is quite annoying every time. I think what is needed are just these two "pre-filters" (for final state and charge) and the rest is up to the user to select stuff, e.g. something like `np.where((event.p_ids == 211) & (np.abs(event.eta) < 2.5))` or way more complex expressions.

Ok, +1 for adding a short for common filters, but ideally, we should provide a simple and familiar system for filtering for the user and then make applying the standard filters an easy-to-use special case. It is good to write the API so that filtering is done always in the same way, even for common filtering operations.

Think about it from the user side. If you provide a special method that does some filtering, but the function does not show how to do the filtering in general, then the user has to learn two things: your special method and the general way. It is more educational to only show the general way and make that simple to understand, so there is "one obvious way" to do it (see Zen of Python, which is excellent design advice that should be taken very seriously, despite the light-hearted tone). This way, people automatically learn how to do powerful filtering from the commonly applied simple filtering.

Here is some example how such filtering could look like. We want to mask all arrays at once when we apply a filter, so we need to be able to apply masks directly on the event object, like so
```py
filtered_event = event[mask]
```
This would require implementing `__getitem__` on the event object, which then forwards the mask to its member arrays and constructs a new event object with the filtered contents.

However, this is a bit inelegant and inefficient. It would be more elegant and efficient if the event object itself maps to the FORTRAN common memory block. event should be a subclass of numpy.ndarray or numpy.ma.masked_array. Subclassing ndarray is not easy, but I have done it before and it is well supported by numpy (even if the process is a bit cumbersome).

If event itself is a numpy array view of the underlying FORTRAN memory block, then masking particles becomes a trivial operation that we do not even have to implement. For convenience, event could still have properties like `event.px` that return a subview into all px values etc. So the interface visible to the user can remain the same.

Some words on masked arrays. Numpy's masked arrays could be a great match for this application. Masked arrays allow one to filter out particles without physically changing the value arrays. The masks act non-destructively, so "clearing filters" remains possible. All that happens without copying any data either.

afedynitch:
[...]

PR #51, opened 2022-08-19
Improved Event classes

HDembinski:
Closes #48 #21

This great simplifies the events, by putting common code in the base class and removing the duplicates in the derived classes. The event is conceptually a non-owning view into the Fortran record, which you should not modify in-place. The filtering logic with complicated caching was removed, which attempted to give the illusion of being able to apply filter in place while keeping the original record intact.

Instead, the event now uses the familiar numpy syntax for applying masks `event[mask]`, which in general returns a copy (same as with the previous filtering logic). This is now explicit, instead of implicit, which is better design. Applying masks on the event returns a simpler EventData object, which intentionally does not have all the methods of the original event and which in general is a copy. The original event is now read-only, to ensure that the record is not accidentally modified. 

The `npart` attribute was dropped from the event, instead the more pythonic `len(event)` may be used. 

The changes greatly simplify the implementation.


HDembinski:
@afedynitch The tests are failing because of particletools. six is no longer installed by impy, and apparently particletools is requiring six but not listing it as its install requirements.

PR #52, opened 2022-08-19
New event

HDembinski:
Closes #48 #21

This great simplifies the events, by putting common code in the base class and removing the duplicates in the derived classes. The event is conceptually a non-owning view into the Fortran record, which you should not modify in-place. The filtering logic with complicated caching was removed, which attempted to give the illusion of being able to apply filter in place while keeping the original record intact.

Instead, the event now uses the familiar numpy syntax for applying masks `event[mask]`, which in general returns a copy (same as with the previous filtering logic). This is now explicit, instead of implicit, which is better design. Applying masks on the event returns a simpler EventData object, which intentionally does not have all the methods of the original event and which in general is a copy. The original event is now read-only, to ensure that the record is not accidentally modified. 

The `npart` attribute was dropped from the event, instead the more pythonic `len(event)` may be used. 

The changes greatly simplify the implementation.

***

@afedynitch The tests are failing because of particletools. six is no longer installed by impy, and apparently particletools is requiring six but not listing it as its install requirements.


afedynitch:
@HDembinski I included `six` in the setup.py install requirements for particletools. It has an ancient setuptools + Azure CI despite being a very simple package.

afedynitch:
The redesign looks very nice! Have to checked some performance numbers?

HDembinski:
> The redesign looks very nice! Have to checked some performance numbers?

I haven't checked performance. What would be a good performance test?

HDembinski:
Since this patchs seems to get some acceptance, I will close #48 and include the change from p_ids to pid here.

HDembinski:
Benchmark results for two models, SIBYLL2.1 and Pythia6. Smaller is better. I compare the this branch with master. 1000 events are generated. See `bench/test_new_event.py` and `bench/test_old_event.py` for details.

The most relevant results are those for SIBYLL2.1, because this model is really fast and any changes to the MCEvent Python layer are only relevant if most of the time is not anyway spend inside the model.

Where the difference is not in the noise, this branch is faster than master.
```
------- benchmark 'test_event[Pythia6]': 2 tests ------
Name (time in s)                         Mean
-------------------------------------------------------
test_event[Pythia6] (0001_new_eve)     1.0631 (1.0)
test_event[Pythia6] (0002_old_eve)     1.2083 (1.14)
-------------------------------------------------------

------- benchmark 'test_event[Sibyll21]': 2 tests -------
Name (time in ms)                          Mean
---------------------------------------------------------
test_event[Sibyll21] (0001_new_eve)     80.4139 (1.0)
test_event[Sibyll21] (0002_old_eve)     81.4511 (1.01)
---------------------------------------------------------

------- benchmark 'test_event_final[Pythia6]': 2 tests ------
Name (time in s)                               Mean
-------------------------------------------------------------
test_event_final[Pythia6] (0002_old_eve)     1.1416 (1.0)
test_event_final[Pythia6] (0001_new_eve)     1.2284 (1.08)
-------------------------------------------------------------

------- benchmark 'test_event_final[Sibyll21]': 2 tests --------
Name (time in ms)                                 Mean
----------------------------------------------------------------
test_event_final[Sibyll21] (0001_new_eve)     111.3343 (1.0)
test_event_final[Sibyll21] (0002_old_eve)     123.1920 (1.11)
----------------------------------------------------------------

------- benchmark 'test_event_final_charged[Pythia6]': 2 tests ------
Name (time in s)                                       Mean
---------------------------------------------------------------------
test_event_final_charged[Pythia6] (0001_new_eve)     1.1324 (1.00)
test_event_final_charged[Pythia6] (0002_old_eve)     1.1319 (1.0)
---------------------------------------------------------------------

------- benchmark 'test_event_final_charged[Sibyll21]': 2 tests --------
Name (time in ms)                                         Mean
------------------------------------------------------------------------
test_event_final_charged[Sibyll21] (0001_new_eve)     115.9063 (1.02)
test_event_final_charged[Sibyll21] (0002_old_eve)     113.2693 (1.0)
------------------------------------------------------------------------
```

HDembinski:
The previous benchmark was for pp at 10 TeV. This creates comparably large events and takes longer. To maximize the impact of the MCEvent code, I did another benchmark at 10 GeV. At such low energies, the model code runs faster and we should see more impact from the Python code.

In this benchmark, the new code is sometimes faster, sometimes slower by <10 %, which indicates that this is within the noise.
```
-------- benchmark 'test_event[Pythia6]': 2 tests -------
Name (time in ms)                          Mean
---------------------------------------------------------
test_event[Pythia6] (0005_new_eve)     843.3506 (1.01)
test_event[Pythia6] (0004_old_eve)     835.5908 (1.0)
---------------------------------------------------------

------- benchmark 'test_event[Sibyll21]': 2 tests -------
Name (time in ms)                          Mean
---------------------------------------------------------
test_event[Sibyll21] (0004_old_eve)     19.3165 (1.0)
test_event[Sibyll21] (0005_new_eve)     20.4654 (1.06)
---------------------------------------------------------

-------- benchmark 'test_event_final[Pythia6]': 2 tests -------
Name (time in ms)                                Mean
---------------------------------------------------------------
test_event_final[Pythia6] (0004_old_eve)     835.1406 (1.0)
test_event_final[Pythia6] (0005_new_eve)     931.1001 (1.11)
---------------------------------------------------------------

------- benchmark 'test_event_final[Sibyll21]': 2 tests -------
Name (time in ms)                                Mean
---------------------------------------------------------------
test_event_final[Sibyll21] (0005_new_eve)     38.1884 (1.0)
test_event_final[Sibyll21] (0004_old_eve)     40.4445 (1.06)
---------------------------------------------------------------

-------- benchmark 'test_event_final_charged[Pythia6]': 2 tests -------
Name (time in ms)                                        Mean
-----------------------------------------------------------------------
test_event_final_charged[Pythia6] (0004_old_eve)     834.2080 (1.0)
test_event_final_charged[Pythia6] (0005_new_eve)     928.1560 (1.11)
-----------------------------------------------------------------------

------- benchmark 'test_event_final_charged[Sibyll21]': 2 tests -------
Name (time in ms)                                        Mean
-----------------------------------------------------------------------
test_event_final_charged[Sibyll21] (0005_new_eve)     36.8182 (1.0)
test_event_final_charged[Sibyll21] (0004_old_eve)     42.3265 (1.15)
-----------------------------------------------------------------------
```

HDembinski:
I wrote the new code so that it is as fast as the old code or faster, because we avoid the caching. The new code is potentially slower in one aspect: event.charge is not calculated on demand but always. However, this should have negligible impact unless the charge calculation is very slow in the model, which is generally not the case.

Only Pythia6 has a comparably slow calculation of the charge, but since Pythia6 is so slow in general, the difference is negligible. For most analyses, we need charge anyway, so that its delayed calculation is not that impactful.

HDembinski:
If anyone wonders why the tests suddenly pass: I just marked the tests that are expected to fail with `pytest.xfail`. Checking unit tests in CI that are permanently broken is not helpful, because we want to see when something breaks, but we cannot detect this, when the tests are broken all the time.

HDembinski:
Not strictly related to this PR, but I cleaned up our test folder. All tests should be run in CI. Tests that do not work and are not expected to work, should not be called "test_X.py", so that they are ignored by pytest. Accordinly, I renamed all currently broken tests from `test_X.py` to `fixme_X.py`.

HDembinski:
The tests are failing again, now because of the multiple initialization issue. We would need to run each model in each test with a multiprocessing Pool to work around this, that's very cumbersome.

HDembinski:
Wohoo, first PR that actually has passing CI tests.

HDembinski:
@afedynitch Is this good to go? I think it would be useful to merge this before #43 (which is orthogonal), because then we can actually use the now working CI to check #43.

afedynitch (approved):
Super cool!! Thanks so much. The design is great.

And I learned two extremely nice things: `dataclass` and how you creatively used `argparse.Namespace` 👍 

Also thanks for the performance numbers, I didn't know there is an integrated benchmark tool in pytest.

afedynitch:
Good to go. I really like the design. It simplifies the maintenance by a lot.

jncots:
I like it. Looks cleaner now.

jncots (approved):

HDembinski:
> Super cool!! Thanks so much. The design is great.

Thanks!
 
> And I learned two extremely nice things: `dataclass` and how you creatively used `argparse.Namespace` 👍

Yes, dataclass is the modern Python equivalent of a namedtuple. Since we are dropping support for Python < 3.7, we can use nice new things. Haha, you found my hack with the argparse.Namespace. I think that class is really useful and I have seen it reimplemented many times, because people don't know where to find it in the standard library. Perhaps even the standard library has several implementations of it.
 
> Also thanks for the performance numbers, I didn't know there is an integrated benchmark tool in pytest.

Since pytest is so popular, people wrote all kinds of plugins, like pytest-benchmark or pytest-cov.

PR #54, opened 2022-08-22
Fix hepmc writer

### LINKED ISSUE #57: SIBYLL-2.1 event has PID array which contains zeros
Here is a random event that I just generated. You can see that quite a few particles in the list have pid == 0, which is invalid.
```
SibyllEvent(kin=Event kinematics:
	ecm      :  100.00000
	pcm      :   49.98239
	elab     : 5328.01821
	plab     : 5328.01813
	gamma_cm :   53.28956
	bgamm_cm :   53.28018
	pdgid 1  :       2212
	nucprop 1: 1/1.0
	pdgid 2  :       2212
	nucprop 2: 1/1.0
, frame='center-of-mass', nevent=0, pid=array([   0,    0,  211, 2212,    0,  321,    0,    0,  211, 2112, -211,
          0,  211, -211, 3122,  211, -211,    0,  211, -211,    0,   22,
         22,   22,   22,   22,   22], dtype=int32), status=array([2, 2, 1, 1, 2, 1, 2, 2, 1, 1, 1, 2, 1, 1, 1, 1, 1, 2, 1, 1, 2, 1,
       1, 1, 1, 1, 1], dtype=int32), charge=array([ 0,  0,  1,  1,  0,  1,  0,  0,  1,  0, -1,  0,  1, -1,  0,  1, -1,
        0,  1, -1,  0,  0,  0,  0,  0,  0,  0], dtype=int32), px=array([-0.63597113,  0.30926612, -0.57559907,  0.46384686,  0.51563632,
        0.14977658, -0.34058014,  0.03916309, -0.233496  ,  0.30795735,
       -0.05713884, -0.57883763,  0.48268995, -0.17342119,  0.47702566,
        0.03861185, -0.45473272,  0.11414974,  0.23528023, -0.06781755,
       -0.12830749, -0.30247974, -0.27651668,  0.13449517, -0.02031411,
       -0.13858746,  0.01024476]), py=array([-0.06726643,  0.21742968, -0.43126103,  0.17648369, -0.26430294,
        0.72727096, -0.14105777, -0.0387254 , -0.28010216,  0.10153134,
        0.15356468, -0.22083166,  0.23095961, -0.01352806, -0.03491727,
       -0.22938628,  0.02613074, -0.1671897 , -0.08275045,  0.02067954,
        0.02334706, -0.18664402, -0.03424823, -0.0877149 , -0.07952067,
        0.05718039, -0.03382691]), pz=array([ 1.59013569e+00,  1.86574388e+00,  8.69777298e+00,  2.52620697e+01,
        5.80550623e+00, -2.51288414e-02,  2.31336451e+00,  3.51808786e+00,
       -3.89408565e+00, -4.51334534e+01,  6.23095334e-02,  1.52783954e+00,
        1.65816951e+00,  2.07590282e-01,  4.39044857e+00,  1.41507137e+00,
        2.05674314e+00,  2.56640792e-01,  2.39994287e+00,  2.20560968e-01,
        8.97658288e-01,  8.76120746e-01,  6.52137876e-01,  1.56916633e-01,
        9.97945592e-02,  7.78088510e-01,  1.19816080e-01]), en=array([ 1.87824392,  2.05776715,  8.72942257, 25.28435707,  5.99597931,
        0.91774952,  2.47411704,  3.60436106,  3.91361141, 45.14439392,
        0.22407411,  1.65418553,  1.75117266,  0.30661198,  4.55515432,
        1.44083834,  2.11730909,  0.35682869,  2.41690087,  0.27046952,
        0.91706872,  0.94547206,  0.7091673 ,  0.22630349,  0.13062309,
        0.79239994,  0.12492038]), m=array([0.7683    , 0.76859999, 0.13957   , 0.93826997, 1.38279998,
       0.49364999, 0.7683    , 0.78195   , 0.13957   , 0.93957001,
       0.13957   , 0.13496999, 0.13957   , 0.13957   , 1.11563003,
       0.13957   , 0.13957   , 0.13496999, 0.13957   , 0.13957   ,
       0.13496999, 0.        , 0.        , 0.        , 0.        ,
       0.        , 0.        ]), vx=array([0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.,
       0., 0., 0., 0., 0., 0., 0., 0., 0., 0.]), vy=array([0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.,
       0., 0., 0., 0., 0., 0., 0., 0., 0., 0.]), vz=array([0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.,
       0., 0., 0., 0., 0., 0., 0., 0., 0., 0.]), vt=array([0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.,
       0., 0., 0., 0., 0., 0., 0., 0., 0., 0.]), parents=None, children=None)
```

HDembinski:
Closes #57 and #56 

This patches repairs IO with pyhepmc, mostly by applying external fixes to pyhepmc. HepMCWriter is removed. Event IO can be done directly with pyhepmc, thanks to automatic conversion of foreign event objects. pyhepmc checks whether the event has a method `to_hepmc3` and calls that automatically. 

Other changes:
- Fixed the source of the SIBYLL-2.1 compilation warning
- impy event got a special function `_repr_html_` which draws the event in a Jupyter notebook if all the required optional libraries are available.
- Fix for parents in EPOSEvent
- Fix min-bias generation of inelastic events with Pythia6 with diffraction enabled
- Add limited particle history for SIBYLL
- New example to demonstrate HepMC3 conversion, IO, visualization


HDembinski:
@afedynitch @jncots have a look at the cool new notebook that demos the visualization and IO with pyhepmc. I think we can remove the internal writer, because IO with pyhepmc is trivial and it is bad to just add thin wrappers over other libraries.

jncots:
Indeed, the visualization is cool and IO API is simple. For the current implementation of `HepMCWriter` there is no reason to have it, instead of direct use of `pyhepmc`. `HepMCWriter` could be useful if in the future there appear some other IO functions which should be packed in one class. One thing I noticed when I tested `crmc` is that constant output greatly slows down program execution. Is there way how to output a bunch of events in a single flush to the file instead of writing them one by one?

HDembinski:
> Indeed, the visualization is cool and IO API is simple. For the current implementation of `HepMCWriter` there is no reason to have it, instead of direct use of `pyhepmc`. `HepMCWriter` could be useful if in the future there appear some other IO functions which should be packed in one class.

For other IO, I suggest to use Python's builtin pickle. Reader and Writer classes are not Pythonic. There are two orthogonal concepts in Python: create a file handle that represents some kind of file (e.g. open, gzip.open, ...) and the converter which serializes an object in memory (e.g. pickle, json, yaml, ...). The ideal design is compatible with this structure. This is also not perfectly implemented in pyhepmc either at the moment.

> One thing I noticed when I tested `crmc` is that constant output greatly slows down program execution. Is there way how to output a bunch of events in a single flush to the file instead of writing them one by one?

That is an optimization that HepMC3 should do, not something we should try to implement.
@afedynitch Isn't impy + pyhepmc already faster than crmc? I think you claimed that in a recent presentation. In that case, we don't have to worry?

By default, text-based IO using a buffer to do exactly what that, it writes the file content first into a memory buffer and only when that buffer is full, the content is written to disk in one chunk. This is more efficient and it is the default when writing binary files. For writing ASCII, the default is line buffering, which is inefficient for HepMC output. Perhaps that's the error in crmc, they forgot to switch to block buffering. I don't know what HepMC3 uses.

jncots:
> Isn't impy + pyhepmc already faster than crmc?

`crmc` is significantly slower than `impy` because of how the output to the file is organized. The speed of both are comparable without file output.

HDembinski:
> > Isn't impy + pyhepmc already faster than crmc?
> 
> `crmc` is significantly slower than `impy` because of how the output to the file is organized. The speed of both are comparable without file output.

Good to hear that impy and crmc are on par without IO and that impy + pyhepmc is faster.

So that means we do not have to optimize IO and the speed increase is mainly coming from using the HepMC3 library instead of HepMC2, is this correct?

HDembinski:
Thinking more about it, by implementing `pyhepmc.dump` and `pyhepmc.load` instead of `pyhepmc.open`, I might be able to address the buffering anyway. To implement load and dump with the existing C++ interface, I would have to write into a stringstream, so that's the intermediate buffer.

afedynitch:
I really love the graphs! I think it is actually useful for model development. After merging #52 this PR will look much simpler. So far I'm totally fine with it to go.

@HDembinski, btw. Do you remember this graph? I think someone from SHERPA produced it years ago. Do you have an idea how they might have done this?

image

HDembinski:
> @HDembinski hope you don't mind if I have "polluted" this PR with some marginally related issues.

I do not mind at all, these are cool additions! Anyway, the main author is allowed to pollute up PRs as they see fit.

HDembinski:
@afedynitch The result with SIBYLL looks a bit better now, but SIBYLL does not place the beam particles on the stack, which is why the result looks still weird:

image

Could you do fix that in the Fortran layer? You can add the two beam particles at the end of the particle stack and then make them the mothers of all other particles without mothers.

We could also fix it in the Python layer, by overriding the `to_hepmc3` method. The fix should be reasonably fast even in Python.

HDembinski:
image

The Pythia6 event also looks different now. Since your change to the settings, there are two additional "initial" gluons now. I saw this before, Pythia6 does not connect all particles to the beam particles for some reason. I think that's ok if this is really how Pythia6 operates. I know from Tanguy that it is sometimes difficult to know where to connect all particles.

HDembinski:
> @HDembinski, btw. Do you remember this graph? I think someone from SHERPA produced it years ago. Do you have an idea how they might have done this?

I know this plot, but I don't know how it was made, assuming it was made automatically and not drawn by hand.

To draw plots like this automatically you need some algorithms that spread the nodes of a graph out like this. Graphviz also has such engines, e.g. "twopi". I can add more colors, no problem, but I cannot draw squiggly lines for gluons.

Please make suggestions how to improve the pyhepmc visualization, e.g. which colors/style to use for what kind of particles.

I think it makes sense to use different colors for mesons, baryons, and leptons.

afedynitch:
> there are two additional "initial" gluons now.

Could/Should be initial state radiation. At parton level a graph it's supposed to come from nowhere :)

HDembinski:
> > there are two additional "initial" gluons now.
> 
> Could/Should be initial state radiation. At parton level a graph it's supposed to come from nowhere :)

That's ok with me! If this is expected, I think we should not try to make the event prettier, but rather give an honest visualization of what's in the record.

I think I found a bug in your recent SIBYLL change. The charges of the new rhos are all zero, I think you need to copy their charges.

HDembinski:
Side remark 1: The current visualization in pyhepmc uses graphviz, an old and mature library. However, graphviz is not native to Python and the Python API is not super nice. I just discovered networkx, which could be a good replacement.

HDembinski:
Hm, test_charge.py now fails also for Pythia6. I changed the energy of the test event, so it is related to that. I will investigate.

HDembinski:
@afedynitch It looks like your change to the Pythia6 configuration also enabled elastic scattering. Can you perhaps enable only inelastic event generation to be consistent with the other generators?

HDembinski:
I think the failing charge test for Sibyll-2.1 can be traced to sibyll_21.f:3606 where ICHP is defined. This table probably lacks the correct charges for non-final particles.

@afedynitch Should we fix this in sibyll_21.f or in our own Fortran layer?

afedynitch:
> I think the failing charge test for Sibyll-2.1 can be traced to sibyll_21.f:3606 where ICHP is defined. This table probably lacks the correct charges for non-final particles.
> 
> @afedynitch Should we fix this in sibyll_21.f or in our own Fortran layer?

@HDembinski I'm opening a branch+PR now to fix DPMJET PHOJET and will also fix this SIBYLL issue in the fortran layer.

afedynitch:
Elastic scattering in Pythia is disabled.

HDembinski:
MCEvent is now "pickleable": you save a MCEvent, but you load an EventData event. This is convenient for users, since it makes manual conversion of MCEvent to EventData superfluous in many cases.

HDembinski:
Not really related to this PR, but we are fixing other things anyway: I fixed the SIBYLL-2.1 compile warning:
```
[  1%] Linking Fortran shared module /Users/runner/work/impy/impy/build/lib.macosx-10.15-x86_64-cpython-37/impy/models/_sib21.cpython-37m-darwin.so
  ld: warning: tentative definition of '_s_chp_' with size 1188 from 'CMakeFiles/_sib21.dir/src/fortran/sibyll/sib21aux.f.o' is being replaced by real definition of smaller size 592 from 'CMakeFiles/_sib21.dir/src/fortran/sibyll/sibyll_21.f.o'
  ld: warning: tentative definition of '_s_chp_' with size 1188 from 'CMakeFiles/_sib21.dir/src/fortran/sibyll/sibyll_init.f.o' is being replaced by real definition of smaller size 592 from 'CMakeFiles/_sib21.dir/src/fortran/sibyll/sibyll_21.f.o'
  [  1%] Built target _sib21
```

HDembinski:
The impy event can now be displayed in a Jupyter notebook without converting it to HepMC3 format first. If all optional requirements are met, the conversion is now done automatically when Jupyter tries to display the event.

HDembinski:
@afedynitch @jncots Can I get your approval to merge this so that I can start working on the next item?

afedynitch (approved):
Excellent work!

HDembinski:
> Excellent work!

Thanks! Btw: @afedynitch @jncots pyhepmc now has beautiful documentation and all the classes and methods have docstrings. This should make working with pyhepmc much easier for you. https://scikit-hep.org/pyhepmc/
 
> @HDembinski one comment maybe on the `SIBYLL_TABLE_SIZE` Is there a reason why you introduced this variable? I think all the 2.3 SIBYLLs have 99 and SIBYLL21 has 49. Because of single precision it's auxiliary functions can not be generalized from those of 2.3.

Right, the SIBYLL_TABLE_SIZE is to adjust the interface of the f2py module to the actual size used in the code. Apparently it works also when you use the larger size 99 for Sibyll-2.1, but we get a warning (which I listed above). My experience with C++ is that warnings should not be ignored, even though the code still works, a warning usually means that it may stop working in the future or it may stop working when optimisations are turned on.

jncots (approved):

PR #61, opened 2022-09-01
Add rng_state

jncots:
As it was discussed there is a need for control of random number generator for testing and debug purposes. The required for this basic functions are: view and copy a current state of generator and restore the state from the saved copy. 

The working example/prototype (currently only for Sophia) is following

```python
import impy
from impy.kinematics import FixedTarget
from impy.constants import TeV, GeV
import numpy as np

ekin = FixedTarget(13 * TeV, "photon", "proton")
generator = impy.models.Sophia20(ekin)

# One can get a view (currently it is copy and it will converted to view):
state0 = generator.rng_state
print("Before run = ", state0.counter, state0.seed)
# Or the same
print("Before run = ", generator.rng_state.counter, generator.rng_state.seed)

# Save a state to a variable:
state0 = state0.copy()

counters = []

for event in generator(10):
    # Print the value of counter (how many time the 
    # random number generator has been called so far) after each event
    print("rng calls = ", generator.rng_state.counter)
    counters.append(generator.rng_state.counter)


# One can save the state to file
generator.dump_rng_state_to("rng_state.dat")
# Or to variable
state1 = generator.rng_state.copy()

# Restore the state from variable
generator.rng_state = state0

# And compare to check whether the state is restored: 
i = 0
for event in generator(10):
    assert counters[i] == generator.rng_state.counter, "Not equal"
    i  = i + 1

state2 = generator.rng_state.copy()
# Restore from file
generator.restore_rng_state_from("rng_state.dat")

# And check for equality
if state2 == generator.rng_state:
    print("Equal")
```

I am not really happy with descriptive but still long names for `restore_rng_state_from` and `dump_rng_state_to`. It would be good to find something more concise.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -336,11 +336,11 @@ f2py_add_module(_pythia6
 f2py_add_module(_sophia
   FUNCTIONS
   eventgen print_event crossection initial
-  icon_pdg_sib init_rmmard toevt ${impy_logging}
+  icon_pdg_sib init_rmmard toevt crranma4 ${impy_logging}
   SOURCES
-  ${f2py_dir}/_sophia.pyf
-  ${f2py_dir}/_sophiamodule.c
-  ${f2py_dir}/_sophia-f2pywrappers.f
+  # ${f2py_dir}/_sophia.pyf
+  # ${f2py_dir}/_sophiamodule.c
+  # ${f2py_dir}/_sophia-f2pywrappers.f
   ${fortran_dir}/sophia/SOPHIA20.f
   ${fortran_dir}/sophia/eventgen.f
   ${fortran_dir}/sophia/sampling.f
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,93 +78,20 @@ else() # intel
 endif()
 
 ### common for all models
+set(impy_definitions IMPY)
 set(fortran_dir ${CMAKE_SOURCE_DIR}/src/fortran)
 set(f2py_dir ${CMAKE_SOURCE_DIR}/src/f2py)
 
 set(impy_logging impy_openlogfile impy_closelogfile)
 set(logging_source ${fortran_dir}/logging.f)
-set(rangen_source rangen.f)
-set(rangen_fpp_source ${fortran_dir}/rangen.fpp)
-set(rangen_sib21_source rangen_sib21.f)
-set(sibyll_init_source sibyll_init.f)
-set(sibyll_init_sib21_source sibyll_init_sib21.f)
-
-#
-set(sophia_jetset74dp_source jetset74dp.f)
-set(sophia_eventgen_source eventgen.f)
-set(pythia6_pythia6428_source pythia-6.4.28.f)
-set(epos_random_source epos-random.f)
-
-# WARNING this is not platform-independent and needs to be adjusted
-# when new fortran compilers should be supported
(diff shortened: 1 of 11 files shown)


afedynitch:
@HDembinski in case you're wondering what this is about. We will update the description, purpose and proposed implementation of the PR after a discussion we had here...

HDembinski:
Instead of custom serialization routines, please support Pythons serialization infrstructure by implementing `__getstate__` and `__setstate__`. You get many benefits for free, like saving in different formats or transparent compression.

jncots:
The above is a commit with working access to random number generator state for Sophia, Pythia6 and EPOS-LHC. `CMakeLists.txt` will be cleaned in the next commits. Here I just fixed the working state. 

`CMakeLists.txt` should have a function for preprocessing original files, because the pattern 

```cmake
set(pythia6_pythia6428_source pythia-6.4.28.f)
add_custom_command(
  OUTPUT ${pythia6_pythia6428_source}
  COMMAND ${CMAKE_Fortran_COMPILER}
    -E -cpp ${fortran_dir}/pythia6/pythia-6.4.28.f
    -o ${pythia6_pythia6428_source}
    -DIMPY
)
```

is met to often. It is needed to remove/substitute functions in original source files. 

Also, it would be convenient to have  function that copy newly generated *.pyf files to `f2py_dir`

afedynitch:
> The above is a commit with working access to random number generator state for Sophia, Pythia6 and EPOS-LHC. `CMakeLists.txt` will be cleaned in the next commits. Here I just fixed the working state.
> 
> `CMakeLists.txt` should have a function for preprocessing original files, because the pattern
> 
> ```cmake
> set(pythia6_pythia6428_source pythia-6.4.28.f)
> add_custom_command(
>   OUTPUT ${pythia6_pythia6428_source}
>   COMMAND ${CMAKE_Fortran_COMPILER}
>     -E -cpp ${fortran_dir}/pythia6/pythia-6.4.28.f
>     -o ${pythia6_pythia6428_source}
>     -DIMPY
> )
> ```
> 
> is met to often. It is needed to remove/substitute functions in original source files.
> 
> Also, it would be convenient to have function that copy newly generated *.pyf files to `f2py_dir`

If you just want to pass a preprocessor flag to the compiler, you can do:
```
target_compile_definitions(target__name PRIVATE IMPY)
target_compile_options(target__name PRIVATE -cpp)
```

HDembinski:
> is met to often. It is needed to remove/substitute functions in original source files.

I can write a function to encapsulate this or you follow Anatoli's advice.
```
target_compile_options(target__name PRIVATE -cpp)
```
This will work for compiling into a library with `add_library`, but it won't work when you need to run numpy.f2py. The function `f2py_add_library` does not run the preprocessor on the sources automatically, because it does not know that numpy.f2py depends on preprocessed sources. So even if you add `-cpp` to all files, numpy.f2py will see the unmodified sources and probably choke on them.

HDembinski:
> Also, it would be convenient to have function that copy newly generated *.pyf files to `f2py_dir`

Do you mean in cmake or as a shell script? The shell script would be a one-liner. We do not want to copy those files automatically in cmake, I think. As a principle, cmake should not modify the source files, only files in its build area that is seprate from the source.

jncots:
I am closing this draft PR because it was merged to PR #59 and later development has been done there.

PR #74, opened 2022-10-06
Fix to_hepmc3() and HepMC3 IO

HDembinski:
This patch tries to address the failures in `test_hepmc_io.py`. It expands `test_to_hepmc3.py`, because at least some failures originate from invalid parent records in the HEPEVT, and those are easier to expose in `test_to_hepmc3.py`.

### DIFF
--- a/tests/test_hepmc_io.py
+++ b/tests/test_hepmc_io.py
@@ -41,7 +41,7 @@ def test_hepmc_io(Model):
 
     with pyhepmc.open(test_file, "w") as f:
         for event in expected:
-            f.write(event)  # this calls to_hepmc3() implicitly
+            f.write(event)
 
     restored = []
     with pyhepmc.open(test_file) as f:
--- a/tests/test_hepmc_io.py
+++ b/tests/test_hepmc_io.py
@@ -15,7 +15,7 @@ models = get_all_models(im)
 
 
 def run(Model):
-    ekin = CenterOfMass(10 * GeV, 2212, 2212)
+    ekin = CenterOfMass(10 * GeV, "proton", "proton")
     if Model == im.Sophia20:
         ekin = FixedTarget(10 * GeV, "photon", "proton")
     gen = Model(ekin, seed=1)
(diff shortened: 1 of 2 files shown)


HDembinski:
The problem with UrQMD34 is that the parent range exceeds the number of particles in the record. This triggers the IndexError inside pyhepmc. I added another assert to `test_to_hepmc3` to expose this.

HDembinski:
I am about to release a new version of pyhepmc 2.7.2 which handles invalid parent records in impy more gracefully than causing the Python interpreter to abort or to segfault. This fixes nothing here, but from now on, you should start to see proper Python exceptions when the parent record is not proper.

afedynitch:
@HDembinski I started looking into DPMJET and phojet. For phojet things look fine if doing 2 fixes to the test. The one you found (probably) is the same I found (-1's added):

`assert not (
      pa[0] <= pa2[0] < pa[1] - 1 or pa[0] <= pa2[1] - 1 < pa[1] - 1
  ), f"vertices {j} and {i} overlap: {pa2}, {pa}"`

The other one is

```
for i, (a, b) in enumerate(unique_vertices):
      assert a >= 0 or a == -1
      if a == event.pid[0] or b==event.pid[0]:
          continue
```

For DPMJET stuff is more complicated. In principle, it generates proton-proton with phojet and so the event stack should be the same. But it is not and I still cannot find the reason for that. So it takes more time to dig deeper in the source code. TBH, the event topology of these events doesn't make sense to me at the moment.

For UrQMD ...

HDembinski:
@afedynitch 

You want to replace 

```
assert not (
                pa[0] <= pa2[0] < pa[1] or pa[0] <= pa2[1] - 1 < pa[1]
            ), f"vertices {j} and {i} overlap: {pa2}, {pa}"
```

```
assert not ( pa[0] <= pa2[0] < pa[1] - 1 or pa[0] <= pa2[1] - 1 < pa[1] - 1 ), f"vertices {j} and {i} overlap: {pa2}, {pa}"
```

I think my version is correct. I formatted the pa ranges beforehand so that they obey C++ conventions of begin, end. end is exclusive, and indices start with 0, see line 39 in the test.

In my version, the first term before "or" checks whether the first index of the pa2 range is inside the pa range and the second term after "or" checks whether the last index of the pa2 range is inside the pa range. If my version is correct, yours cannot be correct.

HDembinski:
I don't know what this does:
```py
for i, (a, b) in enumerate(unique_vertices):
      assert a >= 0 or a == -1
      if a == event.pid[0] or b==event.pid[0]:
          continue
```
Comparing index ranges with pids is invalid, no?

HDembinski:
@jncots @afedynitch This test extension here is useful to have in #77 . Is it ok if I just merge it?

afedynitch:
> Comparing index ranges with pids is invalid, no?

Not really. The event stack contains much more than particles if you take everything and not just the entries with status 2 and 1. The meaning of parents also may change and this is where the problems come from.

afedynitch:
> @jncots @afedynitch This test extension here is useful to have in #77 . Is it ok if I just merge it?

Fine from my side. 

The DPMJET part is still mysterious. I was out for kitesurfing for a couple of days and will be more engaged into this tomorrow.

HDembinski:
Ok, I am going to merge, we can continue to discuss about these issues in #77, because Pythia8 also causes problems.

I identified one issue that can be solved in HepMC3, although it is not really a bug, more about that later.

HDembinski:
> Not really. The event stack contains much more than particles if you take everything and not just the entries with status 2 and 1. The meaning of parents also may change and this is where the problems come from.

@afedynitch I don't follow. Do you mean there is a problem with the test or a problem with the parent array? I thought you talk about the test. I think the test is correct and comparing parent indices with pids is invalid.

afedynitch:
> @afedynitch I don't follow. Do you mean there is a problem with the test or a problem with the parent array? I thought you talk about the test. I think the test is correct and comparing parent indices with pids is invalid.

I mean that treating the event stack as a particle array is a wrong assumption. The event stack contains info entries that simply say what the generator has done when simulating the event. This is up to the authors what to put in the info blocks. Mostly the syntax will be compatible with particles, like it is done for strings. But in this specific case that I worked around the entry was

`Hard Scattering   ….  2212 2212 …. ` that just says that there was a hard scattering event between the projectile and target. It is also impossible to debug without using the native print-out machinery.

PR #75, opened 2022-10-07
Migrate from LFS

jncots:
Currently the data tables in directory iamdata which are used by some models are located on LFS. This doesn't allow to test all models on CI. That should be changed by attaching binary files to release on github. The tables should be downloaded automatically by python when they are required.

This PR is opened to solve this problem.


jncots:
The code is taken from MCEq with minor changes.

HDembinski:
The checksum computation takes some time. I think this is wasteful. We can expect that files on our hard disk are not corrupted.

HDembinski:
- It is best to keep the requirements minimal. We do not need request, we can do this with the Python standardlib.
- I personally also don't like to depend on tqdm, but this may be the right call if we start to show progressbars in many places. Some models like Pythia-8 require a lot of time for initialization and we may want to replace their banners with a short progress bar.
- Checksums should be checked only once when the file is downloaded, not every time the file is loaded into impy. Even then I doubt that checking the checksum is necessary. I never ever downloaded a file from the internet that turned out to be corrupted.
- An elegant API for these cached paths builds on pathlib.Path.
```py
# in impy.utils
from pathlib import Path

class CachedPath(Path):
     def __init__(self, local_path, remote_path):
           super().__init__(local_path)
           if not self.exists():
                 self._download_remote_path(remote_path)
```
We can then use CachedPath instead of Path everywhere where such a thing is needed.

jncots:
The above commit contains lazy downloading and extraction from zip. For a large zip file we can do this:

```python
# Set file containing cache (list of successfully downloaded files) as class property
CachedPath.cache_file = impy_dir / "cache_file.dat"
# Check for existance of  impy_dir / "zipped_files.zip" both in cache_file and on a dist
# If any check fails, the file is downloaded from github_url
zfile = CachedPath(impy_dir / "zipped_files.zip", github_url)
```

If all files are zipped in one file we can unzip any file that is contained in zip file as:

```python
# Set a variable of class with zip file path `zfile` and path to extract files `base_dir`
check_path = PathFromZip(zfile, base_dir)
# Returns path Path(`base_dir/path/in/zip/file.dat`). Extracts file or does nothing if the file already exists
# If path/in/zip/file.dat is not found, the exception is raised
some_file_on_disk = check_path("path/in/zip/file.dat")
```

jncots:
In the above commit a lazy downloading/zip extracting is added to corresponding models via:

```python
pfile = _check_impy_data_path('some_path_to_data')
```

If the path is not found, it extracts corresponding directory/path from zip file, and if there is no zip file, it downloads from given url. It is assumed that zip file root directory is `iamdata`.

There is a catch: most models only ask for directory. In this case `_check_impy_data_path` will check only the existence of directory.

HDembinski:
> There is a catch: most models only ask for directory. In this case _check_impy_data_path will check only the existence of directory.

You could do the usual trick to detect failed downloads by downloading to a different path, which is renamed to the correct path once the download is complete and was not aborted. Example:
- impy requests download of the directory "Pythia8"
- Code starts downloading ".Pythia8"
- Code finishes downloading, renames ".Pythia8" to "Pythia8"

HDembinski:
Merging the Pythia8 branch into this branch before finishing and squashing the Pythia8 branch has the downside that I now cannot find anymore the changes that are specific to this PR. The Pythia8 branch changes more than 400 files.

Can we merge the Pythia8 branch now, even if there are some issues left, and then fix those issues here?

jncots:
> Can we merge the Pythia8 branch now, even if there are some issues left, and then fix those issues here? You then have to do a painful rebase.

Yes, of course.

afedynitch:
Was mit meant to be closed. I also agree with merging pythia8 and clean it up here

HDembinski:
I merged the Pythia8 branch now that the LFS issue is fixed.

jncots:
> You then have to do a painful rebase.

@HDembinski, I haven't understood about rebase. Currently I just merged main into this branch. Is it OK or should I do something more intricate?

HDembinski:
No, if that worked great. I didn't think it would work if the Pythia8 branch is squashed, but I am not a git expert. I hope I didn't forget to squash the Pythia8 branch.

src/impy/iamdata/iamdata_content.json
```diff
@@ -0,0 +1,237 @@
...
+            "Pythia8/xmldoc/RIVETusage.xml",
+            "Pythia8/xmldoc/RopeHadronization.xml",
+            "Pythia8/xmldoc/BeamParameters.xml",
+            "Pythia8/xmldoc/ProcessSelection.xml",
+            "Pythia8/xmldoc/Parallelism.xml",
+            "Pythia8/xmldoc/EventInformation.xml",
+            "Pythia8/xmldoc/Tunes.xml",
+            "Pythia8/xmldoc/TimelikeShowers.xml",
+            "Pythia8/xmldoc/VinciaEW.xml",
+            "Pythia8/xmldoc/MatchingAndMerging.xml",
```
HDembinski, line 17: Is there an advantage in having this as a json file? It is simpler to just make this info a dict in a Python file.

src/impy/models/dpmjetIII.py
```diff
@@ -81,6 +82,10 @@ def __init__(self, event_kinematics, seed=None, logfname=None):
         dpm_conf = impy_config["dpmjetIII"]
 
         info(1, "Initializing DPMJET-III")
+
+        _check_model_data_files(
+            pathlib.Path(dpm_conf["dat_dir"][self.version]).parts[-1]
+        )
         # Set the dpmjpar.dat file
```
HDembinski, line 89: This is what I wanted to avoid by encapsulating all the file retrival in CachedPath that derives from Path. For most or all models, we need to construct a Path objects anyway to initialize the model. Instead of a Path, one makes a CachedPath and the rest works automatically.

src/impy/models/epos.py
```diff
@@ -78,6 +78,8 @@ def __init__(self, event_kinematics, seed=None, logfname=None):
             self._output_frame = "laboratory"
 
         k = event_kinematics
+
+        _check_model_data_files("epos")
         datdir = path.join(base_path, epos_conf["datdir"])
```
HDembinski, line 83: `datdir = CachedPath("epos")`

HDembinski: CachedPath combines two steps that always have to go together anyway for the code to be correct, that's why it is a good design.

src/impy/iamdata/iamdata_content.json
HDembinski: Here is another thing that I don't like about this. We are again putting all information for all models in one file. This is not a good data organisation. When I deal with Pythia8, I should find all the information related to making Pythia8 work in the pythia8.py module. This includes the list of data paths that Pythia8 expects. It is not necessary to list all the files inside xmldoc and pdfdata. Important is that xmldoc and pdfdata are extracted from the zip file. We can expect that all files inside those folders are there.

HDembinski: So, please do not put this data into a json file. Split this data up and put it as literals into the respective model files as Python dicts. Or better, define a dataclass for this in util.py and use that instead of a dict.

HDembinski: We don't need the checksum either, as I have explained below in my comment. When we do operations that need to be "atomic", we perform the operation that takes a while on a fake path name first and then rename the fake path to the final path in the last step, which is super fast. That's how it is done, no checksums required.

jncots: > Is there an advantage in having this as a json file?

The idea was to track all data files belonging to particular model. If, for some reasons, a file is accidentally deleted or corrupted, it will be automatically restored by downloading and unpacking zip file.

Another point, data files could be changed or additional files could be added  in the future. Then another version of zip files should be prepared and uploaded. To check that the list of files in json and zip files are coincide (the same version) I put checksum of zip files in json. If one needs to make sure that the zip file content is the same as in the list, the check of checksum could be turned on (by default it is off). The checksum is not used for verification whether zip file downloaded or not. Although, it is done only once only after downloading. As downloading is a long operation another ms or so for checksum will not be noticeble. In usual mode, only the existence of the files from the list is checked. No checksums.

json and zip files are created simulatenously by `_create_iamdata_content`. I automatized this task in case it will be needed in the future, and because there are a lot of data files. It is problematic to add files to any "model".py file automatically.

> we perform the operation that takes a while on a fake path name first and then rename the fake path

It is cool trick, and it would be enough if we need just download a file and then check the existence of the same file in the future. Initially I implemented it. But soon I realized there is not need in it. We download zip file, but check the existence of unzipped files. Moreover, if the file is corrupted it will not be recognized as zip file (in this case exception will be raised) and it will not be possible to unpack it.

jncots:
Do I understand correctly that there is no need to track down each data file and no need to track versions of zip files? I.e. we check only directory existence?

Then I think the interface of `CachedPath` would look something like:

```python
datdir = CachedPath(path.join(base_path, epos_conf["datdir"]), "epos")

# Later, as fortran takes only str:

self._lib.initializeepos(float(self._seed), k.ecm, str(datdir),...)
```

The CachedPath will look something like:

```python
class CachedPath(Path):
    _flavour = (
        pathlib._PosixFlavour() if os.name == "posix" else pathlib._WindowsFlavour()
    )

    def __new__(cls, local_path, model_name):
        path = super().__new__(cls, local_path)
        if not path.exists():
            model_dict = dict_with_models[model_name]
            model_path = Path(model_dict["path"])
            url = model_dict["url"]
            zip_file = model_path/"zip_file.zip"
            
            if _download_file(zip_file, url):
                if zipfile.is_zipfile(zip_file):
                    with zipfile.ZipFile(zip_file, "r") as zf:
                        zf.extractall(model_path.as_posix())
                    zip_file.unlink()
        if not path.exists():
            raise RuntimeError(f"Something wrong with unpacking")
                
        return path
```

So the logic is: if the path we give is missing then `model_name` dictates what to download and where to unpack it. 

[HDembinski](https://github.com/HDembinski) is this what you have in mind? If so, where is the best place to store the `dict_with_models`?

src/impy/models/dpmjetIII.py
jncots: As I understand models needs paths as strings. Do we need Path only for path construction?

src/impy/models/epos.py
jncots: As we need also url and the path of the directory to unpack the interface should be at least

```python
datdir = CachedPath(path.join(base_path, epos_conf["datdir"]), "epos")
```
If we use the parameter for model ("epos" in this case), we cannot use CachedPath for path construction as:

```python
datdir = CachedPath(base_path, "epos")/ CachedPath(epos_conf["datdir"]), "epos")
```
Then why we need CachedPath? Maybe it is just enough to have a function with interface `((path.join(base_path, epos_conf["datdir"]), "epos")`?

jncots:
After discussion with [afedynitch](https://github.com/afedynitch) I came up with another version of lazy downloading. Now all information is encoded in the name of zip file. For example, "/url/to/Pythia8_v001.zip". This url will be hardcoded for each model in form:

```python
_check_data_dir("url/for/Pythia8_v001.zip")
```

So, as requested [HDembinski](https://github.com/HDembinski) all information in `pythia8.py` file.

`_check_data_dir` checks for existence of the file with name `Pythia8_v001` in directory `iamdata/Pythia8`. If the file is not there, it renames directory `iamdata/Pythia8` to `iamdata/.Pythia8` (in case if something will go wrong), download and unpack "url/for/Pythia8_v001.zip" to `iamdata/Pythia8`. If sucessful, then it deteles downloaded zip file and `iamdata/.Pythia8` and writes file `Pythia8_v001` to `iamdata/Pythia8`.

In the situation when we specify new version, e.g. "url/for/Pythia8_v002.zip", `_check_data_dir` will not find `Pythia8_v002` file and will do the job.

Because we check only version file (`Pythia8_v002`) we don't need CachedPath.

src/impy/iamdata/iamdata_content.json
HDembinski: > The idea was to track all data files belonging to particular model. If, for some reasons, a file is accidentally deleted or corrupted, it will be automatically restored by downloading and unpacking zip file.

Files are not accidentally corrupted and no one deletes files in their Python library path. This is preparing for an eventuality that does not happen in practice. If a feature is not needed, it should not be implemented to reduce maintaince cost downstream. If you adhere to this principle now, you will thank me later.

> Another point, data files could be changed or additional files could be added in the future.

That's also not happening and if it does, we just change the path in `iamdata`. For example, for now Pythia-8.3 uses a folder called `Pythia8`. If it turns out that Pythia-8.4 requires different files, we introduce a folder Pythia84. When you update impy locally, the system will then automatically download the required files next time Pythia8 is used. Again, we don't need checksums to address this. We can add some cleanup code that deletes older Paths in `iamdata` that become obsolete after an impy upgrade if that's an issue.

> json and zip files are created simulatenously by _create_iamdata_content. I automatized this task in case it will be needed in the future, and because there are a lot of data files. It is problematic to add files to any "model".py file automatically.

We don't need to add files automatically, because the list of files is very static. It is not a problem at all to put that static information into the "model".py file. Encapsulation is an important design principle and you are breaking it here.

> It is cool trick, and it would be enough if we need just download a file and then check the existence of the same file in the future. Initially I implemented it. But soon I realized there is not need in it. We download zip file, but check the existence of unzipped files. Moreover, if the file is corrupted it will not be recognized as zip file (in this case exception will be raised) and it will not be possible to unpack it.

The only possible time where the download could be corrupted if because an impatient user hits ctrl+C during download. This trick which is widely used by essentially any downloader (firefox...) can recover from this issue. The logic to do this can be encapsulated in CachedPath, so it is completely transparent for us.

src/impy/models/dpmjetIII.py
HDembinski: Indeed, Path objects are not automatically converted to string when the interface requires it. Whether that is an oversight or intended to make the conversion explicit (which is a good thing), I don't know, but you simply do something like this 
```py
datapath = CachedPath("localpath", "remotepath")
self._lib.init_my_model(str(datapath))
```

HDembinski: Of course, if we never "work" with the CachedPath as an object (use the / operator, check for existence, etc), then you are right that this can be simplified similar to your code to
```
local_path = _cached_path("remotepath")
```
where `_cached_path` is just a function that returns the local path as a string. The local path can be inferred from the remote path. We need some kind of convention anyway on how these paths and files are named. If we need a convention anyway, we can bake it into code.

src/impy/models/epos.py
HDembinski: The CachedPath can look up the base_path automatically, we don't need to do this at the call site. Ideally, this should just be `CachedPath("epos")` or see my comment above regarding the function `_cached_path`

HDembinski:
Sorry, for the redundancy in my comments, I was reading from top to bottom and answering step by step instead of reading everything in one go first.

jncots:
@HDembinski, thank you for review and instructive comments. Some functionalities were indeed redundant. I renamed `_check_data_dir` to `_cached_data_dir`. The usage of `_cached_data_dir` is commented out, because otherwise it will try to download from url which is unknow now. 

Actually, what should be done to get a goal of the branch is to attach zip files to release and delete directories from `iamdata` and from `.gitattributes`. 

Should the branch be merged after release?

HDembinski:
> Actually, what should be done to get a goal of the branch is to attach zip files to release and delete directories from `iamdata` and from `.gitattributes`.
> 
> Should the branch be merged after release?

Can we make a prerelease and then finish the work in this branch? I haven't tested it, but a prerelease should also allow one to attach extra files, so we can use it without adding a dummy release that's essentially a broken package.

Once you have made the prerelease and attached the zip files, we can finish the work in this branch. I think the final goal should be that all models work again after the changes, even on CI.

jncots:
I attached iamdata files to pre-release and deleted it from lfs tracking and repository. The results of tests on CI:

Ubuntu:
```bash
...
FAILED tests/test_epos.py::test_charge - assert 0.3888888888888889 < 0.3
FAILED tests/test_hepmc_io.py::test_hepmc_io[DpmjetIII193] - OSError: error reading event

= 2 failed, 112 passed, 4 skipped, 1 xfailed, 8 warnings in 205.67s (0:03:25) ==

```
@afedynitch, @HDembinski  do you know what are there errors and how to fix the tests?

It is still better than I have on Ubuntu: 

```bash
FAILED tests/test_epos.py::test_charge - assert 0.3888888888888889 < 0.3
FAILED tests/test_to_hepmc3.py::test_to_hepmc3[DpmjetIII306] - AssertionError: assert {(0, 1): [2],...: [5, 7], ...} == {(0, 1...
FAILED tests/test_hepmc_io.py::test_hepmc_io[DpmjetIII193] - OSError: error reading event
FAILED tests/test_to_hepmc3.py::test_to_hepmc3[DpmjetIII191] - AssertionError: assert {(0, 1): [2],...: [5, 7], ...} == {(0, 1...
FAILED tests/test_to_hepmc3.py::test_to_hepmc3[DpmjetIII193] - AssertionError: assert {(0, 1): [2],...: [5, 7], ...} == {(0, 1...
=============================== 5 failed, 109 passed, 4 skipped, 1 xfailed, 8 warnings in 24.99s
```

jncots:
It looks like there is a problem on Mac during file downloading:

```bash
Fatal Python error: Aborted
Current thread 0x0000000109465e00 (most recent call first):
  File "/Users/runner/hostedtoolcache/Python/3.7.15/x64/lib/python3.7/urllib/request.py", line 2630 in getproxies_macosx_sysconf
```

It looks like a problem when urllib is used in multiprocessing on Mac:
* https://bugs.python.org/issue24273
* https://bugs.python.org/issue13829

Maybe `env no_proxy='*' python ...` will help.

jncots:
`env no_proxy='*' python ...` works. Mac has the same failed tests:

```bash
FAILED tests/test_epos.py::test_charge - assert 0.3888888888888889 < 0.3
 +  where 0.3888888888888889 = <function mean at 0x11755d200>(array([False, False,  True,  True,  True,  True,  True,  True,  True,\n        True,  True,  True,  True,  True,  True,  True, False, False,\n       False, False, False, False, False, False, False, False, False,\n       False, False, False, False, False, False, False, False, False]))
 +    where <function mean at 0x11755d200> = np.mean
FAILED tests/test_hepmc_io.py::test_hepmc_io[DpmjetIII193] - OSError: error reading event
= 2 failed, 112 passed, 4 skipped, 1 xfailed, 8 warnings in 189.67s (0:03:09) ==
```

How to fix it?

jncots:
As suggested by @afedynitch, DPMJET was included to the list of models saving only `self.final_state()` in hepmc. Phojet was excluded from this list. However, Phojet still doesn't pass tests:

```bash
FAILED tests/test_hepmc_io.py::test_hepmc_io[Phojet112] - OSError: error reading event
FAILED tests/test_hepmc_io.py::test_hepmc_io[Phojet191] - OSError: error reading event
FAILED tests/test_hepmc_io.py::test_hepmc_io[Phojet193] - OSError: error reading event
```
with error 

```bash
ERROR::ReaderAscii: too few vertices were parsed
ERROR::ReaderAscii: event parsing failed. Returning empty event
```

@afedynitch thinks that this problem is different from the problem of  DPMJET. @HDembinski could you look at this errors? Is it something that could be fixed?

jncots:
The last problem is 

```bash
FAILED tests/test_epos.py::test_charge - assert 0.3888888888888889 < 0.3
 +  where 0.3888888888888889 = <function mean at 0x11755d200>(array([False, False,  True,  True,  True,  True,  True,  True,  True,\n        True,  True,  True,  True,  True,  True,  True, False, False,\n       False, False, False, False, False, False, False, False, False,\n       False, False, False, False, False, False, False, False, False]))
 +    where <function mean at 0x11755d200> = np.mean
```

What about it?

HDembinski:
> As suggested by @afedynitch, DPMJET was included to the list of models saving only `self.final_state()` in hepmc. Phojet was excluded from this list. However, Phojet still doesn't pass tests:
> 
> ```shell
> FAILED tests/test_hepmc_io.py::test_hepmc_io[Phojet112] - OSError: error reading event
> FAILED tests/test_hepmc_io.py::test_hepmc_io[Phojet191] - OSError: error reading event
> FAILED tests/test_hepmc_io.py::test_hepmc_io[Phojet193] - OSError: error reading event
> ```
> 
> with error
> 
> ```shell
> ERROR::ReaderAscii: too few vertices were parsed
> ERROR::ReaderAscii: event parsing failed. Returning empty event
> ```
> 
> @afedynitch thinks that this problem is different from the problem of DPMJET. @HDembinski could you look at this errors? Is it something that could be fixed?

Both DPMJet and Phojet can only save their final_state particles. We should not try to fix this here. To make both pass, the exception in `MCEvent.to_hepmc3()` needs to be applied to both of them for now.

HDembinski:
> The last problem is
> 
> ```shell
> FAILED tests/test_epos.py::test_charge - assert 0.3888888888888889 < 0.3
>  +  where 0.3888888888888889 = <function mean at 0x11755d200>(array([False, False,  True,  True,  True,  True,  True,  True,  True,\n        True,  True,  True,  True,  True,  True,  True, False, False,\n       False, False, False, False, False, False, False, False, False,\n       False, False, False, False, False, False, False, False, False]))
>  +    where <function mean at 0x11755d200> = np.mean
> ```
> What about it?

The threshold of 0.3 is arbitrary, you can just increase it to 0.8 or so. EPOS just generates a lot of intermediate pseudoparticles that `particle` does not know and cannot assign a charge. This test is just there to make sure that not all of the particles are unknown to `particle` because that cannot be right (at least the final state particles must be known to `particle`).

HDembinski:
I am trying this out on my computer. I deleted iamdata manually and then ran the tests. I am getting errors that indicate that the download does not work
```
tests/test_hepmc_io.py::test_hepmc_io[Phojet191]   Data file /Users/hdembinski/Extern/impy/src/impy/iamdata/dpm3191/CT14LL.pds cannot be opened in pho_SetCT14!!
```

HDembinski:
I fixed the tests failing here in https://github.com/impy-project/impy/pull/82. We should close this PR and fix the rest of my comments in https://github.com/impy-project/impy/pull/82, which seems to include this PR.

HDembinski (commented):
I gave a detailed review in #82 which includes these changes done here.

afedynitch:
Since according to @jncots building Pythia requires building an external library first that’s missing on windows (more on this in #82), I’d prefer fixing the test here and merging this PR into main to conclude  its long history.

jncots:
Can I close this PR?

jncots:
> I am trying this out on my computer. I deleted iamdata manually and then ran the tests. I am getting errors that indicate that the download does not work
> 
> ```
> tests/test_hepmc_io.py::test_hepmc_io[Phojet191]   Data file /Users/hdembinski/Extern/impy/src/impy/iamdata/dpm3191/CT14LL.pds cannot be opened in pho_SetCT14!!
> ```

It could happen if tests are run parallel. I tried several times to delete directory. Every time it appears again. 

On Mac there is a problem for `urllib` for forked processes. `env no_proxy='*' python ...`  solves a problem

src/impy/models/dpmjetIII.py
```diff
@@ -81,6 +82,19 @@ def __init__(self, event_kinematics, seed=None, logfname=None):
         dpm_conf = impy_config["dpmjetIII"]
 
         info(1, "Initializing DPMJET-III")
+
+        dpmjet_ver = pathlib.Path(dpm_conf["dat_dir"][self.version]).parts[-1]
+        if dpmjet_ver == "dpm3":
+            _cached_data_dir(
```
afedynitch, line 88: _cached_data_dir should return the data directory for the model given a key from the config, that the model pixks up. The model should not care in which directory exactly the data is stored.
Then it should generate the download link by itself. https://github.. should be in config.

src/impy/models/qgsjet.py
```diff
@@ -81,7 +81,12 @@ def __init__(self, event_kinematics, seed=None, logfname=None):
         super().__init__(seed, logfname)
 
         info(5, "Initializing QGSJET-II")
+        _cached_data_dir(
+            "https://github.com/impy-project/impy/releases/download"
+            "/zipped_data_v1.0/qgsjet_v001.zip"
+        )
         datdir = path.join(base_path, impy_config["qgsjet"]["datdir"])
```
afedynitch, line 88: Same here, get directory from the _Chached_data_dir. Also, no os.path anymore.

src/impy/models/qgsjet.py
```diff
@@ -139,6 +144,12 @@ def __init__(self, event_kinematics, seed="random", logfname=None):
         super().__init__(seed, logfname)
 
         info(5, "Initializing QGSJET01d")
+
+        _cached_data_dir(
```
afedynitch, line 148: same.

src/impy/util.py
```diff
@@ -3,6 +3,13 @@
 import warnings
 import inspect
 import os
+import pathlib
```
afedynitch, line 6: Not needed

afedynitch (commented):
Nice work. I left some comments about simplifying the _cached_data_dir thing. Also please clean up the unused imports util.py.

.github/workflows/test.yml
```diff
@@ -61,4 +61,4 @@ jobs:
         version: 11
     - run: python -m pip install --upgrade pip
     - run: python -m pip install --prefer-binary -v .[test]
-    - run: python -m pytest -vv
+    - run: env no_proxy='*' python -m pytest -vv
```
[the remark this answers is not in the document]
jncots: Put into CI's environmental vars. I check Windows when I fix a problem with Pythin8 building

setup.cfg
```diff
@@ -34,6 +34,8 @@ install_requires =
     scipy
     particletools
     particle
+    tqdm
+    requests
```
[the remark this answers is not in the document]
jncots: Removed

src/impy/models/dpmjetIII.py
jncots: All data regarding paths and urls are now in `config.py` file.  `_cached_data_dir` returns a string (not Path) in format: `/full/path/to/data/dir/", as all models require string. The last slash is critical.

src/impy/models/qgsjet.py
jncots: Done.

src/impy/models/qgsjet.py
jncots: Done

src/impy/models/dpmjetIII.py
HDembinski: We don't need to configure this. The logic that computes the path can be in `_cached_data_dir`. The rule is not complicated. Generate a local directory with the same name as the last part of the remote url inside `impy/iamdata`.

jncots: It was like this, except for version. Maybe I haven't understood correctly, but this is the design which we discussed with @afedynitch.

HDembinski: Then let's discuss it publically, because I don't know what you discussed with Anatoli. I don't want to have a `config.py` file at all. Once this is done, I will work on moving all static config code into the respective models submodules, so that it sits next to the interface code. This is increasing encapsulation and locality of code that belongs together.

Settings that we want to change dynamically should especially not go into a config.py. We call methods on the model class to change behavior or pass options via the `init` if necessary. All dynamic configuration must live in instance variables so that two instances can have different configurations. Currently all models except Pythia8 cannot have two instances in parallel anway, so the difference between instance variables and global variables is mute for those, but Pythia8 exists, too, and therefore the API cannot be designed upfront in this way.

jncots: I think it is easy to change, when you will move configs to the submodules. Actually, `_cached_data_dir` accepts in the end only names of directories `dpm3`, `dpm3191`, `qgsjet`, `epos`, `Pythia8`. They are just taken from config currently. I can move 

```python
data_url = "https://github.com/impy-project/impy/releases/download/zipped_data_v1.0"
data_version = "001"
data_path = Path(__file__).parent.absolute() / "iamdata"
```

to `_cached_data_dir`. And hardcode (or take from the variable in the submodule?) `dpm3`, `dpm3191`... and file names taken from config. 

I can do that , but I would like to be sure that this is what we want in the end.

HDembinski: Yes, what I had in mind was something like this:

```py
class EposLHC:
     def __init__(self, ...):
           _data_dir = _cached_data_dir("https://github.com/impy-project/impy/releases/download/eposlhc.zip")
           # creates local path Path(impy.__file__) / "iamdata" / eposlhc and `_data_dir` is the string to that path
```

Calling _cached_data_dir inside `__init__` is important so that the download is delayed until the point when the user actually creates an instance of the class. If `_data_dir` was a class variable, the download would start already when we import this submodule, which we always do for all models if I remember correctly.

jncots:
In the previous commits I was trying to build Pythia8 on Windows. There were a problem with [dlfcn-win32](https://github.com/dlfcn-win32/dlfcn-win32) library, or to be exactly `dlfcn.h` could not be found. Now I fixed that problem. Unfortunately, now the built library file is not loaded. That is more serious...

jncots:
The main goal of the PR - moving from LFS - has been reached. Another task - build and pass tests on Windows - has been done partially. There are problems with two models Pythia8 and UrQMD. Both models don't load with errors related to dlls. Currently I removed these models from building and tests on Windows.

@HDembinski do you have any idea why Pythia8 build fails on Windows?

I propose to close this PR. @afedynitch, @HDembinski what do you think?

F2Py.cmake
```diff
@@ -171,6 +171,13 @@ function (f2py_add_module target_name)
     PRIVATE ${F2PY_ADD_MODULE_COMPILE_DEFS})
   endif()  
   target_compile_options(${target_name} PRIVATE -cpp)
+  # Link dll statically in Windows
+  # It can be potentially a PROBLEM!!!
+  # But this is the only way found to build a working library on Windows.
+  # However, on Linux and MacOS, it throws an error at the linking stage.
+  if (WIN32)
+    target_link_libraries(${target_name} PUBLIC "-static")
```
afedynitch, line 179: I think this is a hack-ish way to do static linking. Let's postpone this to the CI PR

src/impy/models/dpmjetIII.py
```diff
@@ -78,24 +78,35 @@ def __init__(self, event_kinematics, seed=None, logfname=None):
         self._max_A2 = event_kinematics.A2
 
         self.event_kinematics = event_kinematics
-        dpm_conf = impy_config["dpmjetIII"]
 
         info(1, "Initializing DPMJET-III")
+
+        if self.version == "3.0-6":
```
afedynitch, line 84: These should be class variables. So by subclassing one could override these paths. Like that it's impossible to edit the paths without messing with the source code directly.

src/impy/models/dpmjetIII.py
```diff
@@ -78,24 +78,35 @@ def __init__(self, event_kinematics, seed=None, logfname=None):
...
+            dat_dir_name = "dpm3191"
+            param_file_name = "dpmjpar.dat"
+            evap_file_name = "dpmjet.dat"
+
+        data_dir = _cached_data_dir(dat_dir_name)
+
         # Set the dpmjpar.dat file
         if hasattr(self._lib, "pomdls") and hasattr(self._lib.pomdls, "parfn"):
-            pfile = dpm_conf["param_file"][self.version]
+            pfile = data_dir + param_file_name
```
afedynitch, line 97: Shouldn't that be a path object or so?? Now this totally depends if there is a backslash or not

src/impy/models/phojet.py
```diff
@@ -81,15 +81,24 @@ def __init__(self, event_kinematics, seed=None, logfname=None):
         init_flag = -2 if "dpmjetIII" in self._lib.__name__ else -1
 
         pho_conf = impy_config["phojet"]
+
+        if self.version == "1.12-35":
+            dat_dir_name = "dpm3"
```
afedynitch, line 86: Same. Class vars..

afedynitch (approved):
There are still a few things, which could be improved. Like removal of files or paths from the config because they are not used anymore.
One issue I'd like to see fixed is that the paths to the files in DPMJET, these paths should class vars, so by subclassing one can easily overwrite them.

F2Py.cmake
jncots: Ok. target_link_options can be used instead.

src/impy/models/dpmjetIII.py
jncots: Done

src/impy/models/dpmjetIII.py
jncots: Fortran code requires "/" in the end of directory.

src/impy/models/phojet.py
jncots: Done

HDembinski:
> One issue I'd like to see fixed is that the paths to the files in DPMJET, these paths should class vars, so by subclassing one can easily overwrite them.

I second that.

HDembinski:
> The main goal of the PR - moving from LFS - has been reached. Another task - build and pass tests on Windows - has been done partially. There are problems with two models Pythia8 and UrQMD. Both models don't load with errors related to dlls. Currently I removed these models from building and tests on Windows.

I am fine with that solution. Ideally, Pythia8 and UrQMD should not be importable from impy.models if they are not available on a platform. I am not sure about the easiest and most efficient way to achieve this at the moment. We should not add complicated code to impy to handle this edge case just for a negliglbe user base that runs Windows.

> @HDembinski do you have any idea why Pythia8 build fails on Windows?

No, and I don't think it is worth any more effort in making Pythia8 work on Windows. If Pythia8 is not officially supported on Windows by the Pythia8 authors, we should not try to make it work (unless it was easy to do). The user base on Windows is neglgible.

> I propose to close this PR. @afedynitch, @HDembinski what do you think?

I am ok with that.

HDembinski (approved):

HDembinski (approved):
I think all of Anatoli's last requests are met.

afedynitch:
Yea, let's merge.

On the Pythia8 Windows mystery: it's not related to Pythia8's code but rather the whole python bindings, dynamic libs vs static libs stuff. It's a bit mysterious.

For UrQMD, I agree that the user base is tiny and it's not worth putting any time into it.

PR #81, opened 2022-10-18
Fixes epos interface to to comply with nuclear PDG ids

afedynitch:
This patch fixes two things:

- modifies the functions `InitializeEpos` and `InitEposEvt` to accept only PDG ids instead of A, Z and PDG id. Nuclei are passed via their PDG id.
- Fixes a bug in the GetCharge function, which fails when converting PDG back to nexus IDs for nuclear fragments.


afedynitch:
@HDembinski can I merge this small patch into #77 ? It fixes the EPOS problems with nuclei

HDembinski:
Yes, please. This is great!

HDembinski (approved):

PR #84, opened 2022-10-21
Revision of CompositeTarget

### LINKED ISSUE #80: Revisit of CompositeTarget
CompositeTarget does not work well with Pythia-8, because it randomly switches targets from event-to-event. For a composite target with nuclei A, B, C, which are switched at random from event to event, Pythia that is very slow. But there is a solution. One should instead first simulate all collisions with X A, then X B, then X C. How many need to be simulated per nucleus can be computed from the multinominal distribution.

This is also going to be more efficient for the other models.

jncots:
Fixes #80. The CompositeTarget get a sample of n events from multinomial distribution `component_fractions`  in `_get_sample(n)` and then takes indices of outcome in ordered way: n_a events of Type A, n_b events of Type B.

If generated sample is exhausted `_get_random_AZ` starts to produce (A, Z) in random way. `_get_sample(n)` should be called again to produce another sample for ordered output.

The sample is generated, when event generator calls `__call__(self, nevents)` and the target is CompositeTarget. At rejection, when `nremaining` doesn't change `_update_event_kinematics` doesn't call `_get_random_AZ()`. So nevents will use exactly nevents generated by `_get_sample`


jncots:
Hi, @HDembinski. Could you look at this implementation? The tests/test_comp_target.py passes for it.

src/impy/kinematics.py
```diff
@@ -86,7 +86,7 @@ class CompositeTarget(object):
 
     from numpy.random import default_rng
 
-    def __init__(self, component_list, label="", random_state=None):
+    def __init__(self, component_list, label="", random_state=None, nevents=None):
```
HDembinski, line 89: The composite target should not get `nevents` via the constructor. This breaks the concept. The generator knows how many events to generate, not the EventKinematics object.

src/impy/kinematics.py
```diff
@@ -184,9 +189,47 @@ def _get_maximum_AZ(self):
         max_ind = a_val.index(max(a_val))
         return self.component_A[max_ind], self.component_Z[max_ind]
 
+    def _get_sample(self, nevt):
```
HDembinski, line 192: This should be implemented with a multinomial distribution. https://en.wikipedia.org/wiki/Multinomial_distribution

You upfront how many events are to be generated and the fractions of elements in the composite target. You can then sample random numbers of events per combination from the multinomial distribution.
```py
rng = np.random.default_rng(model_seed)
array_of_nevents_per_fraction = rng.multinomial(nevt, fractions)
```
You should initialize the generator with the model seed, to make this reproducible.

src/impy/kinematics.py
```diff
@@ -184,9 +189,47 @@ def _get_maximum_AZ(self):
...
+        if j < n:
+            while j < n and diff_dist[j] == 0:
+                j += 1
+            if j < n:
+                res = j
+            if diff_dist[j] > 0:
+                diff_dist[j] -= 1
+
+        diff_dist[-1] = j
+        return res
```
HDembinski, line 216: I am not sure what you do here, but this logic should be in MCRun if at all.

src/impy/kinematics.py
```diff
@@ -184,9 +189,47 @@ def _get_maximum_AZ(self):
...
+        random order.
+        """
+        if self._is_ordered:
+            ic = self._get_next_event_index()
+            if ic is None:
+                self._is_ordered = False
+                ic = self.rng.choice(self.ncomponents, 1, p=self.component_fractions)[0]
+        else:
+            ic = self.rng.choice(self.ncomponents, 1, p=self.component_fractions)[0]
         return self.component_A[ic], self.component_Z[ic]
```
HDembinski, line 233: This can probably be removed if you generate the numbers of events from the multinomial distribution.

HDembinski (commented):
It is not a clear-cut case, but I think the logic of how events are generated should reside in MCRun, not in CompositeTarget, because when you need to change this logic, you probably want to change both the normal logic and the special case for CompositeTarget. It is easier to not forget that if they are both in the same function. I think it is the responsibility of MCRun to handle the special case of a CompositeTarget. The CompositeTarget should just be a rather dumb class a data holder with little logic inside.

To generate the number of events per fraction, you should use the multinomial distribution.

src/impy/kinematics.py
HDembinski: This sampling logic is probably also better placed in MCRun. This also solves the issue that CompositeTarget does not know how many events are to be generated, but MCRun does.

jncots: `rng.multinomial` in MCRun now.

src/impy/kinematics.py
jncots: I reverted this function to previous state. It is not used in a process of getting AZ. Maybe it is needed somewhere else? Otherwise it should be deleted.

src/impy/kinematics.py
jncots: Now the logic of returning index is in `_gen_target_ind`. Python generator function is used to save a state.

jncots:
@HDembinski could you look at fixed CompositeTarget?

HDembinski:
Sorry, I missed the review request. I have a feeling that this should be refactored in the future once more, but I don't see any pressing issues with this PR.

jncots:
Hello,@HDembinski. Thank you for reviewing this PR. I checked once again the logic. It  seems to work as intended. The tests pass. If you don't mind, I would like to merge PR. Could you approve it?

HDembinski (approved):

PR #90, opened 2022-11-27
Remove config

### LINKED ISSUE #13: Getting warning from SYBILL "CUT_PRO:   low sqrt(s) "
This is more of a suggestion. The warning indicates that sybill received a value for the center of mass energy which is out of the foreseen range. Maybe the EventKinematics class, or the SIBYLLRun class could warn of this and prevent calls. Instead, at the moment, a wrong setting of the kinematics will yield a large amount of warnings when sampling a large number of events. I found the limits parsing the code , and if they could be part of the config parameters, it would be helpful..

### LINKED ISSUE #31: impy_config needs to go

### LINKED ISSUE #85: Implement warnings in MCEvent.__call__ for too frequent rejections
The current "mechanics" for counting rejections should be improved by warning on too frequent rejections or raise exceptions when the rate of rejections is excessive. The warning from the Pythia8 generate  event call should be removed.

HDembinski:
**This is a large refactoring and cleanup and thus should be merged before any other new PRs are started**.

Closes #13 
Closes #31
Closes #85 

impy.config is removed. We need a config file for the CLI that mimcis the crmc config file, but that's another separate issue.

I fixed bugs along the way and decluttered the interfaces, reduced code duplication and more.
- OutputGrabber and _attach_log methods and friends have been removed. We need a different solution to catch output from the Fortran codes. I have some ideas, which will be explored in future PRs. For now, all models print to stdout.
- Models now use a unified system to check whether input particles are valid, based on attributes `_projectiles` and `_targets`. Models can override the defaults set in MCRun.
- Models also use a unified system to check whether the requested kinematics is valid. Most models have a minimum sqrt(s) of 10 GeV. We need to catch this in Python, because FORTRAN typically calls STOP when the energy is too low, which aborts the Python interpreter or kills the Jupyter kernel.
- CompositeTarget had a separate random number generator and a separate seed, that was bad, because we need a single source of randomness if we want runs to be reproducible. It still uses a seperate random number generator, but at least the seed is now sync'ed with that of the model. Generally speaking, all random sampling must happen in MCRun and should go through RMMARD. This is currently not the case when you use a CompositeTarget.
- The output frame is now set via the `EventFrame` enum class, instead of a string.
- The output frame is defined by the input frame and stored in `EventKinematics`. Users can override the frame to get the output in another frame as the input. If the model supports output in different frames, we use that functionality, otherwise boost ourselves in Python. The model also has a frame, but this is the frame in which the output is actually generated, not the frame that the user requested. Users can change the frame in `EventKinematics`, but in case of the generator it must be read-only. The output frame was previously a mutable global variable, shared between all models, which was a source of potential bugs.
- `EventKinematics` stores the particles as `PDGID` from `particle` instead of a plain int, because `PDGID` has useful properties, like .Z and .A and .is_nucleus that we duplicated before. This allows us to drop some duplicated code and related (not well-designed) API.
- `MCRun.event_kinematics` was shortened to `MCRun.kinematics`
- `MCRun._generate_event` was shortened to `MCRun._generate`
- `MCRun.cross_section` now also accepts a `CompositeTarget` and computes the correct cross-section.
- EPOS initialization was simplified, some variables were initializated twice. Frame is now set in `_set_kinematics`, because a new `EventKinematics` may require changing the output frame.
- I removed the option to give names to components in the `CompositeTarget` names. This is not adding value and just makes the code more complex. I also removed the API to add components, which was just making the design more complicated. The user can always generate the list of components and fractions with standard Python collections then pass the list to the constructor of `CompositeTarget`. We should not provide an API for trivialities, especially when it makes our design more complicated (it was not clear when to normalize the input fractions when incrementally adding components is allowed).
- `test_generators.py` now checks generation with composite targets and all supported output frames. The check is much more elaborate now by including functionality previously in `fixme_all_cms.py`, `fixme_all_lab.py`, `test_composite_target.py`, which were all removed. A plot is automatically generated for visual inspection locally and uploaded to GitHub when the test runs in CI.
- `test_setstable.py` (formerly `fixme_setunstable.py`) was added and improved, now it tests all models and exposes those which do not support setstable.
- `fixme_eposlhc.py` was removed, because it was obsolete


HDembinski:
@jncots @afedynitch I am done with the changes and I fixed a lot of issues, but now I need your help. 
1) The new more exhaustive test_generators.py fails on CI, which is bad because the models should produce identical output on different machines if the seed is fixed. Could you please run `test_generators.py` locally on your computers and check what is going on? It works locally for me.
2) UrQMD: I somehow broke the test_rng_state test with UrQMD, but I cannot figure out why. UrQMD used to pass this test. I did not change anything in regard to rmmard as far as I am aware. So I don't understand what is happening. @jncots Since you made the changes to rmmard, could you please have a look?

HDembinski:
> I ran into this after a fresh checkout. The race conditions were severe for file downloads and the tests were stuck on re-downloading the files. I had to do it manually... If there is a way to trigger the downloads before any of the parallel tests starts, that could be a solution.

Ok, fortunately that's easy to work around. This problem only occurs once when you to a fresh install, so you can run the tests once sequentially.

It would be great if the file downloads would check that another download of the same file is currently in progress and then just block until the download is complete. But I am not sure how this can be achieved. The current implementation of `_cached_data_dir` always deletes the temp_dir, which is the right way to recover from a failed/aborted download attempt. However, to avoid the parallel downloads of the same file, one would have to not delete the temp_dir, but block until it disappears. The two behaviors are opposite and in conflict. If we attempt to fix this, it has to be a robust solution and not some brittle hack. To distinguish the two cases, the process needs to be able to figure out that the other process which writes the temporary directory is still alive and does something. Perhaps the processing which starts the download first can put its process ID into a file into that directory, so that another process can check whether that process still exists. It is not a perfect solution, because process IDs are not collision-free between boots. A perhaps better option is to monitor the temp directory for about 10 seconds to see whether the file size inside changes. That would indicate that there is a download in progress.

jncots:
> The new more exhaustive test_generators.py fails on CI, which is bad because the models should produce identical output on different machines if the seed is fixed.

I checked on my machine. `18 failed, 102 passed, 24 skipped`. The results on plots look similar, but details are different. So yes, random numbers are probably different.

> UrQMD: I somehow broke the test_rng_state test with UrQMD, but I cannot figure out why.

The tests for rng don't work for UrQMD. The problem is that UrQMD consumes some portion of random numbers somewhere during the initialization step. The next steps produce the same sequence of random numbers after resetting of rng state, but not the first one. I could not find the code that consumes random numbers. If I remember correctly it should be located somewhere in the Pythia part of code.

jncots:
[HDembinski](https://github.com/HDembinski), I looked through the files, and found out that `[_eposlhc.pyf](https://github.com/impy-project/impy/blob/main/src/f2py/_eposlhc.pyf)` files is recently regenerated. Did you have a problem with previous versions? I am asking this, because I always should regenerate these f2py files, otherwise Epos-LHC library has import error. I wonder, could it be that f2py files generated on different machines could be incompatible?

HDembinski:
> I checked on my machine. `18 failed, 102 passed, 24 skipped`. The results on plots look similar, but details are different. So yes, random numbers are probably different.

How can they be different when you use RMMARD on every platform?

HDembinski:
> [HDembinski](https://github.com/HDembinski), I looked through the files, and found out that `_eposlhc.pyf` file is recently regenerated. Did you have a problem with previous version? I am asking this, because I always should regenerate these f2py files, otherwise Epos-LHC library has import error. I wonder, could it be that f2py files generated on different machines could be incompatible?
> 
> **Edit**: Just realized that the interface for epos has changed. That's the reason why it is changed. But still, for some reason I need to regenerate it on local machine to make it work.

I changed the interface of how EPOS is initialized, as explained in PR description. This sounds similar to the other problem you had a while ago when I had regenerated some f2py files, no?

HDembinski:
So what do we do to proceed? I will try to fix the failing tests by asking not for exact equality but for statistical equality. That way, we can still recognize drastic deviations from the references. @jncots @afedynitch Anything else that needs to be addressed?

afedynitch:
I had a couple of minor issues when testing. I could reproduce the failing tests on both the ARM Mac and AMD Zen2 linux systems. The produced figures looked also different from the default. I wanted to look into more details but due to travels my on-keyboard time with different systems is still very limited.

Here are some issues I ran into on

AMD Threadripper 3970x, Ubuntu 20.04, gcc 9.4, Python3.9, numpy 1.19.2

  *   /_eposlhc.cpython-39-x86_64-linux-gnu.so: undefined symbol: Npy__PyLong_AsInt. Removing and recreating the f2py files fixed this.

  *   When building in editable mode, pip builds in /tmp despite --no-build-isolation. Then pythia8 fails because _libpythia8 is not found. This was also the case on my Mac with a different numpy, python3.11 setup.
  *   When using CLI "impy -m 4 -s 1000 -n 100" only the help message is returned but no error message. So it's not clear what it misses. But what it misses is the -[capital]S

These must be not directly related to the PR but this is what I had thus far. I’d like to understand if the reason for the different histograms is related to the RND or to something else in the models before proceeding. If the RND generator is the origin, we should consider switching to something else a future PR. I can do some more testing tomorrow morning (JST).



________________________________

HDembinski:
> * /_eposlhc.cpython-39-x86_64-linux-gnu.so: undefined symbol: Npy__PyLong_AsInt. Removing and recreating the f2py files fixed this. 
> * When building in editable mode, pip builds in /tmp despite --no-build-isolation. Then pythia8 fails because _libpythia8 is not found. This was also the case on my Mac with a different numpy, python3.11 setup.

Is is this a problem of your setup or in the code? These issues do not occur on CI.

HDembinski:
> * When using CLI "impy -m 4 -s 1000 -n 100" only the help message is returned but no error message. So it's not clear what it misses. But what it misses is the -[capital]S

Confirmed, will be fixed. When you find something like this, you are welcome to add a test immediately that demonstrates the bug. There are many examples in test_cli.py

HDembinski:
> These must be not directly related to the PR but this is what I had thus far. I’d like to understand if the reason for the different histograms is related to the RND or to something else in the models before proceeding. If the RND generator is the origin, we should consider switching to something else a future PR. I can do some more testing tomorrow morning (JST).

@jncots Since you worked on the RNG, could you add a test which compares the first 10 integers produced by RMMARD with a fixed reference, to rule out that it is a problem with the generator being platform-dependent?

jncots:
> @jncots Since you worked on the PRNG, could you add a test which compares the first 10 integers produced by RMMARD with a fixed reference, to rule out that it is a problem with the PRNG being platform-dependent?

I added a test `test_simrnd()` to test_rng_state.py. It compares states of RMMARD and 10 first random numbers generated by `simrnd` function for a several seeds with reference data located in `test_simrnd_data` file.

HDembinski:
@jncots Can you please add a new compiled extension to test the generator, which contains code to test RMMARD. How you did at the moment is a hack, adding a function to sibyll to get the generator output. There should not a be a need to start a model to test the PRNG.

HDembinski:
@jncots @afedynitch I fixed most of the failing tests by switching to a statistical comparison instead of an exact comparison. I further improved the control plots which are generated by `test_generators.py` and changed our CI scripts so that these plots are uploaded. You can download them for example here (scroll to the bottom) https://github.com/impy-project/impy/actions/runs/3725026763.

There are now only two tests that fail on CI (and pass locally on my machine). Thanks to the upload feature, I can now study the corresponding comparison plots, here they are:

DpmjetIII191_pi-_air_cms

UrQMD34_pi-_p_ft

Both look ok by eye. I cannot spot what is wrong with the DPMJet case. In case of UrQMD34, the failure is clearly related to the peak in the proton distribution around eta = -2, where the reference predicts nothing, while the test produced 13 (anti-)protons. That seems indeed fairly unlikely to occur by random chance.

HDembinski:
The test for RMMARD that @jncots added seems to indicate that generator produces identical streams of random numbers on all machines. It is puzzling to me that the results of the generators still differ from machine to machine. The sizes of integers should be the same everywhere, as all are 64 bit architectures. Basic floating point math is standardized, you should get the exact same result when you add two floats or compute other basic functions that are implemented with CPU instructions.

It is possible that we see differences due to different versions of gfortran, which may use different optimizations and also the standard library implementations may be different that are not implemented directly via CPU instructions. I am locally using gfortran-12, while we use gfortran-11 on CI.

jncots:
> @jncots Can you please add a new compiled extension to test the generator, which contains code to test RMMARD. 

Functions related to rmmard are added as a separate extension. Although I am not sure whether we should mix model extensions  with 'ancillary' extensions.

HDembinski:
> Functions related to rmmard are added as a separate extension. Although I am not sure whether we should mix model extensions with 'ancillary' extensions.

This is fine, an extension is an extension, there is no gain in restricting ourselves to model extensions. Plus, this is going to be temporary if we replace RMMARD with the Numpy PRNG, which I argue for in #104

jncots:
> In case of UrQMD34, the failure is clearly related to the peak in the proton distribution around eta = -2

It looks like all the plots on CI for different systems are identical. So it seems the results don't depend on operating system.

I compared a couple of plots obtained on a local machine (linux, gfortran 11). They were identical to that on CI. 

Then I updated gfortran to version 12.2.0.  With this version the same happened. All the plots that I checked are identical. The only difference is that some plots don't appear on my local machine: 267 (CI) vs 237 (local) plots. Unfortunately, I haven't noticed how many plots were with gfortran 11.

What I noticed is that all the plots related to Pythia8 disappeared on local machine. But why it is a problem on CI, if the problem only with gfortran?

At least some problems can be related to different runtime libraries (https://stackoverflow.com/a/58930551), which is not directly related to gfortran itself.

HDembinski:
Nice work, all that confirms the hypothesis that I formulated above.

> At least some problems can be related to different runtime libraries (https://stackoverflow.com/a/58930551), which is not directly related to gfortran itself.

Yes, that's what I meant with "standard library implementations may be different".

> What I noticed is that all the plots related to Pythia8 disappeared on local machine. But why it is a problem on CI, if the problem only with gfortran?

I am not sure what you mean. The C/C++ standard library implementation from GCC may also differ between v11 and v12, so also Pythia8 results can be different. I cannot tell you why Pythia8 plots are not generated on your computer, what does the pytest output say? Plots are generated unconditionally for all tested generators.

HDembinski:
@jncots https://github.com/impy-project/impy/pull/90/commits/82f6a4385b6f8074809753fd3ab4410e5fa6768c is breaking the compilation on CI. I suggest to remove the new code that tests the PRNG altogether, since we all agree that we want to replace RMMARD with the Numpy generator.

jncots:
>> When building in editable mode, pip builds in /tmp despite --no-build-isolation. Then pythia8 fails because _libpythia8 is not found. This was also the case on my Mac with a different numpy, python3.11 setup.

> Is is this a problem of your setup or in the PR? This should be always clearly stated. These issues do not occur on CI, so I think it is not a problem with this PR?


After update to pip=22.3.1 I also have the same problem. However, installation to pip=21.0 solves the problem. In pip 22.3.1 the developers removed some deprecated functions, so building occurs in temporary directory. After building the module files are copied to `models` directory, except for `libpythia8`. `_pythia8` module referencies this file, which is deleted with temporary directory after installation.

I don't know whether it is possible to restore the behaviour of pip 21.0 with pip 22.3.1, but  setuptools should at least copy `libpythia8` along with `_pythia8`. Moreover, in Linux, `_pythia8` should be patched to refer to correct path.

jncots:
> @jncots https://github.com/impy-project/impy/commit/82f6a4385b6f8074809753fd3ab4410e5fa6768c is breaking the compilation on CI. I don't think it is worth spending time to fix this, instead I suggest to remove the new code that tests the PRNG altogether, since we all agree that we want to replace RMMARD with the Numpy generator.

Could it be that change numpy-1.23.5 to numpy-1.24.0 on CI is the problem? MacOS still works and it uses older version of numpy-1.21.6.

HDembinski:
> Could it be that change numpy-1.23.5 to numpy-1.24.0 on CI is the problem? MacOS still works and it uses older version of numpy-1.21.6.

Yes, I think you are spot on. It looks like our assumption, we can use static f2py wrapper files, is invalid.

HDembinski:
@jncots Btw, I checked locally whether I can reproduce the failures on CI when I compile impy locally with gcc-11 like on CI. I cannot reproduce the failures. So it rather seems that the differences are rather coming from M1 vs Intel.

HDembinski:
The work on this PR is finished now and it is ready for review. It closes three open issues and brings many improvements to the framework and greately expands the tests. The tests revealed some issues that need to be addressed in future PRs. I opened several new issues that we addressed next. The most involved one is to replace RMMARD with Numpy's PRNG.

src/impy/kinematics.py
```diff
@@ -99,164 +33,77 @@ class CompositeTarget(object):
...
-        return self.component_A[ic], self.component_Z[ic]
+    @property
+    def is_nucleus(self):
+        return True
 
-    def __str__(self):
-        ostr = "Composite target '" + self.label + "' \n"
+    def __int__(self):
+        """Return PDGID for heaviest of elements."""
+        return max(int(m) for m in self._components)
```
jncots, line 71: So, the higher is int(m), the larger is mass? I just cannot understand connection between mass and pdg.

jncots (commented):
Hi, @HDembinski, the PR is a huge improvement in a structure and readability. Especially I like how the structure of  kinematics.py is changed. Thank you. I haven't finished yet the review of all the introduced features, so there will be follow up.

src/impy/kinematics.py
HDembinski: It is now fixed.

CMakeLists.txt
```diff
@@ -541,7 +537,7 @@ f2py_add_module(_dpmjetIII193
   pho_harmci pho_fitout pho_mcini pho_ptcut pytune pho_rregpar pho_sregpar
   pho_prevnt ipho_pdg2id ipho_id2pdg pho_harint pho_harxto pho_harxpt
   pho_setpcomb dt_phoxs dt_xshn dt_flahad dt_title pho_ghhias
-  init_rmmard ${impy_logging}
+  init_rmmard simrnd ${impy_logging}
```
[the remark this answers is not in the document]
HDembinski: Fixed. And a couple of other inconsistencies, that seem to originate from the revert that I did.

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.18)
+cmake_minimum_required(VERSION 3.17)
```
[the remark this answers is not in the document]
HDembinski: Yes, 3.17 works, too, and I needed to compile impy on our cluster, which only has cmake 3.17.

src/impy/models/dpmjetIII.py
```diff
@@ -112,99 +105,84 @@ def __init__(self, event_kinematics, seed=None, logfname=None):
             self._lib.pomdls.ipamdl[178] = 5000
 
         # Relax momentum and energy conservation checks at very high energies
-        if k.ecm > 5e4:
+        if evt_kin.ecm > 5e4:
             # Relative allowed deviation
             self._lib.pomdls.parmdl[76] = 0.05
             # Absolute allowed deviation
             self._lib.pomdls.parmdl[77] = 0.05
```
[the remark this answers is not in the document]
HDembinski: I have no idea how dpmjet works, but it is setting two different numbers in the array, so it is not duplicated? Or what do you mean?

src/impy/models/sophia.py
```diff
@@ -41,118 +44,65 @@ class Sophia20(MCRun):
...
+        self._lib.initial(self._nucleon_code)
 
     def _set_stable(self, pdgid, stable):
-
-        # Do not use global stable_list
-        if not impy_config["sophia"]["use_stable_list"]:
-            return
-
-        sid = abs(self._lib.icon_pdg_sib(pdgid))
-        if abs(pdgid) == 311:
```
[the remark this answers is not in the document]
HDembinski: I don't know what you mean. I just simplified the code that was there. It should be equivalent, but eventually, an expert (Anatoli) should check that the changes have not accidentally introduced bugs. Refactoring without unit tests that check whether the output of some code remains the same, is very dangerous. That's why I expanded test_generators.py, so at least in the future we can track whether some refactoring broke something.

src/impy/models/sibyll.py
```diff
@@ -103,43 +116,14 @@ def _cross_section(self, evt_kin):
...
-        elif abs(k.p1pdg) == 211:
-            sigproj = 2
-        elif abs(k.p1pdg) == 321:
-            sigproj = 3
-        else:
-            info(0, "No cross section available for projectile", k.p1pdg)
-            raise Exception("Input error")
-        sigma = self._lib.sib_sigma_hair(sigproj, k.ecm)
+        sigma = self._lib.sib_sigma_hair(self._cross_section_id, self._ecm)
         if isinstance(sigma, tuple):
```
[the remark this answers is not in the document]
HDembinski: Thanks, I fixed it, but this code should be removed in the near future, we have a general mechanism to compute cross-sections for composite targets now.

src/impy/util.py
```diff
@@ -492,6 +424,15 @@ def get_all_models(skip=None):
...
 
+class Singleton(type):
+    _instances = {}
+
+    def __call__(cls, *args, **kwargs):
+        if cls not in cls._instances:
+            cls._instances[cls] = super(Singleton, cls).__call__(*args, **kwargs)
+        return cls._instances[cls]
+
+
```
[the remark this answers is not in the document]
HDembinski: It is not in use yet. This is a metaclass, it changes how classes are created. One idea is to use this to make all models singletons, so you can create multiple instances of a model (which are then actually the same instance).

HDembinski:
The failure on windows is strange.
https://github.com/impy-project/impy/actions/runs/3759309155/jobs/6388722811#step:10:1824
It does not find tk/tcl. We are not directly using that, I assume that matplotlib is using it internally. The tests which fail now have passed before, and I didn't touch code in one of the QGSJets, so this seems like a flaky test.

HDembinski:
@jncots Can we resolve some of the open conversations?

CMakeLists.txt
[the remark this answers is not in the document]
jncots: I see.

src/impy/models/dpmjetIII.py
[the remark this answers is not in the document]
jncots: I meant that, for example, `self._lib.pomdls.parmdl[77] = 0.395` and then `self._lib.pomdls.parmdl[77] = 0.05`. @afedynitch, is this code duplication or not?

src/impy/util.py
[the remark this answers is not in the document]
jncots: That would be great.

CMakeLists.txt
```diff
@@ -1,6 +1,19 @@
...
-project(impylib LANGUAGES C CXX Fortran)
+project(impylib LANGUAGES Fortran)
+
+# Find C/CXX compilers that match the Fortran compiler
+if (CMAKE_Fortran_COMPILER_ID STREQUAL "GNU")
+  # Compiler version is something like XY.Z.N, we want only XY
+  string(REGEX MATCH "[0-9]+" COMPILER_MAJOR_VERSION ${CMAKE_Fortran_COMPILER_VERSION})
+  find_program(CMAKE_C_COMPILER "gcc-${COMPILER_MAJOR_VERSION}")
+  find_program(CMAKE_CXX_COMPILER "g++-${COMPILER_MAJOR_VERSION}")
+else()
```
jncots, line 11: What if we need to use clang for C/C++?

HDembinski: Then you add some code to handle clang here. This is just for convenience when you use gcc, it does not have to work for everything.

.github/workflows/release.yml
```diff
@@ -77,13 +77,12 @@ jobs:
...
     if: ${{ github.ref == 'refs/heads/master' }}
     steps:
-      - uses: actions/download-artifact@v2
+      - uses: actions/download-artifact@v3
         with:
           name: artifact
           path: dist
 
-      - uses: pypa/gh-action-pypi-publish@master
+      - uses: pypa/gh-action-pypi-publish@release/v1
```
afedynitch, line 85: Do we want to publish from master? I'm publishing MCEq from the releases keep track of what's exactly on pypi.

pyproject.toml
```diff
@@ -3,7 +3,7 @@ requires = [
     "setuptools>=42",
     "wheel",
     "cmake",
-    "numpy",
+    "numpy<1.24",
 ]
 build-backend = "setuptools.build_meta"
```
afedynitch, line 8: The latest pip versions have broken the legacy --no-build-isolation behavior that implied that the source is built under the "build" directory in the source tree. 

This broke incremental builds on `-e` installations, since the the object files in `/tmp` are deleted. One can get the old behavior back by passing `--no-use-pep517` to pip, which doesn't work because of this line. What is the reason to use this specific backend?

src/impy/models/dpmjetIII.py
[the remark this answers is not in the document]
afedynitch: Actually, it was setting the wrong variables. I checked this and prefer these new values

src/impy/constants.py
```diff
@@ -50,4 +53,96 @@
...
+    -2112,
+    -13,
+)
+
+nuclei = {
+    p.pdgid for p in Particle.findall(lambda p: p.pdgid.is_nucleus and p.charge > 0)
+}
+
+# only positive PDGIDs!
+standard_projectiles = {
```
afedynitch, line 87: That's a bit restrictive. DPMJET, PHOJET and SIBYLL2.3 can support way more projectiles including charm, lambda, Xi, ... as it was in the config. Also DPMJET supports all nuclei as projectile and target. Not only those, which are defined in Particle but also all arbitrary nuclei as long as the internal formulas to calculate the potential work.

I changed in `dpmjet.py` the projectiles to `standard_projectiles | nuclei`, but still as it is now there is are more projectile particles beyond the standard ones missing in DPMJET and PHOJET.

src/impy/models/pythia8.py
```diff
@@ -56,46 +61,38 @@ def __init__(self, event_kinematics, seed=None, logfname=None):
...
-            raise ValueError(f"Particle {k.p2pdg} not recognized")
-
         config = [
             "Random:setSeed = on",
             f"Random:seed = {self._seed}",
-            # Specify energy in center of mass
+            # use center-of-mass frame
             "Beams:frameType = 1",
             "SoftQCD:inelastic = on",
+            "PhotonParton:all = on",
```
afedynitch, line 89: Which process does require this setting to be true?

src/impy/models/sibyll.py
```diff
@@ -64,32 +85,24 @@ def __init__(self, event_kinematics, seed=None, logfname=None):
...
-        elif abs(evt_kin.p1pdg) == 321:
-            sigproj = 3
-        else:
-            raise ValueError(
-                f"No cross section available for projectile {evt_kin.p1pdg}"
+    def _cross_section(self):
+        kin = self.kinematics
+        if kin.p2.A > 1:
+            # TODO figure out what this returns exactly:
+            # self._lib.sib_sigma_hnuc
```
afedynitch, line 96: It should be ok to use. The docs says:

output:      SIGprod    particle production cross section (mb)
C                  SIGbdif    q.ela and ela beam diff. cross section
C                  SIGela     elastic cross section

.github/workflows/release.yml
HDembinski: I don't understand the question. It is generally recommended to not user master, but releases instead for all Github Actions.

src/impy/models/dpmjetIII.py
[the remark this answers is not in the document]
HDembinski: So can this conversion be closed or is there something to fix in the code?

.github/workflows/release.yml
afedynitch: Sorry I confused the + and - :)

src/impy/models/dpmjetIII.py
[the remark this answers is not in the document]
afedynitch: I fixed it but didn't push the changes, yet. I'll resolve it

src/impy/models/sophia.py
```diff
@@ -41,118 +44,65 @@ class Sophia20(MCRun):
...
-
-        # self.energy_of_photon is the energy in lab frame
-        # where nucleon is at rest and photon is moving
-        total_crossection_id = 3  # 3 is for total crossection
-        # cross section in micro barn
-        total = self._lib.crossection(
-            energy_of_photon, total_crossection_id, nucleon_code_number
+    def _cross_section(self):
+        # code=3 for inelastic cross-section
+        # TODO fill more cross-sections
```
afedynitch, line 68: The cross sections for photo-hadronic processes are different from the usual ones defined in the CrossSectionData (resonances, multipion, direct, etc.). Do you want to keep the class generic then there is no need to do more. Any preferences?

pyproject.toml
HDembinski: Why does upgrading pip has anything to do with this restriction to use numpy<1.24? What you mention about `build` is an issue, but it seems orthogonal to this. See https://github.com/impy-project/impy/pull/90#issuecomment-1359030501

If you install numpy-1.24, the build on CI breaks, @jncots discovered this. https://github.com/impy-project/impy/pull/90#issuecomment-1359080327

afedynitch: I meant the build-backend line

src/impy/constants.py
HDembinski: Feel free to add everything that is possible here or in another PR, but please also add some tests for those cases to check that these projectiles really work.

src/impy/models/pythia8.py
HDembinski: Currently nothing, but I would like to enable photo-hadron interactions for Pythia8.

src/impy/constants.py
afedynitch: It should be part of this PR because it removes the config and the config had the accepted projectiles properly defined for every model and version.

src/impy/models/sophia.py
HDembinski: I would keep it as it is.

impy is primarily a frontend for h-h, h-A, A-A interactions. This means the CrossSectionData should cover those cases. It is useful to further split the inelastic cross-section into non-diffractive and diffractive parts. g-h and g-A are also interesting to support, but I don't think we need to split the inelastic cross-section into parts for these interactions.

src/impy/constants.py
HDembinski: > Not only those, which are defined in Particle

 Which particles do you need which are not included in Particle? `Particle` already contains many nuclei, including rare isotopes. We can support other particles. We already added our own "name2pdg" direct, which can map particle names not included in `Particle` to be used as input.

src/impy/models/pythia8.py
afedynitch: I don't recommend setting this to on because it [restricts the available PDF](https://pythia.org/latest-manual/ElectroweakProcesses.html#section2) selection and requires some thinking by the user what they want to do. The default real photon collisions should work without this flag since the relevant switch `Photon:ProcessType` is on by default.

src/impy/constants.py
afedynitch: For the nuclei DPMJET, and probably EPOS support very exotic configs such as 40 protons and 5 neutrons, or 10 neutrons and 100 protons. Although these don't even exist in nature, such crazy configs can be in principle produced as a remnant. The point is that we shouldn't restrict the models below their capability.

So how can I construct and use a target nucleus with 5 neutrons and 25 protons within the current framework?

pyproject.toml
HDembinski: I don't know how to fix this. On CI, setuptools still works. I also want incremental builds so if you have an idea how to fix this, please go ahead. It is not related to this PR, though, so I would prefer if we discuss this in an issue. I did not change the build-backend in this PR.

src/impy/constants.py
HDembinski: Let me rephrase what you are saying. You think the current system is too limited for nuclei, because some models support nuclei with arbitrary number of protons/neutrons, but the current implementation uses a list of nuclei, and that list only contains some nuclei which actually occur in nature. I think for impy that's fine, why would you want to simulate a target which does not exist in nature?

Model can produce an exotic nuclei as output, this is unrelated to the list of nuclei in impy.constants, which is only used to valide the target that the user provides. A target must be a stable nucleus, so I don't see this is as a problem.

src/impy/models/pythia8.py
HDembinski: Seems you are right. I thought I tested photons without this and got failures, but I just checked that it works without. I will remove this line then.

HDembinski:
@jncots @afedynitch I just switched to Pythia8.308 in this PR, because that's the first version which can be accessed directly on gitlab. This means we can drop our distribution of Pythia8 and include it as a submodule instead, which is much better. I already made this change here and also updated the data file for Pythia8.

src/impy/constants.py
HDembinski: > It should be part of this PR because it removes the config and the config had the accepted projectiles properly defined for every model and version.

I tried to copy the info from the old config regarding allowed projectiles, but I didn't want to list a bunch of PDGIDs and so I came up with these named sets that you can extend. If I made a mistake in that transition, we should correct that.

The old config used a list of projectiles and the "int_support" field, which declared what kind of collisions are supported. I thought I could replace this sets of PDG numbers, but it does not work for arbitrary nuclei. However, some models do not support arbitary nuclei, see SIBYLL and Pythia8. So we need something that allows to restrict the range of nuclei for those models.

HDembinski: If you think that supporting nuclei which do not exist is important, then we can introduce a special "Nuclei" object, which specifies the range of Z and A that the model supports and which is then used in the validation of the input instead or in addition to a list of particles. It could look something like this:

```
class Model
   _targets = {211, 2212, Nuclei(A=(1, 40), Z=(0, 20)}
```

HDembinski: I implemented this. Arbitrary nuclei are now accepted with the class `Nuclei` which haves like a set and can be combined with sets of PDGIDs.

src/impy/models/dpmjetIII.py
```diff
@@ -170,7 +170,9 @@ def _generate(self):
             k.p1.Z or 0,
             k.p2.A or 1,
             k.p2.Z or 0,
-            self._lib.idt_icihad(k.p1),
```
HDembinski: @afedynitch The tests are now failing again on CI since this change. It looks like you regenerated all the references for Dpmjet and phojet, so it should be consistent. Any idea why this is failing?

pyproject.toml
HDembinski: @afedynitch Is this resolved now by your change of commenting out build-backend?

pyproject.toml
```diff
@@ -5,7 +5,7 @@ requires = [
     "cmake",
     "numpy<1.24",
 ]
-build-backend = "setuptools.build_meta"
+#build-backend = "setuptools.build_meta"
```
HDembinski, line 8: According to these doc, removing the explicit line does not do anything. Can we please revert this? Explicit is better than implicit.

https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/

HDembinski: Nevermind, it took a bit to understand that `--no-use-pep517` requires that this line is commented out. Which is very strange.

pyproject.toml
afedynitch: Yes! :)

src/impy/common.py
```diff
@@ -610,21 +618,24 @@ def __call__(self, nevents):
...
+                if nretries % 50 == 0:
                     warnings.warn(
-                        "Event was rejected 5 times in a row. The event generator "
+                        f"Event was rejected {nretries} times in a row. The generator "
                         "may be misconfigured or used outside of its valid range",
                         RuntimeWarning,
                     )
-                if nretries > 20:
-                    raise RuntimeError("More than 20 retries, aborting")
+                if nretries > 1000:
```
afedynitch, line 637: That's really quite high. Which case were you running that required such a high number or retries? Pythia with nuclei?

src/impy/models/dpmjetIII.py
afedynitch: This is indeed weird. The tests pass on my AMD and M1 Mac. I generated the files on the AMD under Linux. The only thing I can imagine doing at the moment is to generate them again on the Mac and try again. 

Ultimately, we/I/Anton need to figure out what causes the RND sequence divergences.

src/impy/common.py
HDembinski: Yes, I was playing with Pythia8 with nuclei. I do not understand what is happening. I got it to work for a while but then it stopped working again, both times consistently over many runs. For now, I would rather have this value large so that we have a chance to get valid Pythia8 events.

src/impy/models/dpmjetIII.py
HDembinski: The RND sequences themselves are consistent between platforms and machines, unless I missed something, the tentative conclusion was that the differences arise from different stdlib implementations on different platforms/compiler versions. I think we have to live with this and use a very lenient threshold for the p-value in the test.

afedynitch (approved):
This is another epic PR. Lot's of changes. Probably too many to check by hand. If the tests pass on the CI, it can be merged as a big step toward a stable release!

Thanks. Great work. I learned also quite a lot from reading the code.

HDembinski:
@afedynitch Thank you for the approval, but to merge this all the tests need to be green. I am concerned about the failures of DPMJetIII193 on Windows, where the model apparently crashed.
https://github.com/impy-project/impy/actions/runs/3806299117/jobs/6474973368#step:10:12167
This should not happen and needs to be further investigated.

The failure on Macos is less severe, because it can be fixed by making the acceptance criterion even more lenient (although it is already very lenient):
https://github.com/impy-project/impy/actions/runs/3806299117/jobs/6474973448#step:10:4105

afedynitch:
I didn’t find any indication of a crash in the logs. The model now supports projectile nuclei and depending on what exactly happens in event generation the run time to run additional initializations is quite significant for the first couple of events. So it just timed out at least for the he-air tests.

The first test fails because of `PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\hostedtoolcache\\windows\\Python\\3.10.9\\x64\\lib\\site-packages\\impy\\iamdata\\dpm3191_v001.zip`.

 Tests on windows pass on my machine. I’ll simply restart them. The file download errors are unavoidable by design when multiprocessing. So we should consider triggering them at the end of the installation “manually”, or so.

HDembinski:
> I didn’t find any indication of a crash in the logs. The model now supports projectile nuclei and depending on what exactly happens in event generation the run time to run additional initializations is quite significant for the first couple of events. So it just timed out at least for the he-air tests.
> 
> The first test fails because of `PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\hostedtoolcache\\windows\\Python\\3.10.9\\x64\\lib\\site-packages\\impy\\iamdata\\dpm3191_v001.zip`.
> 
> Tests on windows pass on my machine. I’ll simply restart them. The file download errors are unavoidable by design when multiprocessing. So we should consider triggering them at the end of the installation “manually”, or so.

- We should increase the timeout then, to make the tests pass.
- To avoid the occasional race condition on CI, we should short-term add a command which downloads all the dependencies before the actual tests run in parallel. In the long-term, we could try to fix the download code so that it is thread-safe. This is not easy to do in a platform-independent way, since there is no cross-platform solution for file-locks AFAIK.

src/impy/models/sibyll.py
HDembinski: I moved this to an issue, so that we can address it in another PR.

PR #111, opened 2022-12-31
Restartable models

### LINKED ISSUE #106: Run models in separate processes to solve multiple issues
One of the major annoyances when working with impy interactively, is that one cannot create several instances of most models without spawning processes. My plan for impy is to make it part of the API that a model is always started in a separate process. This solves a lot of issues.

Here, creating a model spawns a new process, and the user never notices. It is an implementation detail. Finally, the model acts like a normal Python class.
```py
m1 = EposLHC()  # starts process 1 which initializes an independent instance of EposLHC
m2 = EposLHC()  # starts process 2 which initializes another independent instance
for event in m1(kin1, 10):
   ...
for event in m2(kin2, 10):
   ...
```
`MCRun.__call__` communicates with the process and runs the event generation in the other process.

A caveat of this approach is that there are overheads in process communication which may slow impy down, and we want to avoid that (speed is something we care about). I assume that basic interprocess process communication (sending input data to the other process, waiting for its output) has a negligible impact for most models, but I have to see.  For SIBYLL, which is incredibly fast, it may be noticable.

The main overhead of pickling and unpickling an event object can be avoided by using [shared memory](https://docs.python.org/3.10/library/multiprocessing.shared_memory.html). Shared memory allows two processes to read/write to the same section of memory, so that data can be transferred without any overhead.

Implementing this requires some changes to EventData, which will be allocated in shared memory, and the models, because they need to write to EventData allocated in Python instead of their own Fortran buffers wrapped by MCEvent. Most models call a Fortran routine anyway to copy data from their internal buffers into a HepEvt like buffer. This should be replaced by new routines that copy the data into the EventData object instead. This can probably be implemented with Numba, if the original buffers are readable from Python, requiring no new Fortran code.

Since we run the models in another process, we may even be able to solve the issue that most models produce way too much output, since we should be able to redirect stdout of the other process into some buffers that we control from the main thread.

HDembinski:
An attempt to close #106

This branch is based on #112, which should be merged first.

This is using two approaches. For most models, it is sufficient run some initialization code exactly once, which I ensure with the new `_once` private method. Concrete models should not implemented `__init__` anymore, only `_once` which is called by `MCRun.__init__` exactly once and after the underlying implementation has been initialized.

For Phojet and DPMJet, this does not work (at least I could not get it to work), so I followed the idea expressed in #106. I wrote new base class MCRunRemote, which a brilliant and terrible piece of engineering. Behind the scenes, it runs every call to `cross_section` and `__call__` through a remote process which is created and dies just for the call. The beauty of it: a model which requires this workaround just has to inherit from MCRunRemote instead of MCRun, no other change required!

To maintain the illusion of having a single model and not one that is restarted all the time, I rely strongly on the random-state serialization feature that @jncots added. I further improved this so that the PRNG state of the numpy generator is also saved. Since this so important, I added more tests that check the RNG state serialization and found some issues. @jncots could you please look into this?

#### Other changes
- `MCRun.get_stable` returns set of particles that were changed to stable
- Kinematics is now passed to `MCRun.__call__` instead of `MCRun.__init__`; this makes more sense for models that are restartable. Some models need a maximum energy upfront, this can be passed via init.
- `MCRun.kinematics` property was removed. The kinematics should not be part of the state of the generator, we pass them explicitly to the two functions which need them, `MCRun.__call__` and `MCRun.cross_section`. Less class state is generally better, it simplifies the design and the reasoning about something. It does not matter that the underlying Fortran stores the kinematics, because we abstracted that away in our high-level API.
- `MCRun.random_state` now includes the state of the numpy PRNG
- `MCRun._composite_plan` now generates heavy elements first, as a workaround for DPMJet
- `MCRun.set_unstable` was renamed to `MCRun.maydecay`


HDembinski:
@afedynitch You said that you are looking into fixing cross-sections for the models. I added a new test `test_cross_sections.py` which checks the output of all models for various combinations of projectiles and targets. The references for such checks are very rough, of course, because the models seem to disagree a lot especially on nuclear cross-sections, however, the test still checks basic sanity of the model response and I found several bugs this way, too.

HDembinski:
From my side, this PR is ready for review, I need your help to fix the remaining things.

Tests are failing largely because of the new tests for rng_state persistence that I added (that is unrelated to my changes).

HDembinski:
At least one of the tests is stalling, on my own computer and on CI. I haven't figured out which one yet.

HDembinski:
The test which runs forever is test_to_hepmc3.py, I am investigating.

src/impy/models/dpmjetIII.py
```diff
@@ -95,37 +94,27 @@ def __init__(self, evt_kin, *, seed=None):
...
-    def _cross_section(self, kin=None, precision=None):
-        kin = self.kinematics if kin is None else kin
-        # we override to set precision
-        if (kin.p1.A and kin.p1.A > 1) or kin.p2.A > 1:
-            assert kin.p2.A >= 1, "DPMJET requires nucleons or nuclei on side 2."
+    def _cross_section(self, kin, precision=None):
+        self._set_kinematics(kin)
+        assert kin.p2.A >= 1  # should be guaranteed by MCRun._validate_kinematics
+
+        projectile_id = 2212 if kin.p1.is_nucleus else kin.p1
```
afedynitch, line 107: This is incorrect if p1 is a neutron

src/impy/models/dpmjetIII.py
```diff
@@ -95,37 +94,27 @@ def __init__(self, evt_kin, *, seed=None):
...
                 saved = self._lib.dtglgp.jstatb
                 # Set number of trials for Glauber model integration
                 self._lib.dtglgp.jstatb = precision
             self._lib.dt_xsglau(
                 kin.p1.A or 1,
                 kin.p2.A or 1,
-                self._lib.idt_icihad(2212)
-                if (kin.p1.A and kin.p1.A > 1)
-                else self._lib.idt_icihad(kin.p1),
+                projectile_id,
```
afedynitch, line 117: DT_XSGLAU needs bamjet codes. It is also the right routine to call for hadron-nucleus.

src/impy/models/dpmjetIII.py
```diff
@@ -136,17 +125,26 @@ def _cross_section(self, kin=None, precision=None):
...
                 self._lib.dtglgp.jstatb = saved
             return CrossSectionData(inelastic=self._lib.dtglxs.xspro[0, 0, 0])
-        else:
-            stot, sela = self._lib.dt_xshn(
-                self._lib.idt_icihad(kin.p1), self._lib.idt_icihad(kin.p2), 0.0, kin.ecm
-            )
-            return CrossSectionData(total=stot, elastic=sela, inelastic=stot - sela)
+
+        assert kin.p1.is_hadron
+        assert kin.p2.is_hadron
```
afedynitch, line 130: allowed targets are _projectiles, but in reality it's only `Nuclei()`, proton and neutron.

src/impy/models/dpmjetIII.py
```diff
@@ -136,17 +125,26 @@ def _cross_section(self, kin=None, precision=None):
...
+        return CrossSectionData(total=stot, elastic=sela, inelastic=stot - sela)
 
         # TODO set more cross-sections
 
     def _set_kinematics(self, kin):
         # Save maximal mass that has been initialized
         # (DPMJET sometimes crashes if higher mass requested than initialized)
+
+        # Relax momentum and energy conservation checks at very high energies
+        if kin.ecm > 5e4:
```
afedynitch, line 142: Can't do that without setting it back. Like this the output will depend on the sequence of requested energies since energy-momentum conservation checks can reject. Either this has to stay in init or, one needs to save the default values and reset them if kinematics change.

src/impy/models/dpmjetIII.py
```diff
@@ -160,9 +158,9 @@ def _set_kinematics(self, kin):
             self._lib.dt_init(
                 -1,
                 max(kin.plab, 100.0),
```
afedynitch, line 160: Also the maximal initialization energy cannot be exceeded. So, it has to be saved as well.

src/impy/models/epos.py
```diff
@@ -45,29 +46,26 @@ class EposLHC(MCRun):
         + "/releases/download/zipped_data_v1.0/epos_v001.zip"
     )
 
-    def __init__(self, evt_kin, *, seed=None):
-        import impy
-
-        super().__init__(seed)
+    def _once(self, ecm_max=1000 * TeV):
+        from impy import debug_level
```
afedynitch, line 50: Does this imply that the debug level cannot change in runtime? It is important if you search for something that goes wrong at event 12345 but don't want to see the output of 12344 events.

src/impy/models/phojet.py
```diff
@@ -190,10 +179,19 @@ def _set_kinematics(self, k):
...
+        self._beams = k.beams
+
+        # Initialize kinematics and tables (only once needed)
+        # LIMITATION: Once this has been called for a given set of particles,
+        # it cannot be called again for different values of k.p1, k.p2. This
+        # means we cannot change particles in Phojet.
+        if self._lib.pho_event(-1, *self._beams)[1]:
+            raise RuntimeError(
+                "initialization failed with the current event kinematics"
+            )
```
afedynitch, line 191: Nope, this is wrong. pho_event(-1, ....) has to be called once for the highest energy requested for a run. Different combinations of particles and energies are set with `pho_setpar`. Once an event generation is triggered (`pho_event(>0, ..)`), PHOJET runs the remaining initializations. So `pho_event(-1' only once.

src/impy/models/pythia8.py
```diff
@@ -64,11 +61,12 @@ def __init__(self, evt_kin, *, seed=None):
             del environ["PYTHIA8DATA"]
         self._lib.pythia = self._lib.Pythia(datdir, True)
 
-        # must come last
-        self.kinematics = evt_kin
-        self._set_final_state_particles()
-
-    def _cross_section(self, kin=None):
+    def _cross_section(self, kin):
+        if (kin.p2.A or 1) > 1:
```
afedynitch, line 65: It's really that bad at nuclei :)

src/impy/models/sibyll.py
```diff
@@ -125,6 +96,7 @@ def sigma_inel_air(self):
     def _set_kinematics(self, kin):
         self._production_id = self._lib.isib_pdg2pid(kin.p1)
```
afedynitch, line 97: Should be rather `_projectile_id`?

src/impy/models/sibyll.py
```diff
@@ -142,61 +114,54 @@ def _set_stable(self, pdgid, stable):
...
-
 
-class Sibyll21(SIBYLLRun):
+class Sibyll21(SIBYLLRun, MCRun):
     _version = "2.1"
     _library_name = "_sib21"
 
 
-class Sibyll23(SIBYLLRun):
+# For some reason, Sibyll23 requires MCRunRemote, but the others don't
```
afedynitch, line 129: Does this mean that only Sibyll23 crashes on multiple inits? In SIBYLL, the initialization should not be called multiple times even if it works.

src/impy/models/dpmjetIII.py
HDembinski: I think I just moved that code, but maybe that another place to use the new function `is_real_nucleus` then?

src/impy/models/dpmjetIII.py
HDembinski: Oops, I forgot to call idt_icihad on the projectile.

src/impy/models/dpmjetIII.py
HDembinski: asserts are just an executable comment, I put this here to remind myself that we only deal with hadrons at this point in the function. The nuclear targets are handled by the part above.

src/impy/models/dpmjetIII.py
HDembinski: It cannot be in init. 

Ah, I see what you mean, but it does not really matter, I think. We need to use MCRunRemote which always uses a fresh copy of dpmjet when _set_kinematics is called, so it is always called only once per dpmjet instance.

src/impy/models/dpmjetIII.py
HDembinski: dt_init can only called once, right? At least that was my impression from observing the code. MCRunRemote makes it so that _set_kinematics is called exactly once per instance.

src/impy/models/epos.py
HDembinski: If you pass the debug level to fortran in a function that is called only at init, then you cannot change the debug level at run time. This is how all of our models are currently implemented. We never had an interface to change the debug level after init.

HDembinski: Regarding this:
```
impy.debug_level = 0
m1 = Epos() # do something

impy.debug_level = 5
m2 = Phojet() # do something different
```
I thought this works, this is why `debug_level` is imported freshly from top level `impy` (and to avoid some circular import issue). Did you test it and found it does not work?

src/impy/models/phojet.py
HDembinski: Can we call it with a maximum energy that is never exceeded? like 1000 * TeV?

src/impy/models/sibyll.py
```diff
@@ -67,29 +46,21 @@ class SIBYLLRun(MCRun):
         )
     }
 
-    def __init__(self, evt_kin, *, seed=None):
-        super().__init__(seed)
+    def _once(self):
+        from impy import debug_level
```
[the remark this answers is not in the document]
HDembinski: I agree, but I did not design the library in this way, it was already like that.

src/impy/models/sibyll.py
HDembinski: We could unify the names a bit, yes.

src/impy/models/sibyll.py
HDembinski: Yes, that's exactly what it means, you can try it out. Replace MCRunRemote with MCRun further below as base class for Sibyll23 and run the tests.

src/impy/models/pythia8.py
HDembinski: I don't know, perhaps you have to query the cross-section from another variable. At some point I dereference a nullptr in the calls when nuclei are used. I  didn't check what exactly happens.

src/impy/models/dpmjetIII.py
HDembinski: I tried to fix this a bit, but you are the DPMJet expert, so I am relying on your help.

src/impy/models/dpmjetIII.py
HDembinski: Fixed.

src/impy/models/dpmjetIII.py
HDembinski: I used the new function `is_real_nucleus` here to fix this.

PR #112, opened 2023-01-02
Fix kinematics

HDembinski:
This patch fixes some errors in the formulas of EventKinematics and improves its documentation.

- I replaced most of the explicit formulas with named functions, which improves readability, DRYness, and reduces errors from typos (which were present).
- In the process, I added more tests and realized that our implementation specifies the energy per nucleon when we use a nuclear projectile, not the energy of the nucleus. This is clarified now in the documentation of EventKinematics.
- The mass for a nuclear projectile was wrongly computed, it was the mass of the nucleus instead of the mass of a nucleon. It baffles me that our tests did not catch this somewhere.

I fixed a minor bug in Epos, where the result was always computed in the center-of-mass frame. As a consequence, I needed to replace one of the references.

#### Other minor changes
- CompositeTarget was moved to impy.util, which is a good place for all of our lowest level utility classes and functions, it is still importable from impy.kinematics, too, so nothing changes from the user perspective
- EventFrame was also moved to impy.util, but is also still importable from kinematics, see previous point
- `_normalize_particle` is now `process_particle` in impy.util, since it is used in many places
- Energy units (MeV, TeV, ...) can now be imported also from impy.kinematics in addition to impy.constants, which seems fitting and allows one to save a bit of typing in notebooks
- test_generators.py now only produces images for tests that fail, so you can find them easier. If you want to look at all of them, set DEBUG=1: `DEBUG=1 python -m pytest tests/test_generators.py`


src/impy/kinematics.py
```diff
@@ -185,57 +100,60 @@ def __init__(
...
+        self.p1 = process_particle(particle1)
+        self.p2 = process_particle(particle2)
 
         if isinstance(self.p1, CompositeTarget):
             raise ValueError("Only 2nd particle can be CompositeTarget")
 
         p2_is_composite = isinstance(self.p2, CompositeTarget)
 
-        m1 = mass(self.p1)
+        m1 = nucleon_mass if self.p1.is_nucleus else mass(self.p1)
```
afedynitch, line 111: I didn’t test what it numerically does but given that is_nucleus returns true for protons and neutrons, it will be incorrect to pick the nucleon mass instead of the correct hadron masses. :-1: to the `is_nucleus = True` for protons in `particle`.
But ok, need some workaround.

src/impy/kinematics.py
```diff
@@ -185,57 +100,60 @@ def __init__(
...
-            self.beams[1][3] = np.sqrt(m2**2 + p2**2)
+            self.beams[0][3] = momentum2energy(p1, m1)
+            self.beams[1][3] = momentum2energy(p2, m2)
             s = np.sum(self.beams, axis=0)
-            self.ecm = np.sqrt(s[3] ** 2 - np.sum(s[:3] ** 2))
-            self.elab = 0.5 * (self.ecm**2 - m1**2 + m2**2) / m2
+            # We compute ecm with energy2momentum. It is not really energy to momentum,
+            # but energy2momentum(x, y) computes x^2 - y^2, which is what we need. Here,
+            # I use that px and py are always zero, if we ever change this, many formulas
+            # have to change in this class, like all the boosts
```
afedynitch, line 136: Not sure I understand this. The formulae seem identical to what was there before, except that before ecm was computed using the sum over px/y/z whereas now the code really assumes px and py to be zero. Or do you mean something else?

We don’t have the boosts in place to allow for arbitrary incidence and @jncots was working on a generic lorentz trafo, which at that time was downvoted because there is a different maintained lib. I still believe that we should allow for arbitrary kinematic specifications since most of the pieces are already in place for that. Also, we would need to do extensive testing (not unit testing) of an external lorentz trafo code, in particular to mitigate the unavoidable round-off errors for UHECR energies. We have some recipes from FLUKA how one can iteratively mitigate the round-off errors, preserving stuff like particles on-mass shell etc. Not that straight forward, though.

afedynitch (approved):
@HDembinski, thanks for doing more nice work. This PR is fairly straight forward. I left a few comments in the files, and below are a few other ones:

> * I replaced most of the explicit formulas with named functions, which improves readability, DRYness, and reduces errors from typos (which were present).
- This is indeed good. But I didn’t see any obvious typo, could you point it out? (Because of that’s relevant it might be in the MCEq tables etc.)

> * In the process, I added more tests and realized that our implementation specifies the energy per nucleon when we use a nuclear projectile, not the energy of the nucleus. This is clarified now in the documentation of EventKinematics.
That’s how most of the accelerators specify the energy and it also makes sense, the boost is the same if you use gamma=energy_per_nucleon/m_nucleon or gamma=energy/m_nucleus.
> * The mass for a nuclear projectile was wrongly computed, it was the mass of the nucleus instead of the mass of a nucleon. It baffles me that our tests did not catch this somewhere.

- This might not have an effect because the mass is never specified or used if the energy is given in the correct variable, ecm or elab for the corresponding model.
> I fixed a minor bug in Epos, where the result was always computed in the center-of-mass frame. As a consequence, I needed to replace one of the references.
- Noticed that. Strange. Is there a test for energy or momentum conservation of the final state particles?

All the other changes are clean improvements. Cheers.

src/impy/kinematics.py
HDembinski: It is slightly incorrect, yes. A condition that does not consider protons and neutrons as nuclei is
```
p.is_nucleus and not p.is_hadron
```
`is_hadron` is true for protons and neutrons, but not for "real" nuclei with A > 1. Or we use 
```
p.A and p.A > 1
```
for "real" nuclei.

src/impy/kinematics.py
HDembinski: This is just to use named formulas everywhere instead of raw computations, and `energy2momentum` uses a neat numerical trick that we can also use here. That's why I changed it. I can restore the np.sum over the momenta, but right now that's just wasting time, because px and py are always zero. We currently pass the beams via the z-components.

I think we should first fix the stuff that is already in the library and not working correctly, before adding any more features. The addition of CompositeTarget already caused a lot of additional issues, time that would have been better spend on making everything working that the library advertises to do.

src/impy/kinematics.py
```diff
@@ -185,57 +100,60 @@ def __init__(
...
 
         self.beams = (np.zeros(4), np.zeros(4))
 
-        # Input specification in center of mass frame
+        # Input specification in center-of-mass frame
         if ecm is not None:
-            self.frame = EventFrame.CENTER_OF_MASS if frame is None else frame
+            self.frame = frame or EventFrame.CENTER_OF_MASS
             self.ecm = ecm
-            self.elab = 0.5 * (ecm**2 - m1**2 - m2**2) / m2
```
HDembinski, line 205: @afedynitch  You asked about typos. See lines 205 and 219
205 - m1 ** 2 - m2 ** 2
219 - m1 ** 2 + m2 ** 2
I believe the second is correct. I recalculated this and put the computation in the comment in impy.util.

HDembinski:
> This might not have an effect because the mass is never specified or used if the energy is given in the correct variable, ecm or elab for the corresponding model.

Yes, we use the mass to compute momentum from energy, but in probably all tests we always use the energy, and like you said, the models also use energy, except perhaps Dpmjet which wants the full 4-vectors.

src/impy/kinematics.py
afedynitch: For me `p.A and p.A > 1` seems more natural and easier to understand.

src/impy/kinematics.py
afedynitch: No no, it's fine as it is, no need to sum over px, py. I just wanted to know if there were "typos" that could have resulted in wrong formulas, which would impact already derived works.

src/impy/kinematics.py
afedynitch: Uhh. ok. Thanks.

src/impy/kinematics.py
HDembinski: Since this distinction between p, n and "real" nuclei with A > 1 comes up a lot, I added the function `is_real_nucleus` to impy.util.

HDembinski:
> Noticed that. Strange. Is there a test for energy or momentum conservation of the final state particles?

No, but I know that some models, notably EPOS, violate this to some extend. It is even pointed out in a paper https://inspirehep.net/literature/1859944

src/impy/kinematics.py
HDembinski: I didn't think it would matter, but this small change in the mass actually made one of our tests trip. :)

HDembinski:
Very strange, I made a small change, and now UrQMD is failing. And I just regenerated the reference.

HDembinski:
The windows test never finishes either.

HDembinski:
The tests fail because a small shift in the pseudorapidities moves a delta proton peak by one bin.

afedynitch (approved):

afedynitch:
@HDembinski. The PR would be ready to merge except this UrQMD issue. You mentioned that you regenerated the expectation but the test is still failing. What is your plan on this?

HDembinski:
@afedynitch I think it is fine to acccept this, since the issue is understood and not of concern. But I wanted to mark the offending test as xfail before merging, to keep the lights green (otherwise it is hard to detect new failures). I did get around to doing this.

afedynitch:
@HDembinski we're now getting the XCode 14 bug on the macos-latest builds. It seems that our workers migrated to macos 12 and the XCode14 tool chain that has the weird linker bug we encountered. 

The 14.2 version that has the bug fixed will [become default on January 16th](https://github.com/actions/runner-images/issues/6746). How unfortunate is that? :) There is a fix, though

HDembinski:
Thanks!

PR #114, opened 2023-01-14
Attempt to make test_generators more robust

HDembinski:
I reduced the binning of the histograms used by test_generators in an attempt to make the test more robust against the issue we observed in #112


afedynitch:
I noticed that this particular test trips if the output is reference histogram is generated on an M1 Mac. For a previous PR I regenerated it on the thread ripper and then it passed on both. I'll try it and if the tests pass merge before the impy->chromo PR since most changes here are for binaries.

HDembinski:
No, don't bother. It is not practical if I cannot generate the reference myself. I think we should rather skip that test.

Besides, I think we should generally skip test_generators on Windows, where it takes much longer than on Macos and Linux.

afedynitch:
Ok, agree.

Regarding windows, it seems to be related to hepmc. These tests take much longer than on *nix.

PR #115, opened 2023-01-15
Replace RMMARD with NumPy RNG

### LINKED ISSUE #102: Use RMMARD to compute multinomial distribution for CompositeTarget
We should use RMMARD to sample the numbers from the multinomial distribution when the user uses a CompositeTarget. We want to be able to save and restore the Model state, and that only works for CompositeTargets if RMMARD is also used to sample from the nuclei.

When you add new features, please consider how it impacts the self-consistency of the project.

**Edit**: In #104 I suggest to replace RMMARD altogether with Numpy's PRNG. That's a better solution, since we get an excellent PRNG with well-known statistical properties, and it also solves this problem here, since we can use the same PRNG to drive Numpy and the generators.

### LINKED ISSUE #104: Replace RMMARD with a modern PRNG
Is there are reason why we want to use RMMARD for all models, apart from "CORSIKA uses it"? RMMARD is a bad generator.

- It has a complex state and the implementation is also way more complex than necessary, which means it is very likely slow. The latest generation of PRNGs are extremely simple (state is an array of 2 to 4 int64), very fast (typically only a handful lines of code) and some can exploit SIMD instructions to make them even faster.
- It is also limited because it returns always double precision numbers. A PRNG should return integers which are then converted into other numbers. Fundamentally, a PRNG generates random bits.
- It is unknown what its statistical properties are. Lots of PRNGs that used to be state-of-the-art have been dethroned, like the Mersenne-Twister.and they are known to pass the hardest statistical tests.
- It only allows to use one integer for seeding, which only allows 2**32 different streams of random numbers. A proper PRNG allows one to seed the full state of the generator, to access the full number of independent streams which are allowed by the generator, typically 2 ** 128 or more in modern PRNGs.

Modern generators considered good are  xoshiro256++ (see https://prng.di.unimi.it which also has a nice comparison of modern PRNGs) or the PCG64 generator implemented in Numpy
https://github.com/numpy/numpy/blob/main/numpy/random/_pcg64.pyx
https://github.com/numpy/numpy/blob/main/numpy/random/src/pcg64/pcg64.c

Using the latter has many advantages.
- We don't need to implement and maintain and test the generator ourselves.
- Loading/saving state in Python is already supported.
- We can use multinomial distribution from numpy to generate nuclei from the CompositeTarget without using a second random number generator.

We just need a Fortran wrapper for the C implementation, which seems easy to write:
https://engineering.purdue.edu/ECN/Support/KB/Docs/CallingCFromFortran

HDembinski:
Closes #102 
Closes #104 

Any Numpy PRNG can be used (not only PCG64), currently it is the default one from Numpy. There should be no need to use any other. Seeds are handled completely in Python by the base class MCRun, Fortran codes should not set seeds. The Fortran functions which do any kind of RPNG initialization have been replaced with noop stubs.

I performed some crude benchmarks to see whether this change impacts the speed of the generators. SIBYLL-2.1 is the fastest, it got maybe 5 % slower. EPOS is one of the slowest models, it got 13 % faster.

Benchmark: Runtime to generate 10000 pp events at 1 TeV and write as hepmc3.

|                       | EPOS-LHC  | Sibyll-2.1 |
|--------------|-------|----------|
| main              | 51 s     | 3.1 s         |
| this branch   | 45 s     | 3.3 s        |

test_rng_state fails for EPOS and UrQMD. EPOS used to pass. We can fix this in a separate PR. I did some investigation but could find out why this is happening.

gasdev and spgasdev (used by the Sibylls) are now implemented in terms of the numpy random normal generator. This allows us to get rid of the extra state which the old implementations held. If we want reproducible randomness, we need to get rid of random functions which cache some results.

**Other changes**
- In addition to extra.cfg, we now also have model.cfg, which contains the models that are normally build. When changing Fortran code, it can be useful to comment out models which are not affected during development here. CMake will then not build those models.
- I fixed a bug in F2PY.cmake, the compile definitions were not passed correctly to numpy.f2py.
- I removed the obsolete Makefiles.
- Renamed src/fortran/EPOS-LHC to src/fortran/epos to be more consistent with other models.
- Fixes a bug in the CLI: seed was never passed to the model!
- Added a null writer for the CLI, which is useful for benchmarking.
- I deactivated test_generators on the CI Windows runner. It is enough to test all models twice on MacOS and Linux, and the Windows tests generally take much longer than the others. Skipping this should make the runtimes more even.
- I skip test_generators with Model=UrQMD and frame=cms2ft and p2=p, since this test is generally flaky.


HDembinski:
The failure on Windows is caused by a single trip in test_generators.py. Perhaps regenerating the reference fixes the issue.

HDembinski:
I understand the issue with EPOS now. EPOS computes some cached numbers when it generates the first event. If you save the state of the RNG before the first event is created, then generate some events and restore the state, the next event is not going to the be same again, since this time the cache was not filled.

F2Py.cmake
```diff
@@ -89,7 +89,7 @@ function (f2py_add_module target_name)
     # Definitions for source files processing
     set(fortran_defs)
     foreach(_def ${F2PY_ADD_MODULE_COMPILE_DEFS})
-      STRING(APPEND fortran_defs "-D${_def} ")
+      list(APPEND fortran_defs "-D${_def}")
```
jncots, line 92: Is "list" expanded in `add_custom_command` below to just string of items separated by spaces? Because usually it is represented as a string separated by semicolon.

CMakeLists.txt
```diff
@@ -132,34 +132,34 @@ set(f2py_dir ${CMAKE_SOURCE_DIR}/src/f2py)
...
-  ${fortran_dir}/EPOS-LHC/sources/epos-random.f
+  ${fortran_dir}/epos/sources/epos-bas-lhc.f
+  ${fortran_dir}/epos/sources/epos-ids-lhc.f
+  ${fortran_dir}/epos/sources/epos_interface.f
+  ${fortran_dir}/epos/sources/epos-random.f
   ${rangen_source}
   ${logging_source}
   INCLUDE_DIRS
-  ${fortran_dir}/EPOS-LHC/sources
+  ${fortran_dir}/epos/sources ${NUMPY_INCLUDE_DIRS}
```
jncots, line 162: I cannot find NUMPY_INCLUDE_DIRS variable. Where is it defined?

cmake_ext.py
```diff
@@ -147,8 +179,9 @@ def build_extension(self, ext):
         target = ext.name.split(".")[-1]
         suffix = sc.get_config_var("EXT_SUFFIX") or sc.get_config_var("SO")
         output_file = Path(extdir) / (target + suffix)
-        if not output_file.exists() or target == "_eposlhc":  # any target is fine
-            cmd = [cmake_exe, "--build", "."] + build_args
+        targets = [f"_{m}" for m in get_models()]
+        if not output_file.exists() or target == targets[0]:  # any target is fine
+            cmd = [cmake_exe, "--build", ".", "-t", *targets] + build_args
```
jncots, line 184: There are 2 places in CMakeLists.txt, where there is condition for skip a target for Windows (IF WIN32...). I think that can be removed, if this chooses correct targets.

setup.cfg
```diff
@@ -28,7 +28,7 @@ packages = find:
 package_dir =
     = src
 include_package_data = True
-python_requires = >=3.7
+python_requires = >=3.8
```
jncots, line 31: So, we don't compile wheels for python 3.7 anymore?

src/chromo/common.py
```diff
@@ -602,18 +531,10 @@ def __init__(self, seed):
...
-            raise ValueError(f"Invalid seed {seed}")
-
-        if hasattr(self._lib, "init_rmmard"):
-            self._lib.init_rmmard(self._seed)
-
-        # TODO use rmmard for this, too, instead of numpy PRNG
-        self._composite_target_rng = np.random.default_rng(self._seed)
+        self._seed = seed
+        self._rng = np.random.default_rng(seed)
+        if hasattr(self._lib, "npy"):
```
jncots, line 536: Is there any model for which seed should be set in different place, i.e. it doesn't have "npy" common block?

src/fortran/rangen.fpp
```diff
@@ -1,380 +1,164 @@
...
-        NTOT(JSEQ)  = NTOT(JSEQ) - MODCNS
-      ENDIF
+      SUBROUTINE RMMARD( RVEC,LENV,ISEQ )
+      implicit none
+      DOUBLE PRECISION RVEC(*), RVAL
+      INTEGER ISEQ,LENV,IVEC
+
+      DO IVEC = 1, LENV
+        call NPYRNG(RVAL)
+        RVEC(IVEC) = RVAL
```
jncots, line 47: Are the numbers in the sequencies still randomly distirbuted in the way like this? Because before the sequencies were completly independent, and now are filled from a single sequence of random numbers. Probably it is not a problem, because it seems to me that RMMARD function isn't  used directly anywhere.

src/fortran/urqmd-3.4/chromo_urqmd.f
```diff
@@ -62,11 +62,9 @@ SUBROUTINE CHEPEVT
 
 c UrQMD doesn't know K0S or L and produces K0(bar) by default.
 c We will replace those with K0S/L based on a 50/50 rule.
-c Random numbers are sourced from second sequence
-c to avoid interference with the event generator's sequence
 
          IF (ABS(IDHEP(I)).EQ.311) THEN
-            IDHEP(I) = K0SEL(INT(2.D0*SIMRND2()))
+            IDHEP(I) = K0SEL(INT(2.D0*SIMRND()))
```
jncots, line 67: What about the interference with the event generator's sequence?

src/fortran/rangen.fpp
```diff
@@ -1,380 +1,164 @@
...
-        IF ( J97(JSEQ) .EQ. 0 ) J97(JSEQ) = 97
-        C(JSEQ)    = C(JSEQ) - CD
-        IF ( C(JSEQ) .LT. 0.D0 ) C(JSEQ)  = C(JSEQ) + CM
-        UNI        = UNI - C(JSEQ)
-        IF ( UNI .LT. 0.D0 ) UNI = UNI + 1.D0
-C  AN EXACT ZERO HERE IS VERY UNLIKELY, BUT LET''S BE SAFE.
-        IF ( UNI .EQ. 0.D0 ) UNI = TWOM48
-        RVEC(IVEC) = UNI
-      ENDDO
+      spgasdev = real(rval)
```
jncots, line 36: Maybe we should add the check similar to 
```555   S_RNDM = real(SIMRND())
      IF ((S_RNDM.LE.0E0).OR.(S_RNDM.GE.1E0)) GOTO 555
```
below, since casting to real can give the number slightly out of range [0,1].

src/fortran/rangen.fpp
```diff
@@ -1,380 +1,164 @@
...
+      integer dummy
+      integer*8 bitgen
+      COMMON /npy/bitgen
+      double precision rval
 
-      DOUBLE PRECISION RVEC(*)
-      INTEGER          ISEQ,IVEC,LENV
-      SAVE
-C-----------------------------------------------------------------------
+      call NPYGAS(rval, bitgen)
```
jncots, line 34: I noticed that the code in the file is mixture of upper and lower cases. Probably, it is better to convert it to the upper case.

F2Py.cmake
HDembinski: Options for commands need to be passed as lists. There won't be semicolons. The string append was wrong, it generated this `-DIMPY\ -DFOO` instead of the desired `-DIMPY -DFOO`. I am surprised that no one noticed this error before.

src/fortran/rangen.fpp
HDembinski: Fortran ignores case, so let's leave it as it is. Or convert everything to lowercase.

HDembinski: Btw, I actually tried to format a little bit so that variables are lowercase and functions / subroutines are upper case. To make the code more readable than having one default case.

jncots: I also prefer lower case. It is just the most of the code is written in upper case.

CMakeLists.txt
HDembinski: In line 50, where it was before?

cmake_ext.py
HDembinski: I am not sure what you mean, but the target skipping for windows is now implemented in get_models() which defines the targets consistently everywhere.

setup.cfg
HDembinski: 3.7 is not supported by numpy anymore, so we can drop 3.7, too. importlib.metadata is available in 3.8+, which I use to query the version.

HDembinski: https://endoflife.date/python
EOL of 3.7 is officially in 5 months.
Numpy is more strict and dropped support for 3.7 already a while ago. I cannot find the schedule for numpy releases right now.

src/chromo/common.py
HDembinski: Pythia8 is the only one. For Pythia8 we have to do it differently, everyone else uses the code in rangen.fpp which uses the npy common block.

HDembinski: Pythia8 is currently not using the numpy PRNG. I have to see how this can be achieved for Pythia8, that's for another PR.

src/fortran/urqmd-3.4/chromo_urqmd.f
HDembinski: What kind of interference do you worry about? I can see no reason to use a different stream here.

src/fortran/rangen.fpp
HDembinski: No, that's not necessary.

cmake_ext.py
jncots: I meant to remove `if` conditions at line 286 and 361 of CMakeLists.txt.

setup.cfg
jncots: Good, thanks for the information

cmake_ext.py
HDembinski: Ah, now I understand, yes, you are right.

src/fortran/urqmd-3.4/chromo_urqmd.f
jncots: This and previous question boils down to the question: if we take random numbers from one sequence and make from them 2 sequencies (like 1st number goes to 1st sequence and the next one goes to 2nd, and 3rd number goes to 1st sequence again and so on), are two sequencies still good random sequencies (have the same properties as the original)? Most probably yes, but I am not sure 100%.

jncots (approved):
@HDembinski, thanks for making work with rng more simple and clear. It was also a good idea to control model compilation via text file `model.cfg`. Very convenient.

src/fortran/urqmd-3.4/chromo_urqmd.f
HDembinski: A mixed stream of pseudo-random numbers from two PRNGs with different seeds is in theory more "random" than a single stream. A true random number sequences has no autocorrelation whatsoever. Pseudo-random number sequences have autocorrelations, but they so designed so that these auto-correlations are negligible. If  a mixed stream as described in the first sentences has less autocorrelation, so it is closer to the a true randon number sequence. We don't have to worry about this, our PRNG is so good that we do not need to mix numbers from several streams.

In this particular line that we are discussing here, the intend (as explained in the comment) was to draw a random number from a different stream of pseudo-random numbers than the one used by URQMD, so that there is no interference. But there cannot be any interference. URQMD cannot expect the PRNG to be in any particular state, it wouldn't make sense. As a programmer point of view, you need to treat the pseudo-random numbers like a stream of true random numbers. In reality, PRNG give you some added advantages over true randomness, like the ability to make reproducible sequences of random numbers or to rewind the PRNG to an earlier state. There is no interface in URQMD to rewind the PRNG, though, so we don't have to worry about interference.

src/fortran/rangen.fpp
afedynitch: The F77 standard is upper case. F90 is flexible. In DPMJet I standardized the case using a formatter (PlusFORT) a long while ago. Here, I'd say it's not important.

src/fortran/urqmd-3.4/chromo_urqmd.f
afedynitch: To leave one more comment on this multiple sequences topic... The reason for choosing multiple sequences is simple. Often, the generators don't run in a vacuum and are linked to some other user code. If the user code needs random numbers (like here), it is better to use a separate sequence. If a crash occurs in event 1023 then one can pass the code to the developers of the generator and ask to have a look at event 1023. These days, we probably won't work like that and rather ask for the user code (chromo in this case). So as long there is a way to reproduce certain events or problems somehow, it is fine.

setup.py
```diff
@@ -17,53 +16,18 @@
...
-# extra.cfg is not tracked by git, so can be freely modified
-# extra.cfg example:
+# for convenience, support building extra models via EXTRA
+# EXTRA is not tracked by git, so can be freely modified
+# EXTRA example:
 # -----
 # sib23c00
 # sib23c02
 # sib23c03
 # dev_dpmjetIII193=/full/path/to/dir/dpmjetIII-19.3
```
afedynitch, line 26: Is the extra.cfg functionality gone?

HDembinski: No, the comment just got messed up during development. Will be fixed.

HDembinski: done

afedynitch (approved):
Thanks again for the very nice work @HDembinski!

There is one issue remaining, namely the model.cfg broke parallel builds, at least on Windows. I guess this is by design. If you have many cores and no mold, building 2 models takes the same time as building the entire package. Could that be made optional?

Otherwise, the PR can be merged.

HDembinski:
> There is one issue remaining, namely the model.cfg broke parallel builds, at least on Windows. I guess this is by design. If you have many cores and no mold, building 2 models takes the same time as building the entire package. Could that be made optional?

I don't know how to fix this, it should not make a difference for cmake whether we ask it to build the entire directory or only specific targets. Can you give ninja on Windows a try, perhaps it works better.

I think the benefit of the new system outweighs the cons. Support for Windows is anyway broken and putting a lot of effort into this seems like wasted time, since we I don't think we have actual users on Windows.

HDembinski:
@afedynitch The windows runner needs 17 min to compile, the linux runner needs 16 min. It does not look like building on Windows is slower?

afedynitch:
> @afedynitch The windows runner needs 17 min to compile, the linux runner needs 16 min. It does not look like building on Windows is slower?

It doesn't make a big difference because the CI builds on 2 threads not 64. I'll try Ninja, hope it works with Mingw.

src/fortran/urqmd-3.4/chromo_urqmd.f
jncots: Thanks for the enlightening comments. If I had anything in mind when explaining my concerns, it was autocorrelation. Now I realized that the interference here is just a way to separate 2 streams of random numbers.

afedynitch:
Ninja solves the problem. Builds in 82 seconds from scratch.

PR #116, opened 2023-01-16
Refactoring the project name to chromo

afedynitch:
This PR changes everything from impy to chromo.

Additional minor changes:
* The 19.x versions of DPMJET and PHOJET have been modified such that on `chromo.debug_level = 0` they only print the badge and stay silent otherwise. For the old DPMJET and PHOJET versions such a change is not easily possible since all write statements are unprotected and print directly to an output unit. Sending output into the void is tmk impossible without significant workarounds.


CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.17)
 
-project(impylib LANGUAGES Fortran)
+project(chromolib LANGUAGES Fortran)
```
HDembinski, line 3: This should also be called chromo, I think.

examples/compare_models.ipynb
```diff
@@ -24,14 +24,14 @@
...
-    "import impy.models as im"
+    "from chromo.constants import GeV\n",
+    "from chromo.kinematics import CenterOfMass\n",
+    "import chromo.models as im"
    ]
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": null,
```
HDembinski, line 34: I think it is good to ship pre-evaluated notebooks, so that people can see the outputs.

HDembinski:
@afedynitch Thanks for this, that was tedious job, I imagine.

The folder `f2py copy` was committed by accident, I presume, and can be removed?

HDembinski (approved):

afedynitch:
> I approved the change, but the folder `f2py copy` was committed by accident, I presume, and can be removed?

Yes, totally. I removed it. Will merge after the tests pass. Hope it is ok if this one goes before the other PRs since it's kind of intrusive..?

examples/compare_models.ipynb
afedynitch: OK, done

HDembinski:
> Hope it is ok if this one goes before the other PRs since it's kind of intrusive..?

It is not an ideal moment, we could have planned it a bit better, but we have to make the change eventually, so I think you should go ahead. The other PRs can be updated accordingly.

CMakeLists.txt
afedynitch: Sure, why not.

afedynitch:
> It is not an ideal moment,

Agree but to get the package on pypi and the wheels working, we need to do this change as soon as possible.

HDembinski:
> Agree but to get the package on pypi and the wheels working, we need to do this change as soon as possible.

Agreed. I think we should keep the zero in front of the version number a while longer, to indicate that the API is not yet stable, but I also would like to have chromo binary wheels on pypi as soon as possible.

PR #119, opened 2023-02-01
Fixing event copy

HDembinski:
This fixes issues I saw when developing something else with EventData.copy.
- EventData.copy did not copy the kinematics object, it was shared
- EventKinematics object got changed into a tuple by the copy
- I added some tests of copying and pickling of EventData and MCEvent, to ensure that the implementation works

I discovered that dataclasses.astuple works recursively if the dataclass also holds other dataclasses. This caused the second issue.

EventKinematics is now a thin class with just an init over EventKinematicsBase. This was necessary to make copies. I need to be able to call the raw init of the data holder (EventKinematicsBase), but we also want to customize the init. Since Python does not have overloads, I use this solution.

Other changes:
- EventData._fast_selection was replaced with EventData._select, to simplify the code
- I added a test of `EventData.__getitem__` which calls EventData._select

### DIFF
--- a/tests/test_common.py
+++ b/tests/test_common.py
@@ -1,6 +1,39 @@
-from chromo.common import CrossSectionData
+from chromo.common import CrossSectionData, EventData, MCEvent
+from chromo.kinematics import CenterOfMass
 import numpy as np
 import dataclasses
+import pickle
+from types import SimpleNamespace
+import pytest
+from numpy.testing import assert_equal
+
+
+@pytest.fixture
+def evt():
+    i = np.array([1, 2, 3])
+    f = np.array([1.1, 2.2, 3.3])
+    p = np.array([[1, -1], [2, -1], [2, 3]])
+    return EventData(
+        "generator",
+        CenterOfMass(10, "p", "p"),
+        1,
+        0.5,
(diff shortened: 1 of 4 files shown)


tests/test_common.py
```diff
@@ -26,3 +59,60 @@ def test_cross_section_eq():
...
+
+    s = pickle.dumps(evt)
+    evt3 = pickle.loads(s)
+
+    assert evt3 == evt
+
+
+class DummyEvent(MCEvent):
+    def __init__(self):
+        hepevt = SimpleNamespace(
```
afedynitch, line 76: I like the concept of SimpleNamespace. Is it something used only for tests or can I use it for other codes? It's an official language element...

afedynitch (approved):
Ok understood the problem and why you're proposing these changes. I fixed to minor bugs, once tests pass the PR is ready to be merged.

tests/test_common.py
HDembinski: Yes, you can use this everywhere. I use it here for a poor-mans mocking, although the Python standard lib also contains a proper mocking library with unittest.mock.

PR #120, opened 2023-02-01
Simpler models config

HDembinski:
Minor change.

Slightly simpler handling of model.cfg and extra.cfg. Some remove of code in CMakeLists.txt which was duplicated by previous merges/rebases.


CMakeLists.txt
```diff
@@ -17,9 +17,6 @@ enable_language(C CXX)
 
 set(BUILD_dev_dpmjetIII193 CACHE
   PATH "source directory for dev_dpmjetIII193")
-option(BUILD_sib23c00 OFF)
-option(BUILD_sib23c02 OFF)
-option(BUILD_sib23c03 OFF)
```
jncots, line 20: sib23c03, ... will not be build at all. Or `if` conditions using BUILD_... variables  should be removed

CMakeLists.txt
```diff
@@ -465,25 +471,14 @@ list(FILTER dpmjetIII191_sources EXCLUDE REGEX PYR\.f)
 # include those. As a workaround, we generate modified source files and
 # renamed copies of the headers. Not elegant, but works. 🤷‍♂️
 
-
-### dpmjetIII191
 file(GLOB dpmjetIII191_sources
   ${fortran_dir}/dpmjetIII-19.1/src/phojet/*.f
   ${fortran_dir}/dpmjetIII-19.1/src/pythia/*.f
   ${fortran_dir}/dpmjetIII-19.1/src/dpmjet/*.f
 )
```
jncots, line 479: This should be also removed

cmake_ext.py
```diff
@@ -101,20 +101,10 @@ def build_extension(self, ext):
...
-                    if "=" in model:
-                        if model[-1] == "/":
-                            model = model[:-1]
-                        cmake_args.append(f"-DBUILD_{model}")
-                    else:
-                        cmake_args.append(f"-DBUILD_{model}=ON")
+        models = get_models()
+        for model, path in models.items():
+            if path is not None:
+                cmake_args.append(f"-DBUILD_{model}={Path(path)}")
```
jncots, line 108: With this code it is not possible to build sib23c00, sib23c02, sib23c03 by 2 reasons:
* "-DBUILD_sib23c00", ... are not defined
* option(BUILD_sib23c00 OFF) is removed from CMakeLists.txt

jncots (commented):
`extra.cfg` was introduced for development purposes. `model.cfg` was introduced, I guess, by the same reason, to easily remove/add models during debugging. Although it is convenient, there are 2 problems, which I met with `model.cfg`:

* makefiles doesn't work in parallel mode. I made some research, and, indeed, it is not possible to specify targets and build them parallel with makefiles. It would not be a problem by using Ninja, but for some reason, Ninja is slow at linking stage for Linux and MacOS. It takes about 5 - 7 minutes to build all models. It is annoying.
* model.cfg could be missed/lost or forgotten to be restored after debugging.

In terms of usability, it would be good to have old behaviour by default (all models are built in parallel in makefiles), and if one needs to build specific models use `model.cfg` functionality.

CMakeLists.txt
HDembinski: I went back and forth a lot with this PR and eventually forgot to remove the if (BUILD_sib23c00) etc. I now removed them. We don't need these ifs anymore, becaues the source code is available and the targets are just not build.

CMakeLists.txt
HDembinski: Yes, you are right. Fixed.

cmake_ext.py
HDembinski: Yes, that was intentional, but I forgot to remove the corresponding ifs.

HDembinski:
@jncots Thank you for the review and catching my mistakes.

> extra.cfg was introduced for development purposes. model.cfg was introduced, I guess, by the same reason, to easily remove/add models during debugging. Although it is convenient, there are 2 problems, which I met with model.cfg:

Yes, when one makes larger changes to the Fortran layer, it is useful to not build everything during development, but only one model at a time, when you know that the other models will be broken anyway. Some changes, like those to the RNG always trigger a rebuild of all models, since the source file is used by all models.

>    makefiles doesn't work in parallel mode. I made some research, and, indeed, it is not possible to specify targets and build them parallel with makefiles. It would not be a problem by using Ninja, but for some reason, Ninja is slow at linking stage for Linux and MacOS. It takes about 5 - 7 minutes to build all models. It is annoying.

I noticed the same. I don't understand why Ninja is slow at linking, because Ninja does not do any linking by itself. We should check what kind of command is actually run when Ninja does the linking step with VERBOSE=1. Linking usually is not so slow, but if LTO is enabled, it can be very slow. My guess is that Ninja is using LTO by default while make is not. This would be a bug in cmake then, because changing the build system should not change the options for the linker.

Anyway, if LTO is the reason, it can be turned off explicitly in CMake. LTO is quite useful for projects like ours, because it can code faster by inline functions that were compiled in separate translation units and reduce binary size by removing code that cannot be called anyway (because symbols are not exported by the library). It is good to have it enabled for release builds, but it should be disabled for debug builds to increase compilation speed.

> model.cfg could be missed/lost or forgotten to be restored after debugging.

Changes to model.cfg will show up during PR review, so these errors are easy to catch. But maybe there is a better way that I haven't found. I tried to make it so that changes to model.cfg are ignored. You can ignore a file that is checked in, but changes to that file are then nevertheless tracked by git.

One potential solution would be to remove model.cfg from git and ignore it, and create the file dynamically the first time when setup.py is run. Then a developer can change model.cfg, but the changes would not end up upstream. This strategy would also allow us to merge model.cfg and extra.cfg, the latter wouldn't be needed anymore.

> In terms of usability, it would be good to have old behaviour by default (all models are built in parallel in makefiles), and if one needs to build specific models use model.cfg functionality.

I think that we want to use ninja anyway eventually, so we should fix the ninja build and keep the current system. It is simpler than the workarounds we used before to make things work. Workarounds are bad for maintenance, they tend to pile up and eventually the code becomes unreadable/unmaintainable by an outsider.

cmake_ext.py
HDembinski: We only need the BUILD_X flags for models X which use external source files that are not part of the repo.

jncots:
> One potential solution would be to remove model.cfg from git and ignore it, and create the file dynamically the first time when setup.py is run.

If the problem with Ninja will be fixed, then Ninja is fine. I just think that `model.cfg` should be used only for development purposes. For example, if no `model.cfg` is present, then the default set of models are built. If `model.cfg` exists, then only models which are given in the file are built. 

For convenience we can keep the total list of models/extra models + example of for development models in some template `~model.cfg`, which is ignored by the setup.py. If one needs to build specific models he just rename it to `model.cfg` and edit it (for example uncomment required models). `~model.cfg` can be recreated by setup.py if it doesn't exists at build time.

By the way, why it is not `models.cfg` (plural)?

HDembinski:
> If the problem with Ninja will be fixed, then Ninja is fine. I just think that `model.cfg` should be used only for development purposes. For example, if no `model.cfg` is present, then the default set of models are built. If `model.cfg` exists, then only models which are given in the file are built.

That also sounds good. Either way, I don't want to go back to `cmake --build .` without specifying targets, because then we need to hide the targets that should not be build with if-endif in CMakeList.txt and set related cmake variables to trigger these ifs, as we did before. Specifying the targets explicitly that should be build is more elegant.

> By the way, why it is not models.cfg (plural)?

No reason, models.cfg is also fine with me.

> For convenience we can keep the total list of models/extra models + example of for development models in some template ~model.cfg, which is ignored by the setup.py. If one needs to build specific models he just rename it to model.cfg and edit it (for example uncomment required models). ~model.cfg can be recreated at build time if it doesn't exists.

I am not a fan of files starting with `~`. I would rather have a file `default_models.cfg` which lists the default, and one can optionally generate `models.cfg` which is not tracked by git and overrides `default_models.cfg`. In other words, if `models.cfg` exists, `default_models.cfg` is ignored.

HDembinski:
I now implement the following scheme, inspired by @jncots comments.

If `models.cfg` exists, only the models listed there are build. `models.cfg` now plays the role of `extra.cfg` and can be used to build less than the default targets. `models.cfg` is not tracked by git.

If `models.cfg` does not exist, the default models are read from the variable `__all__` in `src/chromo/models/__init__.py`. This works even if chromo is not installed, because I parse the Python code in that file without importing the file.

jncots (approved):
Hello, @HDembinski. Thank you for adding default_models.cfg. It is what I had in mind. I checked that the additional models are built. It seems work fine. I think the PR can be merged.

PR #121, opened 2023-02-01
Epos random dummy

HDembinski:
Minor change.

After talking to Tanguy, I agree that it is better to reduce ifdefs in the original fortran codes. So instead of using ifdef CHROMO in epos-random.f, I made a dummy epos-random-dummy.f with the blank initialization routines and implemented the functions rangen and drangen that produce random numbers in rangen.fpp.


afedynitch (approved):
I thought Tanguy likes the preprocessor.  Anyhow, ok to merge.

PR #122, opened 2023-02-03
Allow Pythia8 to be user-configured

HDembinski:
This allows users to customize the physics processes that Pythia8 runs or change tuning parameters with a new `config` keyword argument to the init. If this argument is not set, Pythia8 uses the standard `SoftQCD:inelastic = on`. If it is set, it can be either a string (representing the content of a standard Pythia input card) or a list of string, the lines of such a file.

With this, users can enable cross-sections for charm production or turn on optional things like rope hadronization or color reconnection.


afedynitch (approved):
Nice, very useful.

PR #123, opened 2023-02-04
chromo logo

### LINKED ISSUE #2: Logo brainstorming
Would be nice to have a logo for impy.

I designed a few logos in the past:

https://github.com/crdb-project
https://github.com/scikit-hep/iminuit
https://github.com/resample-project/resample
https://github.com/boostorg/histogram

In hindsight, I like particularly the CRDB logo. There a lot of rules when designing logos that I researched and found out in discussions with other people. Logos should be minimalist, only a small number colors, one should be able to scale them down to tiny size without loss of essential information. They should look good on white and on black background. Ideally, the logo should include the brand name or at least the initials.

So for impy, a stylized version of the IMPY letters would be nice. Let's use this thread for brainstorming.

HDembinski:
Closes #2 

Here is a logo in metallic font. I made this myself with inkscape. The metal effect uses three colors blue, green, and red, which represent the three color charges of QCD. The font I used is Nova Mono, which is released under a free software licence that allows this kind of use.


afedynitch (approved):
It's a very cool logo!! Nice idea and style. Totally approve :)

PR #124, opened 2023-02-05
Simpler and more efficient Pythia8 event handling

HDembinski:
The Pythia8 event is now a view into the original Pythia8 event structure. Copies are avoided and the new design allows us to use several independent Pythia8 instances in parallel, since there are no global shared data structures. In summary, the event access is faster and the code is more straight-forward, too.

Since direct views require access to private members, I once again use cpp-member-access, which uses a loophole in the C++ standard to do just that. It is safe and works on all platforms, I already use this in iminuit and pyhepmc.

This PR builds on #120 which should be merged first.


afedynitch (approved):
I have not been following the cpp world that much lately but this hack to access private members is really useful here. 

Very good that you forked and simplified it. I would prefer if the submodule points to a tag instead of the main since this is an auxiliary library and we don't need to track every update except for serious bugfixes.

HDembinski:
> I have not been following the cpp world that much lately but this hack to access private members is really useful here.

The good thing: it is a legal hack, supported and mandated by the C++ Standard. So we don't need to worry that it will break in the future. The technique is at least 13 years old, btw.

> Very good that you forked and simplified it. I would prefer if the submodule points to a tag instead of the main since this is an auxiliary library and we don't need to track every update except for serious bugfixes.

The original repo has no releases, only main. We could also simply include the header and be done with it.

afedynitch:
> The original repo has no releases, only main. We could also simply include the header and be done with it, then it is guaranteed to be stable.

If that's a viable option...? Then that would be a good solution.

HDembinski:
> If that's a viable option...? Then that would be a good solution.

If we retain the license for this code it is ok. The original project uses MIT license. Redistribution and modification is allowed if the copyright notice and license is kept in place.

afedynitch:
I'm fine with both solutions and just made my initial comment because I had no idea how stable/mature the code is. We already have several submodules with their licenses. So submodule is ok as well, since it already works.

PR #126, opened 2023-02-09
avoid linking to Python

HDembinski:
This seems to avoid the linking to Python, which is causing issues in #82

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,8 +30,7 @@ include(CMakePrintHelpers)
 include(F2Py)
 find_package(pybind11 CONFIG)
 
-# cannot use find_package(Python), we need PythonLibsNew from pybind11
-find_package(PythonLibsNew REQUIRED)
+find_package(Python REQUIRED COMPONENTS Interpreter Development)
 
 # in Numpy-1.22+, this becomes easier: import numpy.f2py; print(numpy.f2py.get_include())
 if (NOT F2PY_INCLUDE_DIR)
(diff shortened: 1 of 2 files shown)


HDembinski:
I removed the mold code here after realizing a) the mold linker flags were actually never set, because the if conditions never triggered, b) the linker flag `-ld-path=<path to mold>` does not work when you link with gfortran.

HDembinski:
This should be merged into main and then into #82 to remove the need for `fix_macos_installation` and a couple of other workarounds.

afedynitch:
This looks really elegant, thanks. I'm wondering how did you find this solution?

jncots (approved):
Thank you for the solution!

afedynitch:
> These three software packages have expertly written build systems, commits are reviewed by experts with a lot of knowledge, so the solutions which end up in those packages tend to be very good. Code documentation is also often very good.

I checked scipy and numpy but they all use meson for a while 😛

eli-schwartz:
> I checked scipy and numpy but they all use meson for a while stuck_out_tongue

The meson handling calls out to the meson "python" support module's `py.dependency()` method, which defaults to `embed: false`. This roughly corresponds to e.g. the python pkg-config file's split between python.pc and python-embed.pc (the latter is used for producing executables that load a python interpreter, the former for extension modules).

There's platform-specific handling here, it's not so simple as "don't link to libpython", but that's mostly a difference between Windows (do link) and everywhere else (don't link, starting with python 3.8).

HDembinski:
We use the cmake command Python_add_library now, which is supposed to handle this in a platform independent way.

PR #127, opened 2023-02-11
Stop F2PY caching

### LINKED ISSUE #117: Stop f2py caching
The bug in numpy.f2py will soon be fixed https://github.com/numpy/numpy/pull/22657. Once that PR is through, I would like to remove the f2py cache that we have in src/f2py, it is brittle and gets in the way when you work on the Fortran interface.

I recently worked a lot on the Fortran interface. Our build implementation of the f2py caching with files in src/f2py is not working correctly when you change the fortran source files. The f2py files are not correctly updated when the source files change. Generating the f2py files takes negligible amount of time, so I would like to go back to a simpler build system that does not change these files in the source repo.

We originally took this route because of the numpy bug and because we said that the fortran files will not change a lot. But when we work on chromo we also have to change the fortran code. I made several changes to the EPOS interface and the random number generators. So the assumption does not hold in practice that this part is stable. We have regenerated the src/f2py files many times now.

It is also possible that our cached files become invalid when numpy is updated to a newer version. The generated files are not intended to be stored permanently.

HaoZeke:
@HDembinski could you try building `numpy` with https://github.com/numpy/numpy/pull/22657 and running this branch? It should pass. Closes #117.


HDembinski:
Thank you for your help, we appreciate it very much, I am currently traveling, but I will try to check this later today.

HDembinski:
Here is the log from my machine for reference.

[_phojet191.log](https://github.com/impy-project/chromo/files/10720915/_phojet191.log)

HDembinski:
To confirm that this branch still works with f2py from numpy-v1.19.5, I deleted the f2py folder again in my local numpy installation and copied that from numpy-v1.19.5. I then deleted the `build` folder and ran `python setup.py develop`. This builds the module `_phojet191` correctly.

You can test whether the code not only compiles but also runs correctly with this command `python -m pytest tests/test_generators.py -k phojet191 -n0`. You need a couple of extra packages for the tests to run:

```
    pytest
    pytest-benchmark
    pytest-xdist
    pyhepmc>=2.9.0
    graphviz
    uproot
    awkward
    pyyaml
    boost_histogram
    matplotlib
```

HDembinski:
For the record, one can install @HaoZeke branch with `pip install -U -v git+https://github.com/HaoZeke/numpy@fix22648`

HDembinski:
As another test, I now used `pip install -U -v git+https://github.com/HaoZeke/numpy@fix22648` to install the branch with the fix locally, and building `_phojet191` fails. When I just copy crackfortran.py (only this file) from numpy-v1.19.5, it works again.

HaoZeke:
@HDembinski could you enable the workflow for this? I added a dependency on my branch with the fix in https://github.com/numpy/numpy/pull/22657. Locally I'm able to build it, and run the test in the `readme`. If it works on CI we can wait for it to be merged into `main` and pin it.

HDembinski:
Thank you very much, I started the workflow.

HDembinski:
The failure is in the installation step for setuptools, we need to fix this.

HaoZeke:
> The failure is in the installation step for setuptools, we need to fix this before we can test the patch by @HaoZeke

Sorry I had a typo in my last commit, could you try running it now?

.gitignore
```diff
@@ -39,3 +39,4 @@ models.cfg
 pythia*.tgz
 src/chromo/iamdata
 fig/*
+/src/f2py/
```
HDembinski, line 42: I don't think we need to ignore this, the directory can be removed completely, no?

pyproject.toml
```diff
@@ -4,7 +4,7 @@ requires = [
     "wheel",
     "cmake",
     "ninja",
-    "numpy<1.24",
+    "numpy @ git+https://github.com/HaoZeke/numpy.git@fix22648",
```
HDembinski, line 7: Cool, I did not know this syntax.

src/fortran/epos/sources/epos-random.f
```diff
@@ -191,66 +191,6 @@ subroutine ranfini(seed,iseq,iqq)
...
-C  ADAPTATION  : T. PIEROG    IK  FZK KARLSRUHE FROM D. HECK VERSION
-C  DATE     : Feb  17, 2009
-C-----------------------------------------------------------------------
-      IMPLICIT NONE
-      INTEGER          KSEQ
-      PARAMETER        (KSEQ = 2)
-      COMMON /CRRANMA3/CD,CINT,CM,TWOM24,TWOM48,MODCNS
-      DOUBLE PRECISION CD,CINT,CM,TWOM24,TWOM48
-      INTEGER          MODCNS
-      COMMON /CRRANMA4/C,U,IJKL,I97,J97,NTOT,NTOT2,JSEQ
```
HDembinski, line 221: Please don't remove this. The idea is to keep "epos-random.f" untouched, because this is not our code.

HDembinski:
@HaoZeke It is not failing because the directory src/f2py has been removed. Let me fix that in the cmake scripts.

.gitignore
HDembinski: I resolved this in my patch.

HDembinski:
@HaoZeke I fixed the main build issue. The macos-latest test fails, because some dependency (numpy? setuptools?) wants Python-3.9 as the minimum. I would like to keep 3.8 as the minimum for chromo, but is something we don't have to worry about here for now.

The crackfortran code now seems to fail in chromo/src/fortran/urqmd-3.4/sources/make22.f:5976 and following, see
https://github.com/impy-project/chromo/actions/runs/5175988472/jobs/9324260629?pr=127#step:7:6568

HaoZeke:
> @HaoZeke I fixed the main build issue. The macos-latest test fails, because some dependency (numpy? setuptools?) wants Python-3.9 as the minimum. I would like to keep 3.8 as the minimum for chromo, but this is something we don't have to worry about here for now.

Makes sense, yup.


> The crackfortran code now seems to fail in chromo/src/fortran/urqmd-3.4/sources/make22.f:5976 and following, see https://github.com/impy-project/chromo/actions/runs/5175988472/jobs/9324260629?pr=127#step:7:6568

I'll look into this, I can reproduce this error with the `pip` install, but not with `python setup.py develop` oddly enough.

afedynitch:
@HaoZeke, I’ll look into this. It could be rather an issue with this legacy code.

afedynitch:
@HaoZeke, I identified a couple of issues. One of these issues [only occurs on Windows](https://github.com/impy-project/chromo/actions/runs/5187001330/jobs/9348818544#step:8:323) and I could reproduce it on a second machine. One of the variables `mamx` in _eposlhcmodule.c is undefined. On linux this passes. You can save time on compiling the rest by commenting out all the entries in `default_models.cfg` leaving only `eposlhc`.

Some other issues with the test is related to how we were setting character arrays using numpy arrays of `dtype="S1024" or equivalent to pass filesystem paths from python. This doesn't seem to work anymore but worked reliably before.

```from chromo.util import fortran_chars
from chromo.models import _dpmjet306

 _dpmjet306.dtchro.fnevap = fortran_chars(_dpmjet306.dtchro.fnevap, "hello")


------------------------------------------------------
ValueError           Traceback (most recent call last)
Cell In[4], line 1
----> 1 _dpmjet306.dtchro.fnevap = fortran_chars(_dpmjet306.dtchro.fnevap, "hello")

ValueError: too many axes: 1 (effrank=1), expected rank=0```

Directly passing bytestrings seems to works and I asked @jncots to look into the chars issue. It seems to prevent most of our codes from passing the tests. If this is an intentional change to the numpy interface we will adapt but it may not be that - just thay you're aware of this.

**Comment by afedynitch** [comment]
It freezes if it doesn't find the data files. Actually it doesn't freeze, it starts a 48h single core computation of the missing files :) Seems like there is still an issue with passing the correct paths.

> a python string to fortran string like

I wrote this above, too, but it doesn't solve the problem reliably since
```
_dpmjet306.dtchro.fnevap = "hello world"
print(_dpmjet306.dtchro.fnevap) -> hello world
_dpmjet306.dtchro.fnevap = "crazy"
print(_dpmjet306.dtchro.fnevap) -> crazy world
```
Passing bytestrings to character arrays also worked in older numpy versions. For the "fix", to take the first element, I believe it just would pass "h" in the upper example.

You can print the path in cqgsini and return. It must be related to that.

**Comment by jncots** [comment]
I see the problem with just an assignment of python string. The solution with first item that work, actually should look like this:

```python
np.array(
        [char_seq + (len_arr - len_seq) * " "], dtype="S" + str(len_arr)
    )[0]
```

jncots:
The problem with `QGSJet` was in declaration `CHARACTER DATDIR*(256)` of `cqgsini` function. The function received some garbage. Changing the declaration to `CHARACTER*256 DATDIR`  solves the problem.  It seems that something's changed in how `f2py` parses such constructs.

HDembinski:
> The problem with `QGSJet` was in declaration `CHARACTER DATDIR*(256)` of `cqgsini` function. The function received some garbage. Changing the declaration to `CHARACTER*256 DATDIR` solves the problem. It seems that something's changed in how `f2py` parses such constructs.

I am a Fortran beginner, but `CHARACTER*256 DATDIR` makes sense to me, while `CHARACTER DATDIR*(256)` looks really awkward. Is this correct Fortran syntax? I thought it should be `CHARACTER DATDIR(256)`.

When passing the bytes as a numpy array, you should not make the array length a function of the string. Just make it 256 as in Fortran. Fortran does not know how long your numpy array is and you cannot pass the length of the string in this way. Where the string ends is determined by the character `0` (that's a binary zero, not the character "0" in ASCII), just like in C.

In other words, instead of
```py
np.array(
        [char_seq + (len_arr - len_seq) * " "], dtype="S" + str(len_arr)
    )[0]
```
which is wrong, this might work
```py
a = np.zeros(256, dtype=np.uint8)
for i, c in enumerate("my_string"):
    a[i] = ord(c)
```

jncots:
Hello @HaoZeke, we are on the way merge this PR. Do I understand correctly that the patch will available only for numpy>=1.26.0rc1 and will never be backported to earlier versions? I have found only the backport https://github.com/numpy/numpy/pull/24526. If it is so, do you aware of some standard techniques to backport fixes for dependent project like chromo?

HaoZeke:
> Hello @HaoZeke, we are on the way merge this PR. 

Awesome

> Do I understand correctly that the patch will available only for numpy>=1.26.0rc1 and will never be backported to earlier versions? I have found only the backport [numpy/numpy#24526](https://github.com/numpy/numpy/pull/24526). If it is so, do you aware of some standard techniques to backport fixes for dependent project like chromo?

@charris handles the backport mechanism, so I've pinged him, I must admit to not being wholly clear how many versions it typically is applied to. 

Generally I would advise against dependent projects trying to backport fixes, it can quickly get out of hand.

jncots:
> @charris handles the backport mechanism, so I've pinged him, I must admit to not being wholly clear how many versions it typically is applied to.

Thank you.

> Generally I would advise against dependent projects trying to backport fixes, it can quickly get out of hand.

I've got it.

jncots:
> @jncots We only do backports during a release cycle, which is generally ~6 months.

I see. Thanks.

patch_local_numpy_f2py.py
afedynitch, line 1: Not needed

pyproject.toml
```diff
@@ -13,7 +13,7 @@ ignore_missing_imports = true
 
 [tool.pytest.ini_options]
 minversion = "6.0"
-addopts = "-q -ra --ff -n auto"
+addopts = "-q -ra --ff -n auto -p no:warnings"
```
afedynitch, line 16: Warning should stay

setup.cfg
```diff
@@ -30,7 +30,7 @@ package_dir =
 include_package_data = True
 python_requires = >=3.8
 install_requires =
-    numpy
+    numpy @ git+https://github.com/HaoZeke/numpy.git@fix22648
```
afedynitch, line 33: Should be also RC

src/fortran/epos/sources/epos-bas-lhc.f
```diff
@@ -5246,8 +5246,9 @@ subroutine aepos(nin)
 
       include 'epos.inc'
       include 'epos.incems'
-      double precision eppass,etpass
-      common/emnpass/eppass(mamx,4),etpass(mamx,4)
+c**ANFE is unused!
+c      double precision eppass,etpass
+c      common/emnpass/eppass(mamx,4),etpass(mamx,4)
```
afedynitch, line 5251: Maybe unnecessary. Had a problem on windows

afedynitch (approved):
Few things remain to fix. Check also if some the changes in the model Fortran codes can be reverted to default with latest version.

setup.cfg
jncots: I guess there is no requirements for numpy in install section. I checked that chromo built with numpy 1.26.0rc1 works with numpy>1.20.

patch_local_numpy_f2py.py
jncots: Ok

pyproject.toml
jncots: Ok. Reverted

src/fortran/epos/sources/epos-bas-lhc.f
jncots: Reverted

jncots:
Hello @HaoZeke, we have an issue with f2py again. All models fails to produce f2py for a recent release of numpy==1.26.0. Everything worked for numpy==1.26.0rc1. 

The simplest model `sophia` fails at producing .pyf file with traceback:

```
  File "/hetghome/antonpr/miniconda3/envs/env_test/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/hetghome/antonpr/miniconda3/envs/env_test/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/hetghome/antonpr/miniconda3/envs/env_test/lib/python3.9/site-packages/numpy/f2py/__main__.py", line 5, in <module>
    main()
  File "/hetghome/antonpr/miniconda3/envs/env_test/lib/python3.9/site-packages/numpy/f2py/f2py2e.py", line 734, in main
    run_main(sys.argv[1:])
  File "/hetghome/antonpr/miniconda3/envs/env_test/lib/python3.9/site-packages/numpy/f2py/f2py2e.py", line 462, in run_main
    postlist = callcrackfortran(files, options)
  File "/hetghome/antonpr/miniconda3/envs/env_test/lib/python3.9/site-packages/numpy/f2py/f2py2e.py", line 363, in callcrackfortran
    postlist = crackfortran.crackfortran(files)
  File "/hetghome/antonpr/miniconda3/envs/env_test/lib/python3.9/site-packages/numpy/f2py/crackfortran.py", line 3355, in crackfortran
    readfortrancode(files, crackline)
  File "/hetghome/antonpr/miniconda3/envs/env_test/lib/python3.9/site-packages/numpy/f2py/crackfortran.py", line 552, in readfortrancode
    dowithline(finalline)
  File "/hetghome/antonpr/miniconda3/envs/env_test/lib/python3.9/site-packages/numpy/f2py/crackfortran.py", line 858, in crackline
    analyzeline(m, pat[1], line)
  File "/hetghome/antonpr/miniconda3/envs/env_test/lib/python3.9/site-packages/numpy/f2py/crackfortran.py", line 1452, in analyzeline
    vtype = vars[v].get('typespec')
KeyError: 'pi'
Reading fortran codes...
	Reading file 'CMakeFiles/_sophia.dir/SOPHIA20.f' (format:fix,strict)
```

Other models also have the similar problems related to `vtype = vars[v].get('typespec')` and giving `KeyError: 'name_of_data'` where `name_of_data` in Fortran code is kind of:

```fortran
DATA name_of_data /some_value/
``` 
For sophia it is :

```fortran
DATA pi /3.141593D0/
```

Some models (`dpmjetIII191`) output the following traceback (only last 2 lines):

```bash
  File "/hetghome/antonpr/miniconda3/envs/env_test/lib/python3.9/site-packages/numpy/f2py/crackfortran.py", line 1471, in analyzeline
    vars[v]['='] = matches[idx]
IndexError: list index out of range
```


It seems all the problems are related to recently changed code in https://github.com/numpy/numpy/commit/0df1365f8e4d1ee7502312956b59ed1bd7675144 , specifically in the file `[numpy/f2py/crackfortran.py](https://github.com/numpy/numpy/commit/0df1365f8e4d1ee7502312956b59ed1bd7675144#diff-3cebebbb97dffa3ec8e404c3a72222457ad6e043303ae23e36bcff24d473addb)`:

```python
        vtype = vars[v].get('typespec')
          vdim = getdimension(vars[v])

          if (vtype == 'complex'):
              cmplxpat = r"\(.*?\)"
              matches = re.findall(cmplxpat, l[1])
          else:
              matches = l[1].split(',')

          if v not in vars:
              vars[v] = {}
          if '=' in vars[v] and not vars[v]['='] == matches[idx]:
              outmess('analyzeline: changing init expression of "%s" ("%s") to "%s"\n' % (
                  v, vars[v]['='], matches[idx]))

          if vdim is not None:
              # Need to assign multiple values to one variable
              vars[v]['='] = "(/{}/)".format(", ".join(matches))
          else:
              vars[v]['='] = matches[idx]
          last_name = v
```

Should I add a new issue for numpy to fix this?

HaoZeke:
> Hello @HaoZeke, we have an issue with f2py again. All models fails to produce f2py for a recent release of numpy==1.26.0. Everything worked for numpy==1.26.0rc1. 
> 
> The simplest model `sophia` fails at producing .pyf file with traceback:
> 
> ```
> Traceback (most recent call last):
>   File "/hetghome/antonpr/miniconda3/envs/env_test/lib/python3.9/runpy.py", line 197, in _run_module_as_main
>     return _run_code(code, main_globals, None,
>   File "/hetghome/antonpr/miniconda3/envs/env_test/lib/python3.9/runpy.py", line 87, in _run_code
>     exec(code, run_globals)
>   File "/hetghome/antonpr/miniconda3/envs/env_test/lib/python3.9/site-packages/numpy/f2py/__main__.py", line 5, in <module>
>     main()
>   File "/hetghome/antonpr/miniconda3/envs/env_test/lib/python3.9/site-packages/numpy/f2py/f2py2e.py", line 734, in main
>     run_main(sys.argv[1:])
>   File "/hetghome/antonpr/miniconda3/envs/env_test/lib/python3.9/site-packages/numpy/f2py/f2py2e.py", line 462, in run_main
>     postlist = callcrackfortran(files, options)
>   File "/hetghome/antonpr/miniconda3/envs/env_test/lib/python3.9/site-packages/numpy/f2py/f2py2e.py", line 363, in callcrackfortran
>     postlist = crackfortran.crackfortran(files)
>   File "/hetghome/antonpr/miniconda3/envs/env_test/lib/python3.9/site-packages/numpy/f2py/crackfortran.py", line 3355, in crackfortran
>     readfortrancode(files, crackline)
>   File "/hetghome/antonpr/miniconda3/envs/env_test/lib/python3.9/site-packages/numpy/f2py/crackfortran.py", line 552, in readfortrancode
>     dowithline(finalline)
>   File "/hetghome/antonpr/miniconda3/envs/env_test/lib/python3.9/site-packages/numpy/f2py/crackfortran.py", line 858, in crackline
>     analyzeline(m, pat[1], line)
>   File "/hetghome/antonpr/miniconda3/envs/env_test/lib/python3.9/site-packages/numpy/f2py/crackfortran.py", line 1452, in analyzeline
>     vtype = vars[v].get('typespec')
> KeyError: 'pi'
> Reading fortran codes...
> 	Reading file 'CMakeFiles/_sophia.dir/SOPHIA20.f' (format:fix,strict)
> ```
> 
> Other models also have the similar problems related to `vtype = vars[v].get('typespec')` and giving `KeyError: 'name_of_data'` where `name_of_data` in Fortran code is kind of:
> 
> ```fortran
> DATA name_of_data /some_value/
> ``` 
> For sophia it is :
> 
> ```fortran
> DATA pi /3.141593D0/
> ```
> 
> Some models (`dpmjetIII191`) output the following traceback (only last 2 lines):
> 
> ```bash
>   File "/hetghome/antonpr/miniconda3/envs/env_test/lib/python3.9/site-packages/numpy/f2py/crackfortran.py", line 1471, in analyzeline
>     vars[v]['='] = matches[idx]
> IndexError: list index out of range
> ```
> 
> 
> It seems all the problems are related to recently changed code in https://github.com/numpy/numpy/commit/0df1365f8e4d1ee7502312956b59ed1bd7675144 , specifically in the file `[numpy/f2py/crackfortran.py](https://github.com/numpy/numpy/commit/0df1365f8e4d1ee7502312956b59ed1bd7675144#diff-3cebebbb97dffa3ec8e404c3a72222457ad6e043303ae23e36bcff24d473addb)`:
> 
> ```python
>         vtype = vars[v].get('typespec')
>           vdim = getdimension(vars[v])
> 
>           if (vtype == 'complex'):
>               cmplxpat = r"\(.*?\)"
>               matches = re.findall(cmplxpat, l[1])
>           else:
>               matches = l[1].split(',')
> 
>           if v not in vars:
>               vars[v] = {}
>           if '=' in vars[v] and not vars[v]['='] == matches[idx]:
>               outmess('analyzeline: changing init expression of "%s" ("%s") to "%s"\n' % (
>                   v, vars[v]['='], matches[idx]))
> 
>           if vdim is not None:
>               # Need to assign multiple values to one variable
>               vars[v]['='] = "(/{}/)".format(", ".join(matches))
>           else:
>               vars[v]['='] = matches[idx]
>           last_name = v
> ```
> 
> Should I add a new issue for numpy to fix this?

Ouch, sorry, yes please open an issue and I'll get to it ASAP.

jncots:
> Ouch, sorry, yes please open an issue and I'll get to it ASAP.
@HaoZeke, issue is opened https://github.com/numpy/numpy/issues/24746

jncots:
> @jncots We only do backports during a release cycle, which is generally ~6 months.

@charris, in relation to backport, we encountered the following problem. We started to get reports from users about the error `RuntimeError: module compiled against API version 0x10 but this version of numpy is 0xd`. It occurs for `chromo` installed from PyPI using `pip install --pre chromo`. The problem can be fixed by installing `pip install numpy==1.23.0` or higher. I guess, it is because we built the wheels with numpy==1.23.

The comment in https://github.com/numpy/numpy/blob/maintenance/1.26.x/numpy/core/setup_common.py suggests:

> By default, when compiling downstream libraries against NumPy,
> pick an older feature version.

However, to use bugfix in f2py we can build `chromo` only with numpy==1.26.0rc1 (or in the future 1.26.1). C_API_VERSION version for 1.25 and 1.26 is the same. Does it mean that our package can only be used with numpy>=1.25?

We didn't noticed the problem with API, because when we build wheels locally we don't have this problem, i.e. if our package is built locally with numpy==1.26.0 it still can be used with any numpy version > 1.19.5. However, wheels build on CI using `cibuildwheel` produce the "API version" error if numpy version being used is lower than the one used for the build. Maybe you know the reason behind this difference between the local and CI versions?

charris:
> Maybe you know the reason behind this difference

That's odd, sounds like some sort of bug. How do you build locally?

jncots:
> That's odd, sounds like some sort of bug. How do you build locally?

@charris, the initial steps in both CI and local builds are the same:

1. Build *.pyf file
2. Using *.pyf file create *module.c and *f2pywrappers.f
3. Compile fortranobject.c, *module.c with gcc and  *f2pywrappers.f, sources in fortran with gfotran
4. Link them. Here is linking stage of simplest model:

```bash
/hetghome/antonpr/miniconda3/envs/env_test/bin/gfortran -fPIC -O3 -shared  -o /hetghome/antonpr/chromo/build/lib.linux-x86_64-cpython-39/chromo/models/_sib21.cpython-39-x86_64-linux-gnu.so CMakeFiles/_sib21.dir/fortranobject.c.o CMakeFiles/_sib21.dir/_sib21module.c.o "CMakeFiles/_sib21.dir/_sib21-f2pywrappers.f.o" CMakeFiles/_sib21.dir/src/fortran/sibyll/sibyll_21.f.o CMakeFiles/_sib21.dir/src/fortran/sibyll/sib21aux.f.o CMakeFiles/_sib21.dir/src/fortran/sibyll/sibyll_init.fpp.o CMakeFiles/_sib21.dir/src/fortran/logging.f.o CMakeFiles/_sib21.dir/src/fortran/rangen.fpp.o CMakeFiles/_sib21.dir/src/fortran/rangen.c.o CMakeFiles/_sib21.dir/src/fortran/normal.c.o
```

Of course, it is all done with CMake. Another point which maybe relevant is that the compilation occurs with the flag "NPY_NO_DEPRECATED_API=NPY_1_7_API_VERSION".

The difference between CI and local build is that CI uses cibuildwheel which fixes extension files to be dependent on general dynamical libraries, like, for example, on libraries contained in manylinux. 

So do you expect that the wheels build with numpy==1.26 should not work with numpy < 1.25? Or, on opposite, they should?

charris:
> Or, on opposite, they should?

They should. The versioning is based on the hash of the C API and 1,.25 and 1.26 have the same hash.
```
# Version 16 (NumPy 1.23)
# NonNull attributes removed from numpy_api.py
# Version 16 (NumPy 1.24) No change.
0x00000010 = 04a7bf1e65350926a0e528798da263c0

# Version 17 (NumPy 1.25) No actual change.
# Version 17 (NumPy 1.26) No change.
0x00000011 = ca1aebdad799358149567d9d93cbca09
```

jncots:
> The versioning is based on the hash of the C API and 1,.25 and 1.26 have the same hash

@charris, it is clear about 1.25. I am more interested in lower version. Will the package work with numpy==1.20 to numpy=1.24  if it is build with numpy==1.26?

charris:
> Should the package work with numpy==1.20 to numpy=1.24 if it is built with numpy==1.26?

In general, no. The hash changes when new features are added to the API, so any code that makes use of those features will fail on earlier versions. We try to keep forward compatibility, not backward compatibility. If you are making wheels for distribution, you should compile against the earliest version of NumPy that supports what you need.

.github/workflows/release.yml
```diff
@@ -50,15 +50,15 @@ jobs:
...
+    if: ${{ github.event_name == 'workflow_dispatch' || startsWith(github.ref, 'refs/tags/tpypi') }}
     name: ${{ matrix.py }} ${{ matrix.os }} ${{ matrix.arch }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
         arch: [native, arm64]
-        py: [cp38, cp39, cp310, cp311]
+        py: [cp39, cp310, cp311, cp312]
```
afedynitch, line 61: There are no binaries for numpy 1.23 and Python 3.12, and this case is not in the PR tests. So it’s likely to fail. To get this PR merged I’d recommend to avoid building wheels on 3.12 for now, or build those with np1.26

pyproject.toml
```diff
@@ -1,10 +1,11 @@
 [build-system]
 requires = [
+    "packaging",
     "setuptools",
     "wheel",
     "cmake",
-    "ninja",
-    "numpy<1.24",
+    "numpy==1.23.0",
```
afedynitch, line 7: See comment above related to 3.12

afedynitch (approved):
Looks all ready to be merged, except the likely issue with 3.12, which can be pushed to a different PR.

.github/workflows/release.yml
jncots: Ok, I will change it in #164

pyproject.toml
jncots: I see

Issue #129, opened 2023-02-15
No way to set the "precision" parameter for _cross_section in DpmjetIII


jncots:
I would like to use the "precision" parameter to in `_cross_section` for DpmjetIII models. However, now it is not possible to use it via general `cross_section` method of `MCRun`, because it has only one parameter `kinematics`.

Issue #128, opened 2023-02-15
'CrossSectionData' object is not subscriptable


jncots:
There is a problem in `cross_section` function in common.py. When the cross_section is calculated with CompositeTarget an error `'CrossSectionData' object is not subscriptable` appears.

PR #130, opened 2023-02-15
Fix cross_section function

jncots:
These are small fixes to solve the issues #128 and #129

### DIFF
--- a/src/chromo/common.py
+++ b/src/chromo/common.py
@@ -95,47 +95,11 @@ class CrossSectionData:
     def __ne__(self, other):
         return not self == other
 
-    def __iadd__(self, other):
-        if not isinstance(other, CrossSectionData):
-            raise ValueError("+= works only with CrossSectionData class")
+    def __getitem__(self, item):
+        return getattr(self, item)
 
-        fields_to_add = [
-            "total",
-            "inelastic",
-            "elastic",
-            "diffractive_xb",
-            "diffractive_ax",
-            "diffractive_xx",
-            "diffractive_axb",
-        ]
-
-        for field, other_value in dataclasses.asdict(other).items():
-            if field in fields_to_add:
-                self_value = getattr(self, field) + other_value
-                setattr(self, field, self_value)
-        return self
-
-    def __mul__(self, number):
-        if not isinstance(number, (int, float)):
-            raise ValueError("* works only with float")
-
-        fields_to_multiply = [
-            "total",
-            "inelastic",
-            "elastic",
-            "diffractive_xb",
-            "diffractive_ax",
-            "diffractive_xx",
-            "diffractive_axb",
-        ]
-        result = CrossSectionData()
-        for field, value in dataclasses.asdict(self).items():
-            if field in fields_to_multiply:
-                setattr(result, field, value * number)
-        return result
-
-    def __rmul__(self, number):
-        return self.__mul__(number)
+    def __setitem__(self, item, value):
+        setattr(self, item, value)
 
 
 # Do we need EventData.n_spectators in addition to EventData.n_wounded?
@@ -774,7 +738,9 @@ class MCRun(ABC):
                 for component, fraction in zip(kin2.p2.components, kin2.p2.fractions):
                     kin3.p2 = component
                     # this calls cross_section recursively, which is fine
-                    cross_section += fraction * self.cross_section(kin3)
+                    cs = self.cross_section(kin3)
+                    for field in dataclasses.fields(cs):
+                        cross_section[field.name] += fraction * cs[field.name]
                 return cross_section
             else:
                 return self._cross_section(kin)
--- a/src/chromo/common.py
+++ b/src/chromo/common.py
@@ -95,6 +95,48 @@ class CrossSectionData:
     def __ne__(self, other):
         return not self == other
 
+    def __iadd__(self, other):
+        if not isinstance(other, CrossSectionData):
+            raise ValueError("+= works only with CrossSectionData class")
+
+        fields_to_add = [
+            "total",
+            "inelastic",
+            "elastic",
+            "diffractive_xb",
+            "diffractive_ax",
+            "diffractive_xx",
+            "diffractive_axb",
+        ]
+
+        for field, other_value in dataclasses.asdict(other).items():
+            if field in fields_to_add:
+                self_value = getattr(self, field) + other_value
+                setattr(self, field, self_value)
+        return self
+
+    def __mul__(self, number):
+        if not isinstance(number, (int, float)):
+            raise ValueError("* works only with float")
+
+        fields_to_add = [
+            "total",
+            "inelastic",
+            "elastic",
+            "diffractive_xb",
+            "diffractive_ax",
+            "diffractive_xx",
+            "diffractive_axb",
+        ]
+        result = CrossSectionData()
+        for field, value in dataclasses.asdict(self).items():
+            if field in fields_to_add:
+                setattr(result, field, value * number)
+        return result
+
+    def __rmul__(self, number):
+        return self.__mul__(number)
+
 
 # Do we need EventData.n_spectators in addition to EventData.n_wounded?
 # n_spectators can be computed from n_wounded as number_of_nucleons - n_wounded
@@ -732,10 +774,7 @@ class MCRun(ABC):
                 for component, fraction in zip(kin2.p2.components, kin2.p2.fractions):
                     kin3.p2 = component
                     # this calls cross_section recursively, which is fine
-                    cs = self.cross_section(kin3)
-                    for field, value in dataclasses.asdict(cs).items():
-                        val = getattr(cross_section, field) + fraction * value
-                        setattr(cross_section, field, val)
+                    cross_section += fraction * self.cross_section(kin3)
                 return cross_section
             else:
                 return self._cross_section(kin)
--- a/src/chromo/common.py
+++ b/src/chromo/common.py
@@ -715,7 +715,7 @@ class MCRun(ABC):
         """
         self.set_stable(pdgid, False)
 
-    def cross_section(self, kin=None, precision=None):
+    def cross_section(self, kin=None, *, precision=None):
         """Cross sections according to current setup.
 
         Parameters
@@ -738,7 +738,10 @@ class MCRun(ABC):
                         setattr(cross_section, field, val)
                 return cross_section
             else:
-                return self._cross_section(kin, precision)
+                if precision:
+                    return self._cross_section(kin, precision)
+                else:
+                    return self._cross_section(kin)
 
     @abstractmethod
     def _cross_section(self, kin, *args):
--- a/src/chromo/common.py
+++ b/src/chromo/common.py
@@ -715,7 +715,7 @@ class MCRun(ABC):
         """
         self.set_stable(pdgid, False)
 
-    def cross_section(self, kin=None):
+    def cross_section(self, kin=None, precision=None):
         """Cross sections according to current setup.
 
         Parameters
@@ -738,10 +738,10 @@ class MCRun(ABC):
                         setattr(cross_section, field, val)
                 return cross_section
             else:
-                return self._cross_section(kin)
+                return self._cross_section(kin, precision)
 
     @abstractmethod
-    def _cross_section(self, kin):
+    def _cross_section(self, kin, *args):
         pass
 
     @abstractmethod
--- a/src/chromo/common.py
+++ b/src/chromo/common.py
@@ -119,7 +119,7 @@ class CrossSectionData:
         if not isinstance(number, (int, float)):
             raise ValueError("* works only with float")
 
-        fields_to_add = [
+        fields_to_multiply = [
             "total",
             "inelastic",
             "elastic",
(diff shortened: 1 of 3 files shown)


afedynitch (commented):
1. No test. 
2. Pls, overload _iadd_ in CrossSection data to write less implementation-dependent code in MCRun that depends on current content of the CrossSectionData class. 
3. Update the docstring

HDembinski:
I don't like that we add a keyword to the method in the base class which is inherited by every model, even though only one model (with several versions) uses this keyword, while all other models have to ignore it.

A better API is to add an overload for `cross_section` on the derived model class, then the additional keyword is only present there. It is still not great because the derived class has to duplicate the implementation of the base class.

Both of these two options have downsides. So a third option is to add a `set_cross_section_precision` method on the derived model class. That way, you don't need to pass it through the base class. Setting the precision is anyway an expert feature. We should pick a proper default precision that works for most people and users should normally have no need to change it.

jncots:
* The `precision` parameter for `DpmjetIII` cross section are set now via `set_cross_section_precision` method. One can get the currently set `precision` via `get_cross_section_precision`
* `__iadd__` and `__mul__` operators are overloaded for `CrossSectionData` class. It allows to write a clearer code like `cross_section += fraction * self.cross_section(kin3)`. 
* Test for `DpmjetIII` cross section and cross section precision parameter is added.

src/chromo/models/dpmjetIII.py
```diff
@@ -146,6 +148,15 @@ def _cross_section(self, kin=None, precision=None):
 
         # TODO set more cross-sections
 
+    def set_cross_section_precision(self, precision=None):
+        self._cross_section_precision = precision
+
+    def get_cross_section_precision(self):
```
afedynitch, line 154: Why don't you simply use setter and getter for jstatb directly? There is no need keep track of a separate variable. The default value of this variable is also a sensitive choice that @HDembinski asked for.

Also I would call it more technical and representing what it is, like `set_hA_AA_glauber_trials`, so people don't want to touch it. 

Since it is a special option, I would prefer the usage pattern to be:
```
model = DPMJETxxx()
model.set_hA_AA_glauber_trials(100000)
#... do something ... 
```

afedynitch (commented):
The overloading looks fine and the test. For the how the setters and getters are implemented look  at the comment.

src/chromo/models/dpmjetIII.py
jncots: Setter and getter are renamed to  `_hA_AA_glauber_trials`. Now they set `dtglgp.jstatb` parameter directly.

afedynitch (approved):
It is a non-invasive fix to something that users typically don't need to change and only applies to DPMJET, so it's implemented there.

It is not an uncommon function, so doesn't need to be hidden. It is useful if one wants to pre-tabulate tables of cross sections for some other code, like FLUKA or CORSIKA.

From my side, it's good to go. @HDembinski, I hope your previous concerns regarding this PR are resolved.

src/chromo/common.py
```diff
@@ -732,9 +774,7 @@ def cross_section(self, kin=None):
                 for component, fraction in zip(kin2.p2.components, kin2.p2.fractions):
                     kin3.p2 = component
                     # this calls cross_section recursively, which is fine
-                    cs = self.cross_section(kin3)
-                    for i, val in enumerate(dataclasses.astuple(cs)):
-                        cross_section[i] += fraction * val
+                    cross_section += fraction * self.cross_section(kin3)
```
HDembinski, line 777: I suggest to revert the change to CrossSectionData. You are replacing three lines of readable code here are only required in one place with over 30 lines. That is only an improvement if we expect to do multiply and add on CrossSectionData in many places, which I don't see happening.

src/chromo/models/dpmjetIII.py
```diff
@@ -146,6 +140,21 @@ def _cross_section(self, kin=None, precision=None):
 
         # TODO set more cross-sections
 
+    def set_hA_AA_glauber_trials(self, ntrials=1000):
```
HDembinski, line 143: A pure setter should not have a default argument. This should be a property
```
@property
glauber_trials(self):
     """<Documentation goes here, default values of 1000 is only mentioned in the docs>"""
     return self._lib.dtglgp.jstatb

@glauber_trials.setter
def _(self, ntrials):
      <content of set_hA_AA_glauber_trials goes here>
```

afedynitch: Nope. The default number should be set by DPMJET’s initialization and mentioned only in the docs if at all.

src/chromo/common.py
afedynitch: I don’t like how it was written before. It is only technically readable but boilerplate wrt to the physics… dataclasses.astuple is not readable, IMO.

My only concern about the operator implementations is that they have to list the fields by hand, apart from the names for the cross sections, which are aligned with Pythia’s internal naming scheme.

afedynitch: Actually one more concern here, why the operators don’t follow the implementation of what has been done in the other methods?? Iteration over .astuple?

afedynitch:
We need to get this going through, though. It’s really such a tiny thing…

src/chromo/models/dpmjetIII.py
HDembinski: @afedynitch It sounds like we agree, right?

src/chromo/common.py
HDembinski: dataclasses.astuple is the official way to handle a tuple-like dataclass. It is better and more readable to use this than setattr and getattr like in the replacement code.

HDembinski: > It is only technically readable but boilerplate wrt to the physics…

The change replaces this mathematical three-liner which expresses compactly what is happening mathematically here with a generic API that locks CrossSectionData into the role of a value-like object that supports addition and multiplication. Effectively, CrossSectionData now acts like a numpy array.

That would be ok, if that's what we need in many places, and if users should be able to multiplity and add CrossSectionData objects. I don't see that. And if there is no general use case for this functionality, it should **not** define the public API of the class. (I missed a "not" in my original comment in a crucial place...).

Public API you cannot easily take back later. From a maintenance point of view, we have to be stingy about what is exposed publically, only expose what is really needed for sure.

Here, we are not doing this. We expose a public API for functionality that we only use internally and in only one place. It is a bad trade-off.

HDembinski:
> We need to get this going through, though. It’s really such a tiny thing…

I don't understand the rush, but let me note that I reacted promptly when I was asked to review. This change touches a couple of design principles, which I can only point out when I see them violated. Perhaps we need a design document.

Some principles that are touched here:
- Prefer properties over setters/getters, the latter are not Pythonic.
- Do not expose public API (which includes support for operators) for functionality is only used internally (the change to CrossSectionData).
- A collary to the previous: Especially think twice when you implement operators on classes which are not obviously mathematical objects, like vectors, matrices, etc.

src/chromo/common.py
jncots: > dataclasses.astuple is the official way to handle a tuple-like dataclass. It is better and more readable to use this than setattr and getattr like in the replacement code.

I used setattr and getattr because  `'CrossSectionData' object is not subscriptable`. I decided to add `__getitem___` and `__setitem__`  to make it  subscriptable and increase readability. In this case we don't need + and += operators.

The other concern to address when + and += operators were introduced was that in the future CrossSectionData might have fields which should not be added. When such fields will appear `dataclasses.fields(cs)` should be changed to the function which returns only appropriate fields.

src/chromo/models/dpmjetIII.py
jncots: I changed the code to use properties.

jncots:
> Prefer properties over setters/getters, the latter are not Pythonic.

I will have that in mind.

> Do not expose public API (which includes support for operators) for functionality is only used internally (the change to CrossSectionData).

I added `__getitem` and `__setitem__` operators for readability and make CrossSectionData subscriptable.  Otherwise we need ro return to less readable getattr and setattr. 

> A collary to the previous: Especially think twice when you implement operators on classes which are not obviously mathematical objects, like vectors, matrices, etc.

I removed + and += operators.

src/chromo/common.py
```diff
@@ -95,6 +95,12 @@ def __eq__(self, other):
     def __ne__(self, other):
         return not self == other
 
+    def __getitem__(self, item):
+        return getattr(self, item)
+
+    def __setitem__(self, item, value):
+        setattr(self, item, value)
```
HDembinski, line 102: That is still bad, the Zen of Python says: "There should be one (obvious) way to do it". Adding two ways to access the fields of a dataclass `foo.x` and `foo["x"]` is going against that principle. It is also superfluous.

src/chromo/common.py
```diff
@@ -733,8 +739,8 @@ def cross_section(self, kin=None):
                     kin3.p2 = component
                     # this calls cross_section recursively, which is fine
                     cs = self.cross_section(kin3)
-                    for i, val in enumerate(dataclasses.astuple(cs)):
-                        cross_section[i] += fraction * val
+                    for field in dataclasses.fields(cs):
+                        cross_section[field.name] += fraction * cs[field.name]
```
HDembinski, line 743: Sorry, I don't understand what the problem is with dataclasses.astuple. How is this a better solution?

src/chromo/common.py
HDembinski: > The other concern to address when + and += operators were introduced was that in the future CrossSectionData might have fields which should not be added. When such fields will appear dataclasses.fields(cs) should be changed to the function which returns only appropriate fields.

I also think that we may add further fields that should not be added, and that is a reason why I am against adding operators to CrossSectionData at this point.

From the point of future maintainability, I can see a case for moving the code for adding a scaled copy into the class itself. That way, if someone changes the class in the future, they are more likely to notice that this code has to be changed as well.

A solution which does not expose this to users is to add a private method to CrossSectionData like so:
```py
class CrossSectionData:
     ...
     def _mul_add(self, factor, other):
           for field in self.fields:
                 setattr(self, field, factor * getattr(other, field))
```

src/chromo/common.py
HDembinski: I hope we can converge on the solution I describe above, which adds a private method `CrossSectionData._mul_add`. The code here would change to `cross_section._mul_radd(fraction, cs)`

src/chromo/models/dpmjetIII.py
```diff
@@ -146,6 +140,23 @@ def _cross_section(self, kin=None, precision=None):
...
+    @hA_AA_glauber_trials.setter
+    def hA_AA_glauber_trials(self, ntrials):
+        """Set number of trials for Glauber model integration.
+
+        Larger number of `ntrials` reduces the fluctuations in the cross section,
+        thus, making it more smooth. Smaller number of `ntrials` makes calculations of
+        cross section faster.
+
+        Args:
+            ntrials: number of trials. Default is 1000 (set at model initialisation).
```
HDembinski, line 156: There are no args in this case. The comment for the property should go into the getter. Instead of "Get/set X" the brief description should say "Access X"

src/chromo/common.py
HDembinski: If we want CrossSectionData to act more like an array of values, we should inherit from `typing.NamedTuple`, which provides that for free.

src/chromo/models/dpmjetIII.py
HDembinski: I still have some comments on the name and the doc string which I added below.

src/chromo/common.py
HDembinski: Sorry, I did not respond to the issue with `'CrossSectionData' object is not subscriptable`. I am a bit surprised by this, because I thought the original code passed CI tests. I was under the wrong impression that a dataclass is subscriptable, but I was wrong. So we indeed need setattr and getattr, that is correct.

src/chromo/common.py
jncots: There is no problem with dataclasses.astuple. I would be happy not touch the previous implementation, but the line:

```
cross_section[i] += fraction * val
```

gives an error: `'CrossSectionData' object is not subscriptable`.

The only way I found how to iterate over fields is to use getattr.

src/chromo/common.py
jncots: How to iterate over fields in dataclass?

src/chromo/common.py
jncots: _mul_radd is used now.

src/chromo/common.py
jncots: > However, I think we should just implement the method _mul_radd and be done with it. 

Ok

src/chromo/common.py
jncots: I changed to use `cross_section._mul_radd(fraction, cs)`.

src/chromo/models/dpmjetIII.py
jncots: > Note the use _ for the irrelevant name of the setter.

I tried, it doesn't work. Although I thought that this should work.

The long technical name `hA_AA_glauber_trials` was intended to "scare off" users that don't need it. But I guess, if we would really wanted to hide it, then we would make this property private. So, indeed, if it is public, then the simple name is better.

HDembinski: > I tried, it doesn't work. Although I thought that this should work.

Ok, you are right, I am sorry that I remembered that wrong. It is really surprising.

HDembinski (approved):

HDembinski (approved):

PR #131, opened 2023-02-17
Allow to include vertex info via format option system

HDembinski:
The Root writer does not write the vertex info by default to save disk space. It was foreseen to optionally save the vertex info, but a configuration option was missing.

I now added a configuration system for the output format. You can now provide options to the format via colon, e.g. `format:option1:option2[...]`. Currently, there are two formats which accept options.
* `hepmc` accepts `gz`, e.g. you can pass `hepmc` and `hepmc:gz`
* `root` accepts `vertex`, e.g. you can pass `root` and `root:vertex`


afedynitch (approved):
Neat. Windows test failed because of download limit, so feel free to merge.

PR #133, opened 2023-02-19
CLI configuration option

HDembinski:
This adds very generic configuration for the chromo CLI. The configuration file is passed to the CLI via the option `-c` or `--config`.

The configuration file contains normal Python code. A variable called `model` exists when this code is called, which represents the generator instance. The code can then manipulate this instance via the public model API. The typical use case is to change which particles are final state, but complex scenarios are also possible, where the configuration does different things to different models. Since it is Python, one can also import variables and functions from other modules.

Example: make_pi0_stable.cfg
```py
from particle import literals as lp

model.set_stable(lp.pi_0.pdgid)
```
call CLI with this config:
```sh
chromo -s 1 -S 100 -c make_pi0_stable.cfg
```

### Other changes

- Added Pythia8._get_stable 
- Found a fix for `test_format_3` on Windows


HDembinski:
@jncots @afedynitch Can I get a review on this? This is a simple generic solution to add configuration files to the CLI. It follows my conviction that configuration files for Python software should be written in Python, too, unless one has to make sure that the configuration code cannot execute arbitrary commands for safety reasons. This is not the case here.

Since the configuration uses the public API of chromo, we do not have to manually keep the configuration language in sync when we add new configuration features to one of the models. We also do not need extra documentation for the configuration language.

Finally, since the configuration is just Python, I can even use branches or use other packages for convenience.

Example 1:
```py
if model.pyname == "Pythia8":
    # do something specific to Pythia8

# generic configuration for all models 
```

Example 2:
```py
from particle import literals as lp

model.set_stable(lp.Lambda.pdgid, False)
```

src/chromo/cli.py
```diff
@@ -328,6 +344,14 @@ def main():
     task_id = None
     try:
         model = args.model(evt_kin, seed=args.seed)
+        if args.config:
```
jncots, line 347: At first glance it is confusing to use arg.config for config file name and then for the content of the file.

src/chromo/models/pythia8.py
```diff
@@ -165,6 +165,15 @@ def _set_kinematics(self, kin):
     def _set_stable(self, pdgid, stable):
         self._pythia.particleData.mayDecay(pdgid, not stable)
 
+    def _get_stable(self):
+        r = set()
+        for p in self._pythia.particleData.all():
+            if p.tau0 > 1e-5 and not p.mayDecay:
```
jncots, line 171: Isn't `not p.mayDecay` enought? Or at least p.tau0 > 1e-5 **or** not p.mayDecay. Why `and` is here?

src/cpp/_pythia8.cpp
```diff
@@ -110,6 +110,16 @@ PYBIND11_MODULE(_pythia8, m)
         .def("addParticle", py::overload_cast<int, string, string, int, int, int, double, double, double, double, double, bool>(&ParticleData::addParticle), "pdgid"_a, "name"_a, "antiname"_a, "spinType"_a = 0, "chargeType"_a = 0, "colType"_a = 0, "m0"_a = 0, "mWidth"_a = 0, "mMin"_a = 0, "mMax"_a = 0, "tau0"_a = 0, "varWidth"_a = false)
         .def("isParticle", &ParticleData::isParticle)
         .def("findParticle", py::overload_cast<int>(&ParticleData::findParticle))
+        // TODO better style is .def("__iter__", ...) and return make_value_iterator
+        // here, but this leads to segfault which I cannot fix right now
```
jncots, line 114: I don't know whether it is related here, but I returned iterator for Pythia like this (and it worked):

```
py::class_<Event>(m, "Event")
...
.def("__iter__", [](const Event &event) { return py::make_iterator(event.begin(), event.end()); },
                         py::keep_alive<0, 1>() /* Essential: keep object alive while iterator exists */)
```

jncots (approved):
@HDembinski, it is very interesting trick to inject a configuration code. Although a user should be familiar with chromo API. In the config file of crmc a user can just toggle (comment/uncomment) the existing parameters. Probably, what I've seen is just a basic config file. Maybe someday it would be useful to provide such kind of basic config file for chromo.

I have only one real question: the one about definition of stable particles in Pythia.

src/chromo/cli.py
HDembinski: Yes, I was lazy and reused the same attribute. Do you have a suggestion to make it more clear? Perhaps with a comment?

src/chromo/models/pythia8.py
HDembinski: I thought like you, but surprisingly, one needs both. If I remember correctly, without the p.tau0 > 1e-5 condition you also get all the naturally stable particles like gamma, p, etc, which should have tau0 == inf, but have tau0 == 0.

src/cpp/_pythia8.cpp
HDembinski: I tried something like this, including the keep_alive, which I am aware of, but it was segfaulting. I did not have time to figure out why.

src/chromo/cli.py
HDembinski: I now write the code to `args.config_code`. Is that better?

HDembinski: I am not sure if I like it better with args.config_code. Both `args.config` and `args.config_code` store a `str`, so even if you think in static types, it is fair reuse. Perhaps a comment is sufficient to declare that args.config is overridden with the code.

HDembinski:
> Although a user should be familiar with chromo API.

True, but we want them to learn and use our API anyway. If you add a special configuration language that is "simpler", users need to learn the chromo API **and** this configuration language. And we need to document both and keep them in sync. That would be so much work for us without real benefit. As a small project with very limited human resources, we must keep it as simple as possible.

src/cpp/_pythia8.cpp
HDembinski: I checked again, but same result. It is probably because the list contains smart pointers instead of values and somewhere the life-time management goes wrong. pybind11 tries its best to use handle tricky cases like this, but sometimes this has to be handled manually. I don't know how to fix this properly, and since we use this API only internally, I will go over this issue to move forward.

src/chromo/cli.py
HDembinski: I now return `(args, configuration)` from `parse_arguments`, where `configuration` contains the optional configuration code. `args.config` remains untouched and is now used to show the configuration file in the CLI banner.

HDembinski:
I added some notes to test_cli.py, which contains the powerful `run` function that automates a lot of boilerplate to generate and check files in parallel. `run` was improved to also allow passing configuration code. The configuration file is created automatically in a temporary directory just like the output file, so that different tests run in parallel do not interfere with each other.

PR #136, opened 2023-02-27
Fix rangen for EPOS

HDembinski:
This is a hotfix for a mistake I introduced when I changed the random number generation. Without this fix, EPOS eventually runs into an infinite loop at some point after X events generated.

Since this is a simple hotfix, I will merge this myself.


afedynitch:
Fixing bugs is great. Thanks. How did you find it?

PR #138, opened 2023-02-28
Add CRMC performance comparison

HDembinski:
I wanted to check how the results that @jncots got a while ago on the performance comparison between CRMC and chromo look when we use the current dev version of Chromo. I cannot reproduce the large speed-ups of up to 35x. In my comparison, chromo is faster for QGSJet-II-04 and SIBYLL-2.3d and slower for EPOS-LHC. For full transparency, I added the notebook with my code to the folder `bench`.

image

The results look reasonable, because I don't expect chromo to be faster by an order of magnitude. After all, both run the same Fortran codes and that should dominate the overall CPU load. I expect some differences from the random number generator and compiler options, and postprocessing. But these should be all small effects. In the original comparison, was CRMC perhaps compiled in DEBUG mode but Chromo in release mode? This could explain the large differences in speed.

Some general notes:
- The comparison depends on the machine where the benchmark is run. We are testing a mix of CPU and disk load. My computer has a harddrive, results with a SSD should be different. I tried to take that out in my benchmark by writing to /dev/null, which is a virtual file that just discards arriving blocks. 
- For a meaningful comparison, we need to make enough events per generator so that the startup time becomes negligible.


HDembinski:
> The main problem with CRMC performance I noticed was the output part. CRMC flashed every event to disc instead of flash all events at once. This is where 35x occurs. The output was on SSD. Maybe, there are options how to fix this. But I am not familiar much with CRMC, so in the default mode it was like that.

I think that is fine for a benchmark between different projects. We are testing their default against our default, which is fair. We just should make sure that both are release versions, but it sounds like that was the case.

I don't have a SSD on my current computer, but I also didn't see a big difference to my numbers when I wrote to a file instead of /dev/null. But please check.
 
> There is a function in CRMC “bjinta”. If it is commented out, then CRMC faster than "chromo" even for SIBYLL-2.3d.

I guess that depends on Ecm. Assuming that both CRMC and Chromo are equally optimized by the compiler, we should only see the constant overhead from the Python interface for small Ecm values, where the events are small and time spend in Fortran is not large compared to the Python overhead. That's why I ran the benchmark for different values of Ecm, but I did not see this.

Contrary to my intuition, SIBYLL performance drops in Chromo relative to CRMC as Ecm increases.

> I think CRMC uses original random number generator. At lease, I remeber that the call stack of CRMC for generation of random numbers was larger.

Yes, I would think so.
 
> I haven't tried since then to make a comparison. The only thing that can influence change in performance is change of rng since then.

I also think so.
 
> For CRMC, I cannot find now which settings (Debug or Release flags) I used. But it is possible that I used Debug.

I built crmc-v1.8.0, and there the default was to build in release mode. I checked which compiler flags are set by running `VERBOSE=1 make` after `cmake`. They compile with `-O`.

jncots:
I  launched `crmc_comparison.ipynb` and I have slightly different results.

image

The main difference is the run time of QGSJet-II-04. It is longer than in other models. It seems freeze for chromo and for crmc at energies 1e4, 1e5, so I omit these energies.

There is probably another difference. I could not launch `crmc_comparison.ipynb` without providing additional paramameter for CRMC: "-c", which is the path to config file `crmc.param`. Maybe here is the difference? @HDembinski, how do you launch CRMC without it? 

Maybe the difference is in the version. I just found out that I use CRMC v2.0.1. I compiled CRMC with Release.

I switched off the decay of particles in CRMC. In this case the following results appear:
image

HDembinski:
> There is probably another difference. I could not launch `crmc_comparison.ipynb` without providing additional paramameter for CRMC: "-c", which is the path to config file `crmc.param`. Maybe here is the difference? @HDembinski, how do you launch CRMC without it?

When you run crmc from a directory which contains the crmc.param, then you don't need the -c option. That's how I used it.

> The main difference is the run time of QGSJet-II-04. It is longer than in other models. It seems freeze for chromo and for crmc at energies 1e4, 1e5, so I omit these energies.

That is bad, but if it happens for both CRMC and Chromo, it is at least not a bug in Chromo.

HDembinski:
> I switched off the decay of particles in CRMC. In this case the following results appear.

Interesting, I did not think that this makes such a noticable difference. So another thing we should control is which particles are considered "stable" in both models so that the results are comparable. Comparing CRMC without decays to Chromo with decays of non-stable particles is enabled is not fair.

jncots:
I was trying to understand where 35x came from.  Still could not find it, but I notice that when I launched chromo and crmc from command line, chromo was almost always was faster. It was not 35x, but 2.5x. The below plot is produced using slightly modified script:

```
nevent = int(1e5 / sqrt(energy))
```

I also changed lines

```
subp.run(cmd, stdin=subp.PIPE, stdout=subp.PIPE)
```

to contain `stdin=subp.PIPE, stdout=subp.PIPE`. It somehow gives the same speed as in command line. Not sure how it works, but  without it there are some problems with speed in python script.

image


I think when the programs are compared on small number of events, then a loading time plays a significant role in estimations. On long runs, when contribution of loading time is reduced, we can measure the real execution time per event. Plus, some modules loads data from files. For, example, QGSJet loads a large file, and crmc does this faster than chromo. It is the main reason, why on small number of events crmc faster.

Since chromo is Python, it loads usually longer than c++ crmc. When it comes to simulations, I think the run times should be comparable. I guess, the problem with crmc is output.

---
We can conclude:
chromo is **slightly faster** in event simulations, but **slower for loading**.

jncots:
> How did you get that version? The latest release here is v1.8.0 https://web.ikp.kit.edu/rulrich/crmc.html

I compiled it from https://gitlab.iap.kit.edu/AirShowerPhysics/crmc.

jncots (approved):

HDembinski:
> to contain `stdin=subp.PIPE, stdout=subp.PIPE`. It somehow gives the same speed as in command line. Not sure how it works, but without it there are some problems with speed in python script.

I did not expect that this makes a difference, but we generate more output in chromo through the progress bar that CRMC does not have. `stdin` should be irrelevant, but `stdout=subp.PIPE` redirects the output into a string buffer. That may be faster than the output stream that a Jupyter kernel otherwise uses to redirect the output of the command into the notebook. The latter has to pass the text through the html server and javascript, which costs time.

HDembinski:
@jncots Thank you for checking, I invite you to commit your patches to the notebook to this branch.

jncots:
I added patches to the file. Noticeable differences in run time are only seen when the name of the output file is omitted (-f option). For redirection to devnull or single file the differencies are smaller. 

Anyway, I guess the PR can be merged.

PR #143, opened 2023-03-02
Pythia8 interfaces for decay

jncots:
@HDembinski, I need a couple interfaces for using Pythia8 decay capabilities. I use them for some time privately. I guess it will be useful to have those functions exposed anyway when the decay capabilities of Pythia8 will be introduced in chromo. Can I add them now?

### DIFF
--- a/src/cpp/_pythia8.cpp
+++ b/src/cpp/_pythia8.cpp
@@ -209,6 +209,7 @@ PYBIND11_MODULE(_pythia8, m)
         .def("init", &Pythia::init, py::call_guard<py::scoped_ostream_redirect, py::scoped_estream_redirect>())
         .def("next", py::overload_cast<>(&Pythia::next))
         .def("readString", &Pythia::readString, "setting"_a, "warn"_a = true)
+        .def("forceHadronLevel", &Pythia::forceHadronLevel, "find_junctions"_a = true)
         .def_readwrite("particleData", &Pythia::particleData)
         .def_readwrite("settings", &Pythia::settings)
         .def_readwrite("event", &Pythia::event)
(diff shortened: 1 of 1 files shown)


src/cpp/_pythia8.cpp
```diff
@@ -209,6 +209,7 @@ PYBIND11_MODULE(_pythia8, m)
         .def("init", &Pythia::init, py::call_guard<py::scoped_ostream_redirect, py::scoped_estream_redirect>())
         .def("next", py::overload_cast<>(&Pythia::next))
         .def("readString", &Pythia::readString, "setting"_a, "warn"_a = true)
+        .def("forceHadronLevel", &Pythia::forceHadronLevel, "find_junctions"_a = true)
```
HDembinski, line 212: I cannot check right now (on my phone): forceHadronLevel should support the same arguments as the C++ version. If that's the case, it is all good.

src/cpp/_pythia8.cpp
```diff
@@ -244,6 +245,8 @@ PYBIND11_MODULE(_pythia8, m)
         .def("vt", event_array_v<Vec4_tt>)
         .def("parents", event_array_parents)
         .def("children", event_array_children)
+        .def("reset", &Event::reset)
+        .def("append", py::overload_cast<int, int, int, int, double, double, double, double, double, double, double>(&Event::append), "pdgid"_a, "status"_a, "col"_a, "acol"_a, "px"_a, "py"_a, "pz"_a, "e"_a, "m"_a = 0, "scale"_a = 0, "pol"_a = 9.)
```
HDembinski, line 249: Same point as above.

HDembinski (commented):
Look good to me, see minor comments.

HDembinski (approved):

src/cpp/_pythia8.cpp
jncots: I checked it once again. forceHadronLevel has only one argument. I use a default value.  It seems to work fine.

src/cpp/_pythia8.cpp
jncots: I use only one of the versions of append and arguments to that overload are the same as in this interface.

PR #145, opened 2023-03-03
Fix for seed issue in Pythia8; return correct seed when seed is None

HDembinski:
Since I changed the PRNG, the correct way to use a random seed is to pass seed=None to `MCRun.__init__`. However, this introduced a bug for Pythia8, which is the only generator that is currently not using the Numpy PRNG. `None` in this case was directly forward to the Pythia config, which failed since it cannot handle that value. This patch fixes the issue. Pythia8 now gets the seed which Numpy generates when `seed=None` is used.

I also fixed a related bug, which affects all models. If `seed=None` was used, `MCRun.seed` also incorrectly returned `None`. This is not intended behavior, we want to know the seed that was actually used even if we do not specify it, so that we can reproduce the output of the generator. I changed the seed property in `MCRun` now so that the actual seed is returned that Numpy generated when `seed=None` is used.

### DIFF
--- a/src/chromo/common.py
+++ b/src/chromo/common.py
@@ -554,7 +554,6 @@ class MCRun(ABC):
         assert hasattr(self, "_frame")
         self._lib = importlib.import_module(f"chromo.models.{self._library_name}")
 
-        self._seed = seed
         self._rng = np.random.default_rng(seed)
         if hasattr(self._lib, "npy"):
             self._lib.npy.bitgen = self._rng.bit_generator.ctypes.bit_generator.value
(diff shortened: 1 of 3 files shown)


afedynitch (approved):
Approved. 

Using a private API is an bit unfortunate but if it breaks there would be plenty of ways around to fix it.

afedynitch (approved):

HDembinski:
@afedynitch Thanks for the approval. I just made some last minute optical changes and merged the changes in main. I will squash and merge once the tests succeed.

afedynitch:
OK, thanks for fixing this.

PR #146, opened 2023-03-03
Gamma+p example

HDembinski:
The Leptohadronic modelling workshop made me curious what needs to be done so that Pythia8.3 supports gamma+p interactions like SOPHIA. Turns out, Pythia8.3 as implemented in Chromo, already supports that. I added a simple example which compares SOPHIA with Pythia-8.3. This can be used to advertise Chromo to the workshop participants.

Comments and feedback are welcome.


HDembinski:
@afedynitch @jncots Could you have a quick look at this? I think this an uncontroversial change. Perhaps the example can be improved, but it is nice to have already.

afedynitch (approved):
Works great.

PR #147, opened 2023-03-05
Fix workflow file

jncots:
There were 2 problems in the workflow file, which were not tested while they were not on a main branch. It resulted in the condition which doesn't allow to lauch PyPI upload step. I tested the problematic conditions on test repository. So it should work now. Please approve this PR, so I could make upload to PyPI.


afedynitch (approved):
ОК

PR #149, opened 2023-03-06
Fix workflow file 2

jncots:
This is a second trial to launch PyPI. There is a problem with conditions. It seems that `github.event.base_ref` is null. At least this is what it shows when the workflow was launched on a fork. On my test repository `github.event.base_ref` has correct value. I tested conditions on a fork of chromo and made a release. It worked. I hope this time it should work.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -50,7 +50,7 @@ jobs:
 
 
   wheels:
-    if: ${{ github.event_name == 'workflow_dispatch' || (github.event.base_ref == 'refs/heads/main' && startsWith(github.ref, 'refs/tags/v')) }}
+    if: ${{ github.event_name == 'workflow_dispatch' || startsWith(github.ref, 'refs/tags/v') }}
     name: ${{ matrix.py }} ${{ matrix.os }} ${{ matrix.arch }}
     runs-on: ${{ matrix.os }}
     strategy:
@@ -163,6 +163,9 @@ jobs:
         with:
           submodules: true
 
+      - name: Dump GitHub context
+        run: echo '${{ toJSON(github) }}'
+
       - uses: actions/setup-python@v4
         name: Install Python
         with:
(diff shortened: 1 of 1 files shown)


HDembinski:
I use `contains(github.event.ref, '/tags/')`, see
https://github.com/scikit-hep/pyhepmc/blob/main/.github/workflows/wheels.yml#L83
Your condition may also work. I wonder why you not just follow pyhepmc. I made an effort make the CI code in pyhepmc as simple as possible.

Other suggestions for a follow-up PR, which should clean up the code a little bit:

Remove wheels_on_PR to simplify the code. We don't need to automatically check whether generating wheels still works after each PR, because most PRs won't affect the wheels. I first thought that you want to generate "nightlies" automatically, but then you should generate wheels for all platforms, so use the step "wheels" instead of "wheels_for_PR". 

In all my repositories, generating the wheels just works, unless I actually touch the CI code. And in that case, one can trigger the workflow manually to check that wheels still work.

Remove `name:` tags. They are often useless and overused. They fill a purpose like comments in code. In style guides for comments in code, it is discouraged to do something like this
```py
def evaluate(x):  # evaluates x
    ...
```
because the comment does not add any value, it is just noise. It is the same with Github CI scripts, where I often see superfluous `name:` tags, also in otherwise fine projects. That is nonsense that should not be copied (avoid "cargo cult"). Examples where the name is redundant:
- everywhere in `upload_to_PyPI`
- `Install Python` https://github.com/impy-project/chromo/blob/2cd6cb8c38efa1a32a736ba5357bbac34954ea55/.github/workflows/release.yml#L170

For context, I think that you have places where `name:` tags are useful, where they serve as comments what the following step is doing. The tag is useful when that is not immediately obvious from the executed commands.
This is good use:
https://github.com/impy-project/chromo/blob/2cd6cb8c38efa1a32a736ba5357bbac34954ea55/.github/workflows/release.yml#L31
And where in wheels. This list is not complete.

HDembinski (approved):
This is fine, if it works, but see my comment for suggested follow-up PR to do cleanup.

jncots:
Thank you @HDembinski  for the comment. I will  improve the workflow file in next PRs.

Issue #142, opened 2023-03-01
Support gamma + A?


HDembinski:
In the [SFB Leptohadronic workshop](https://indico.uni-wuppertal.de/event/230/), I learned about the importance of gamma+p and gamma+A interactions for this community.

- Do we have models which support gamma + A?
- Can we enable this process in some models, where the model supports it but our interface does not?
- A function in chromo.utils would be useful which returns a list of models that support a given process. The function can use the public interface of each model to check whether projectile and target are supported.

PR #153, opened 2023-03-13
Support for gamma-gamma in Pythia and Phojet 

### LINKED ISSUE #152: How to support photon-photon collision?
I try to support photon-photon in Pythia-8.3. Doing that in `Pythia8` is easy, but I don't know how to fix the issues with the computation of elab2ecm and ecm2elab.

There is no rest frame for photon-photon, so the our calculation of Ecm from Elab gives zero exactly, and ecm2elab produces a zero-division error.

HDembinski:
Closes #152 if we find the answer to the question posed in the issue. I enabled the support in Pythia8, but we cannot run it like this because we always compute everything in the Ecm system and that does not exist for gamma-gamma.


afedynitch:
Will also close #142

jncots (approved):
Thank you - it is very important changes. Only not critical amendments in comments are required.

PR #158, opened 2023-06-29
Remove some unused packages and provide a Writer base class

HDembinski:
During writing of the ICRC proceeding, I noticed that it would be more consistent to have a base class for the Writer classes. This patch adds such a base class and makes the other classes derive from it. I think we originally had such a base class, but I removed it when I removed all the writer classes. At the time, I thought we don't need them, but I then realized that we need them for the CLI.

In addition, I removed the graphviz package from the list of packages needed for testing, which is no longer necessary since  I replaced the graphviz code in pyhepmc.


HDembinski:
@afedynitch @jncots Can I get a review on this? It is not a big change.

afedynitch (approved):
Indeed, not a big change. It's good to go

PR #161, opened 2023-07-13
add refill_decay_stack

jncots:
When Pythia8 is used for decay of particles, it is required to put the particle that should be decayed on a `event` stack . Currently it is only possible to do like this:

```python
for i in range(len(stack)):
    pid = stack.pid[i]
    en = stack.energy[i]
    m = self._pythia.particleData.findParticle(pid).m0
    pz = sqrt((en - m)*(en + m))
    self._pythia.event.append(pid, 91, 0, 0, 0, 0, pz, en, m)
```

So, to put a lot of particles from vector stack, it is needed to call `append` for every particle. This is time consuming because of Python `for` loop and the call to `append` itself. The time of filling loop is the same as the time required for decay in Pythia, which from calculation point of view are not comparable tasks at all. 

The added function `self._pythia.refill_decay_stack(stack.pid, stack.energy)` accepts 2 numpy arrays with pid and energy, and calls the above loop in C++. These significanly (~40 times for my test cases) reduces the time for filling the `event` array.


afedynitch:
To be more generic, I think that there should be a version of the function taking a vector of four-vectors and a vector of PIDs. This is also required for building a replacement for the decay afterburner, since the particles on the event stack have 4-momenta defined with px, py != 0.

src/cpp/_pythia8.cpp
```diff
@@ -103,6 +104,28 @@ py::array_t<int> event_array_children(Event &event)
     return py::array_t<int>(shape, strides, ptr1);
 }
 
+// refills "event" stack with particles 
+// having pid and energy
+void refill_stack(Event &event,
```
HDembinski, line 109: Please change name to `fill_event` When you fill something, it is usually implied that the original content is deleted.

src/cpp/_pythia8.cpp
```diff
@@ -103,6 +104,28 @@ py::array_t<int> event_array_children(Event &event)
...
+    double pz = 0.;
+    
+    event.reset();
+    for (int i = 0; i != pid.size(); ++i) {
+        id = pid.at(i);
+        en = energy.at(i);
+        mass = pd.findParticle(id)->m0();
+        pz = std::sqrt((en - mass)*(en + mass));
+        event.append(id, 91, 0, 0, 0., 0., pz, en, mass);
+    }
```
HDembinski, line 126: This is not an efficient way to iterate over a numpy array. You should request the raw memory buffer from the array and then iterate over that. It should be even faster.

HDembinski: The simplest way is to call `int* pid_raw = pid.data();` to get the raw buffer.

HDembinski: Actually, no, don't use this, because this only works if `py::array_t<int>` is a contiguous c-style array. This is not guaranteed in Chromo. Please read https://pybind11.readthedocs.io/en/stable/advanced/pycpp/numpy.html#arrays for more information. You should use this code instead: https://pybind11.readthedocs.io/en/stable/advanced/pycpp/numpy.html#direct-access

HDembinski (commented):
I agree with Anatoli, this should be more general. You should not assume that the user passes the energy. The function should accept the four-momentum in the way how it is provided by our EventData attributes, so four numpy arrays called px, py, pz, en.

A test is missing, as Anatoli pointed out. A useful unit test is to reset and fill an event in Python with `append` and then check you get the same if you fill using the new `fill` method. `Event` objects are not equal-comparable, but you can consider two events equal if they arrays are equal (pid, status, m, px, ...)

src/cpp/_pythia8.cpp
HDembinski: The user should also have the option to set another status code than 91.

HDembinski: And mass should also come from the user, not from ParticleData.

src/cpp/_pythia8.cpp
HDembinski: Actually no, the name should be just `fill`, since this is a method of `Event`. Repeating "event" in the name is redundant.

jncots: I change the name to `fill` and put it as method for Event.

src/cpp/_pythia8.cpp
jncots: * To iterate in an efficient way I used `py::array_t.unchecked<1>()`
* To append efficiently the underlying `Event.enty` vector is accessed for size reservation. 
* 2 version of `fill_event` are added. One to append pid, status, px, py, pz, energy, mass and another is the initial version to append pid and energy (it is the variant that I use, as I don't need momenta).

jncots:
* `pythia.fill_event` and `pythia.event.fill` are added
* `pythia.event.fill` is a general case
* `pythia.fill_event` is a special case where only energy is required and mass is taken from Pythia itself.
* tests of both functions are added

src/cpp/_pythia8.cpp
```diff
@@ -103,6 +105,73 @@ py::array_t<int> event_array_children(Event &event)
...
+void fill_event(Event &event,
+                    py::array_t<int> &pid,
+                    py::array_t<int> &status,
+                    py::array_t<double> &px,
+                    py::array_t<double> &py,
+                    py::array_t<double> &pz,
+                    py::array_t<double> &energy,
+                    py::array_t<double> &mass) 
+{
+    const double spare_mem = 1.5L;
```
HDembinski, line 118: If you use reserve, you should not use spare_mem, I think. Unlike for vector::push_back, you won't get O(1) asymptotic behavior this way.

src/cpp/_pythia8.cpp
```diff
@@ -103,6 +105,73 @@ py::array_t<int> event_array_children(Event &event)
...
+
+    for (int i = 0; i != fill_size; ++i) {
+        id = pid_[i];
+        en = energy_[i];
+        mass = pd.findParticle(id)->m0();
+        pz = std::sqrt((en - mass)*(en + mass));    
+        event.append(pid_[i], 91, 0, 0, 0., 0., pz, en, mass);
+    }    
+}
+
```
HDembinski, line 174: Is this special case really necessary? We keep the C++ bindings minimal, because these can break as we are using private details of Pythia that are not guaranteed to be stable by any convention.

I would prefer to use the generic fill_event and compute pz with numpy arrays. Might even be faster.

Again, the mass should come from the particle itself via the generated mass and not from ParticleData. For long-lived particles, there is not much difference, but for short-lived particles, using the generated_mass is important to maintain energy conservation.

src/cpp/_pythia8.cpp
```diff
@@ -237,6 +305,12 @@ PYBIND11_MODULE(_pythia8, m)
                  return result;
              })
 
+        .def("fill_event",
+             [](Pythia &self, py::array_t<int> &pid,
+                              py::array_t<double> &energy)
+             {
+                fill_event(self.event, self.particleData, pid, energy);
+             })
```
HDembinski, line 313: You should not have "fill_event" and "fill". "fill_event" is still a bad name, I already argued against it. See my previous point about removing "fill_event" completely. If that's not feasible, then please please choose a more fitting name that does not include the redundant "event" part.

src/cpp/_pythia8.cpp
```diff
@@ -257,6 +331,16 @@ PYBIND11_MODULE(_pythia8, m)
...
+            [](Event &self, 
+                py::array_t<int> &pid,
+                py::array_t<int> &status,
+                py::array_t<double> &px,
+                py::array_t<double> &py,
+                py::array_t<double> &pz,
+                py::array_t<double> &energy,
+                py::array_t<double> &mass) {
+                fill_event(self, pid, status, px, py, pz, energy, mass);
+            })
```
HDembinski, line 344: What about col, acol, scale, pol?

tests/test_pythia8_fill.py
```diff
@@ -0,0 +1,140 @@
+import chromo
+import numpy as np
+from chromo.common import EventData
+from pathlib import Path
+import importlib
+from .util import run_in_separate_process
+
+
+def init_pythia8():
+    lib = importlib.import_module(f"chromo.models._pythia8")
```
HDembinski, line 10: There is no need to use importlib.import_module here, just do a normal import.

HDembinski: `from chromo.models import _pythia8 as lib`

tests/test_pythia8_fill.py
```diff
@@ -0,0 +1,140 @@
...
+    pythia = lib.Pythia(str(xml_path), False)
+    seed = 1
+    pythia.settings.resetAll()
+    pythia.readString("Random:setSeed = on")
+    pythia.readString(f"Random:seed = {seed}")
+    pythia.readString("Print:quiet = on")
+    pythia.readString("ProcessLevel:all = off")
+    pythia.readString("ParticleDecays:tau0Max = 1e100")
+    pythia.init()
+    return pythia
```
HDembinski, line 22: Why are you doing this? You can test this functionality more easily on a Pythia8 instance. You can access the underlying Pythia object in the Pythia8 instance using private attributes. You can set extra options with the `config` argument to the Pythia8 init.

HDembinski (commented):
Still some questions to resolve. Mostly whether having both `fill_event` (should be renamed) and `fill` is necessary. Having both is going to cause a maintenance burden, which is only justified if `fill_event` gives a significant advantage over the more general `fill`.

src/chromo/common.py
```diff
@@ -230,7 +230,7 @@ def __getstate__(self):
             self.vy.copy(),
             self.vz.copy(),
             self.vt.copy(),
-            self.parents.copy(),
+            self.parents.copy() if self.children is not None else None,
```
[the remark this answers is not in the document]
jncots: Yes, it's wrong. It should be `self.parents.copy() if self.parents`

src/cpp/_pythia8.cpp
jncots: > Unlike for vector::push_back, you won't get O(1) asymptotic behavior this way.

I didn't get what you mean. But thanks to this comment I realized that I need to reserve  `entry` 2x larger of event, because it is used to save initial entry + decayed particles. I checked that for the default settings the array usually increase 1.5 - 1.8 times. If all particles decay, then the size increases 5x. I guess 2x is conservative estimation.

HDembinski: Ah, I see now what this factor 1.5-1.8 was for. I thought you try to imitate how std::vector grows.

Anyway, you are overoptimising this. Pythia is not shrinking the particle vector in the event. So all this optimisation of trying to reserve enough space for all particles will only run once or twice until the particle vector has grown large enough to store all particles. Afterwards, it will just keep its size.

I think you should remove all this code to reserve space in the vector and just call the public API. Use Event::reset() and Event::append(). This is going to be efficient enough.

jncots: Ok. I leave a bare minimum version of the function. I will change when I am sure that the optimization with reserve works.

src/cpp/_pythia8.cpp
jncots: My case: I have only pid, energy arrays. Each 1e4-1e6 long. That is why I need to put them efficiently. So I don't have px, py, pz, or mass. I can use numpy for mass and pz calculations. The only thing which I find not nice: I need initiate long px and py zero arrays.

Anyway, I guess it is not such a big bottleneck. So, I removed the special case function.

src/cpp/_pythia8.cpp
jncots: `fill_event` was a Pythia method whereas `fill` is Event method. The names were as the were to reflect that fact.  I remove the special case mostly because the functions were the methods for different objects, which didn't look good.

src/cpp/_pythia8.cpp
jncots: If I would add them, I would need to initialize another 2 long zero numpy arrays that I don't need.

tests/test_pythia8_fill.py
jncots: By the way, this is how I initialize and use pythia8 shared library. I was trying to use Pythia8 instance, but it's initialization times are very long. It would be nice to have some kind of shortcut initialization (without kinematics and loading tables), when Pythia8 is used only as afterburner.

For this test I can change the code to use Pythia8 instance, but it will add another ~10 sec to tests.

jncots:
I leave only bare minimum `fill` function.

src/cpp/_pythia8.cpp
HDembinski: Yes, fair enough. If we ever need them, we can add them.

tests/test_pythia8_fill.py
HDembinski: Where is the difference in loading time coming from? I think it is coming only from the standard settings which enable all SoftQCD processes. This requires computing the inelastic cross-section. You can use config argument to set your own settings and override the standard. Then the initialization should be as fast as the manual one you are doing here.
Correct me if I am wrong.

src/cpp/_pythia8.cpp
HDembinski: I don't understand what you mean. I am indifferent how the function is named in C++, but in Python the name should be short and nice.

jncots: I meant `pythia.event.fill` is method of `event` and `pythia.fill_event` is method of `pythia`, i.e. similar functionality but related to different objects.

tests/test_pythia8_fill.py
jncots: I found the obstacle that didn't allow to initiate fast. I changed Pythia8 class constructor, so when no kinematics is passed it works in decay mode.

jncots:
@HDembinski, I forgot to mention that Pythia8 `fill` function is ready. In tests the library is initialized using Pythia8 object in "decay mode" (without loading tables).

src/chromo/models/pythia8.py
```diff
@@ -71,7 +71,7 @@ class Pythia8(MCRun):
         + "/releases/download/zipped_data_v1.0/Pythia8_v002.zip"
     )
 
-    def __init__(self, evt_kin, *, seed=None, config=None):
+    def __init__(self, evt_kin=None, *, seed=None, config=None, banner=True):
```
HDembinski, line 74: evt_kin should not have a default value of none, please change that back. Noramlly, you cannot initialize a generator without initial state. The situation for Pythia8 is special, but that does not justify breaking this general expectation. You can still pass None for evt_kin even if it is not the default value.

This function signature was specifically designed with the * so that the only accepted positional argument is evt_kin.

src/chromo/models/pythia8.py
```diff
@@ -118,10 +121,20 @@ def _cross_section(self, kin=None):
...
+        config = self._config[:]
+        config += [
+            # use our random seed
+            "Random:setSeed = on",
+            # Pythia's RANMAR PRNG accepts only seeds smaller than 900_000_000,
+            # this may change in the future if they switch to a different PRNG
+            f"Random:seed = {self.seed % 900_000_000}",
+            # reduce verbosity
+            "Print:quiet = on",
+        ]
```
HDembinski, line 134: Please share this common code between set_decay_mode and set_kinematics to keep it DRY. You can add these common options to the config in the init.

src/cpp/_pythia8.cpp
```diff
@@ -103,6 +103,32 @@ py::array_t<int> event_array_children(Event &event)
...
+    auto pz_ = pz.unchecked<1>();
+    auto energy_ = energy.unchecked<1>();
+    auto mass_ = mass.unchecked<1>();
+
+    event.reset();
+    for (int i = 0; i != pid.size(); ++i) {
+        event.append(pid_[i], status_[i], 0, 0,
+                      px_[i], py_[i], pz_[i], energy_[i], mass_[i]);
+    }    
+}
```
HDembinski, line 130: Please adjust the indentation or install an autoformatter. We should add one to pre-commit...

src/cpp/_pythia8.cpp
```diff
@@ -257,6 +281,16 @@ PYBIND11_MODULE(_pythia8, m)
...
+            [](Event &self, 
+                py::array_t<int> &pid,
+                py::array_t<int> &status,
+                py::array_t<double> &px,
+                py::array_t<double> &py,
+                py::array_t<double> &pz,
+                py::array_t<double> &energy,
+                py::array_t<double> &mass) {
+                fill(self, pid, status, px, py, pz, energy, mass);
+            })
```
HDembinski, line 294: The lambda is superfluous, just pass fill here.

tests/test_pythia8_fill.py
```diff
@@ -0,0 +1,100 @@
+import chromo
+from chromo.common import EventData
+from .util import run_in_separate_process
+import pytest
+import sys
+
+pytestmark = pytest.mark.skipif(
```
HDembinski, line 7: Do you need the return value? If yes, call it `_` because its name is irrelevant and the `_` signifies that. Note that there is also `pytest.skipif`, maybe you want that instead?

tests/test_pythia8_fill.py
```diff
@@ -0,0 +1,100 @@
...
+    sys.platform == "win32", reason="Pythia8 does not run on windows"
+)
+
+
+def init_pythia8():
+    config = ["ProcessLevel:all = off" "ParticleDecays:tau0Max = 1e100"]
+    pythia8 = chromo.models.Pythia8(seed=1, config=config, banner=False)
+    return pythia8._pythia
```
HDembinski, line 15: Why is this a function? It looks like you use this code exactly once. So I would inline it for clarity.

src/cpp/_pythia8.cpp
jncots: Removed

src/chromo/models/pythia8.py
jncots: Default None is removed.

src/chromo/models/pythia8.py
jncots: Moved the code to constructor

src/cpp/_pythia8.cpp
jncots: Adjusted

tests/test_pythia8_fill.py
jncots: Removed. No, return value is not needed. `pytest.skipif` doesn't exist, only `pytest.mark.skipif`.

tests/test_pythia8_fill.py
jncots: Removed

tests/test_pythia8_fill.py
jncots: It turns out that the name is required https://docs.pytest.org/en/7.1.x/reference/reference.html#globalvar-pytestmark

HDembinski: This code does not require the variable:
```py
if sys.platform == "win32":
     pytest.skip("Pythia8 does not run on windows", allow_module_level=True)
```

HDembinski: You are right, `pytest.skipif` does not exist, but `pytest.skip` does. When you use `skip` as a decorator, the `if` has to be part of the function, but when you call `skip` in normal imperative code, you can also write the `if` in normal Python.

jncots: I see. Thanks for explanation.

afedynitch:
Didn't this became redundant with #167 ?

HDembinski:
> Didn't this became redundant with #167 ?

No this should be merged first and then #167 should be rebased.

jncots:
> No this should be merged first and then https://github.com/impy-project/chromo/pull/167 should be rebased.

Yes, I was thinking about something like this.

Issue #163, opened 2023-07-18
'CompositeTarget' object has no attribute 'copy'


jncots:
EventData.copy() produces an error, if kinematics contains CompositeTarget:

```python
File [~/chromo/src/chromo/kinematics.py:121](https://vscode-remote+ssh-002dremote-002bvscode.vscode-resource.vscode-cdn.net/hetghome/antonpr/quick_tests/09_pythia_afterburner/~/chromo/src/chromo/kinematics.py:121), in EventKinematicsBase.copy(self)
    117 def copy(self):
    118     return EventKinematicsBase(
    119         self.frame,
    120         self.p1,
--> 121         self.p2.copy() if isinstance(self.p2, CompositeTarget) else self.p2,
    122         self.ecm,
    123         self.plab,
    124         self.elab,
    125         self.ekin,
    126         (self.beams[0].copy(), self.beams[1].copy()),
    127         self._gamma_cm,
    128         self._betagamma_cm,
    129     )

AttributeError: 'CompositeTarget' object has no attribute 'copy'
```

jncots:
In addition to that:

```bash
  File "/hetghome/antonpr/quick_tests/09_pythia_afterburner/pythia_test.py", line 33, in init_event
    return event.copy()
  File "/hetghome/antonpr/chromo/src/chromo/common.py", line 246, in copy
    copies = self.__getstate__()
  File "/hetghome/antonpr/chromo/src/chromo/common.py", line 233, in __getstate__
    self.parents.copy(),
AttributeError: 'NoneType' object has no attribute 'copy'
```

PR #164, opened 2023-07-19
Chromo release 0.4

jncots:
This is a PR towards chromo's 1.0 release. I hope it should be quick. There are a couple of issues.

* The biggest current issue is #127. It is a problem for the source code, but wheels should work. This should be ignored for this release.
* There is a small bug #163 which I will fix.
* Licenses:
```
DPMJet/PhoJet -> Anatoli Fedynitch
EPOS -> Tanguy Pierog and Klaus Werner  (permission of  usage)
Sibyll -> Felix Riehn (permission of usage)
QGSJet -> Sergei Ostapchenko (permission of usage)
UrQMD-3.4 - ?
Sophia - ?
Pythia6 - ?
Pythia8 - GPL2.0
```
Anatoli suggested to put 3 permissions to corresponding source code directories. I guess `license.txt` with something like:

```
Sibyll- Redistribution Notice

This software is distributed as part of "Chromo" with the permission of the original author. Please refer to the original license for terms and conditions. No warranty provided.
```
should be fine.

What else should be done for release?


jncots:
The above fixes #163.

jncots:
Preliminary versions of licenses have been added as described above.

HDembinski (approved):
Look ok to me.

HDembinski (changes requested):
It looks ok, but the version number should not be 1.0. There are some things that we have not settled in the API, where we probably need to make breaking changes. A 1.0 version indicates that the product is stable. Chromo is not stable yet. Previously, we had 0.3.0 as the version number and shoud still be the version.

HDembinski:
@afedynitch Regarding the version number 1.0, please see #165 why I argue against going for a 1.0 at this point.

1.0 means you have a stable product with a stable API, while 0.x indicates that the API is not finalized yet. Our API is not fixed yet.

afedynitch:
> @afedynitch Regarding the version number 1.0, please see #165 why I argue against going for a 1.0 at this point.
> 
> 1.0 means you have a stable product with a stable API, while 0.x indicates that the API is not finalized yet. Our API is not fixed yet.

Oh, I totally agree with that. It will be changed. But since there is an interface change in kinematics between the pypi and even the current main, we should bump it up to 0.4. There are also now some additional changes that would justify upping the minor version.

I'm looking into #160 now, and there are some warnings in test about decays in PHOJET & DPMJET, which shouldn't be there. All of these models can decay the same stuff as PYTHIA6, since this is what is used for decays.

I would prefer getting the final state stuff unified, it is quite important. Then merge it into main, and rebase of FF this PR onto the latest main. Target time is for this one is some time next week, since we removed the --pre flag from the poster ;).

afedynitch:
So in principle, this seems complete. We should try to finalize the #169 PR, and possibly #167, and merge those before proceeding here.

jncots:
We currently have 6 open PRs that are nearly ready for merging.

### Proposed Merging Order:

* [ ] **Sibyll\* - Approved**
   - Added the Sibyll23Star model.
   - Included cross sections.

* [ ] **Test of Final State - Ready for Merging**
   - Addressed fixes for PhoJet and DpmJet stable/unstable settings.

* [ ] **Unified Treatment of Beam Particles - Review Required**
   - Enables reading/writing to HepMC files.

* [ ] **Pythia8 Afterburner - Review Required**
   - Validated if particles requested for decay have indeed been decayed, decaying undecayed particles

* [ ] **Stop F2PY Caching - Review Required with Unresolved Conversations**
   - Resolved issues related to f2py pyf file generation.

* [ ] **Chromo Release 0.4 PR - To be Closed After All the Above**

HDembinski:
> graphviz don't show loops, i.e. multilple particles cannot originate from one vertex and then merge to another. If such situation occurs, then only there first particle is displayed, all other are missing

If I remember correctly, this is not only an issue with graphviz but already with HepMC3, which models the event as a directed acyclic graph, so loops are not supported. I think it is unfixable and we need a workaround, as before. So far I used the workaround to remove all the preshower history when converting from a Pythia event to HepMC3. Is there an issue with this workaround after your changes?

jncots:
> Is there an issue with this workaround after your changes?

The problem with cyclic graph appears in Phojet and Dpmjet. As a workaround, when conversion to hepmc occurs, only the beam and final particles are used. I hope that we will figure out how to write the whole history later.

HDembinski:
> The problem with cyclic graph appears in Phojet and Dpmjet. As a workaround, when conversion to hepmc occurs, only the beam and final particles are used. I hope that we will figure out how to write the whole history later.

This is worse than the workaround I use for Pythia. In Pythia, I only have to remove the preshower information, which is acceptable, because as an experimentalist you cannot exploit this information. The decay history of short or long-lived particles is kept, however. It would be great to keep the decay history in DPMJet / PhoJet as well.

jncots:
> This is worse than the workaround I use for Pythia.

I agree. I will try to fix it.

afedynitch:
> Dear @HDembinski, @afedynitch, we currently have 6 open PRs that are nearly ready for merging.
> 
> ### Proposed Merging Order:
> * [ ]  **Sibyll* - Approved** ✅
>   ....

The order is fine. Please start merging one by one

HDembinski:
@jncots A release of pyhepmc with the required extension is currently running https://github.com/scikit-hep/pyhepmc/releases/tag/v2.13.0. If all goes well, it will be online on PyPI in about an hour. To use it in chromo, you need to use the new keyword `fortran=False` wherever you call the `GenEvent.from_hepevt` method.

jncots:
> @jncots A release of pyhepmc with the required extension is currently running https://github.com/scikit-hep/pyhepmc/releases/tag/v2.13.0.
@HDembinski, excellent! Thank you.

HDembinski:
@jncots There were some issues with the release job due to outdated CI action scripts, but now the updated pyhepmc package is available on PyPI. https://pypi.org/project/pyhepmc/

jncots:
@HDembinski do we need python3.8 on Linux? If yes, then https://pypi.org/project/pyhepmc/2.13.2/#files doesn't have version for python3.8 and it should be added.

jncots:
@afedynitch, do we need to fix #177  and #178 for the release?

afedynitch:
> @afedynitch, do we need to fix #177 and #178 for the release?

Does or prevent HEPMC+RIVET from functioning?

jncots:
> Does or prevent HEPMC+RIVET from functioning?
No, HEPMC+RIVET works fine. For Phojet there is no history, but it is not so critical. For Dpmjet intermediate particles has wrong energies, so they are end up in HEPMC.

afedynitch:
> > Does or prevent HEPMC+RIVET from functioning?
> 
> No, HEPMC+RIVET works fine. For Phojet there is no history, but it is not so critical. For Dpmjet intermediate particles has wrong energies, so they are end up in HEPMC.

Ok, then this is not critical for this release. We’ll do it in the next release.

I’ll go through the review a bit later today, so we can merge this soon

setup.cfg
```diff
@@ -1,7 +1,7 @@
 [metadata]
 name = chromo
 # version = attr: chromo.__version__
-version = 0.3.0rc2
+version = 0.4.0rc1
```
afedynitch, line 4: How do you want to proceed after this is merged? You modify this on the main branch to promote to real pypi?

afedynitch (approved):
OK, good to merge!

I’m just wondering how you’re going to promote this pre-release to full release? Modify the version tag in main directly?

setup.cfg
jncots: On real PyPI the version would be 0.4.0. I wanted to submit it to Test PyPI, but build for python3.8 on Linux doesn't pass tests because latest `particle` is not available for python3.8.

afedynitch:
Overriding @HDembinski review, because the main criticism was directed toward version 1.0. Now it’s 0.4, so should be fine

Issue #100, opened 2022-12-15
make parents index zero-based


HDembinski:
The parents array in EventData points to the parents of a particle. Currently, we use the original Fortran index (first particle is index 1, etc), but in Python, it is more natural to use zero-based indices, so that once can use parents to directly index the other arrays.

afedynitch:
OK, but how would it work? Can we change default indexing in numpy? I didn't anything on the web that would allow to change the default.

One way would be to decrement parents and daughters by one. Since this is the only place where that would matter for the EventData.

PR #166, opened 2023-07-21
make parents index zero-based

jncots:
This is PR to resolve issue #100

### DIFF
--- a/src/chromo/models/epos.py
+++ b/src/chromo/models/epos.py
@@ -15,9 +15,9 @@ class EPOSEvent(MCEvent):
 
     def __init__(self, generator):
         super().__init__(generator)
-        # EPOS sets parents of beam particles to (-1, -1).
+        # EPOS sets mothers of beam particles to (-1, -1).
         # We change it to (0, 0)
-        self.parents[self.status == 4] = 0
+        self.mothers[self.status == 4] = 0
 
     def _charge_init(self, npart):
         return self._lib.charge_vect(self._lib.hepevt.idhep[:npart])
--- a/src/chromo/models/epos.py
+++ b/src/chromo/models/epos.py
@@ -17,7 +17,7 @@ class EPOSEvent(MCEvent):
         super().__init__(generator)
         # EPOS sets mothers of beam particles to (-1, -1).
         # We change it to (0, 0)
-        self.mothers[self.status == 4] = 0
+        self.mothers[self.status == 4] = -1
 
     def _charge_init(self, npart):
         return self._lib.charge_vect(self._lib.hepevt.idhep[:npart])
--- a/src/chromo/models/epos.py
+++ b/src/chromo/models/epos.py
@@ -13,12 +13,6 @@ from chromo.constants import standard_projectiles
 class EPOSEvent(MCEvent):
     """Wrapper class around EPOS particle stack."""
 
-    def __init__(self, generator):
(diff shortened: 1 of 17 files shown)


jncots:
I changed `parents` -> `mothers` and `children`->`daughters`.

* As I understood from #100, the original arrays can contain `-1`, therefore I restore `-2` which will appear after subtraction by:

```python
# Fix special cases
if mothers is not None:
    mothers[mothers == -2] = -1

if daughters is not None:
    daughters[daughters == -2] = -1
```
I am not sure about this snippet.

* Another issue: `pyhepmc` seems start indexing from 1, so 1 is added for  genevent.from_hepevt

```
parents=ev.mothers + 1 if ev.mothers is not None else None,
children=ev.daughters + 1 if ev.daughters is not None else None,
```

* After change of indexing many tests failed. I fixed most of them, but tests related to Pythia8 still fail.

HDembinski:
The failing tests may come from the rather complex rules how ranges of particle indices can be defined. There are probably some special cases that you need to handle.

* I agree that we don't want -2. One option is
```
parent = np.maximum(-1, parent - 1)
```
Likewise for children. This is fairly elegant, but I am not sure wether this is faster than your option with the mask. If the speed difference does not matter, I would prefer this compact code over masking.

* pyhepmc expecting indices starting with 1, that's a workaround for chromo's behavior. which I can remove then. I will adapt the pyhepmc interface in a separate PR.

jncots:
I tested 2 variants and `parent = np.maximum(-1, parent - 1)` is faster (~3 times) than 2 separate numpy operations. It is, probably, because numpy should not return to python level. So I changed the code to `parent = np.maximum(-1, parent - 1)`.

jncots:
It seems that tests are fixed. Most of the problems were because of `_select_mothers` function. The tests passes, but the function should be checked on whether the initial logic is preserved.

HDembinski:
Ok, I will have a look.

src/chromo/models/epos.py
```diff
@@ -15,9 +15,9 @@ class EPOSEvent(MCEvent):
 
     def __init__(self, generator):
         super().__init__(generator)
-        # EPOS sets parents of beam particles to (-1, -1).
+        # EPOS sets mothers of beam particles to (-1, -1).
         # We change it to (0, 0)
-        self.parents[self.status == 4] = 0
+        self.mothers[self.status == 4] = -1
```
HDembinski, line 20: This line is superfluous now, right? According to the comment?

HDembinski:
> It seems that tests are fixed. Most of the problems were because of _select_mothers function. The tests passes, but the function should be checked on whether the initial logic is preserved.

It looks ok, but rather than checking the implementation manually, we should have tests that cover all the common and special cases.

src/chromo/models/epos.py
jncots: I removed the constructor.

jncots:
`_select_mothers` function seems to be used in many tests. I don't know whether we need additional one.

HDembinski (approved):
Looks good to go! Sometime in the far future we can deprecate mothers and daughters and move back to the PC versions parents and children.

jncots:
> Sometime in the far future we can deprecate mothers and daughters and move back to the PC versions parents and children

Sure, `parents` and `children` are better.

Issue #165, opened 2023-07-19
Roadmap for version 1.0


HDembinski:
Before we release a version 1.0, we need to fix/settle a couple of things that will be breaking changes for some users.

- [x] Unify the event records for all generators. SIBYLL does not include the beam particles. The event record should always include the beam particles. #101
- [x] In the parent and children vector, the indices start with 1. They should start with 0 so that these indices can be directly to fetch other numbers from the EventData arrays px, pid, ... #100
- [x] The QGSJet family does not decay pi0. When Chromo is run with default settings, the final state should be defined identically for all models. For QGSJet, we need to decay the pi0 to make it consistent. #87

Ping @afedynitch @jncots

jncots:
Ok. I will try to fix #100 for a start.

HDembinski:
@jncots Thanks! It is technically easy, but it is a delicate change to the API. We should not just change the indices returned in the children and parent arrays in EventData, because that would silently break everyones code that is doing the subtraction of 1 manually at the moment. I do that in all my scripts, for example.

I propose the following.
- Visually deprecate the current accessors with warnings. numpy has a `VisibleDeprecationWarning` that we can use.
- Add two new accessors which return the index - 1. These need new names. Unfortunately, we used the best names already for the current accessors. We should not just call them parents2 and children2. It is not as political correct, but I propose to use mothers and daughters.
- For a few versions we provide all of them, parents, children, mothers, daugthers.
- Then we eventually remove parents and children.

This how one gracefully changes the API without too much pain for the users.

Since we do not have so many users yet, we could shorten this and just remove parents and children without deprecation. But it is important that code which uses the old accessors with the old indexing convention (parents, children) fails in obvious ways instead of doing something wrong without giving any errors.

PR #167, opened 2023-07-29
Pythia8 afterburner

### LINKED ISSUE #87: What to decay in generators by default: let pi0s decay
The impy default at the moment is to not decay pi0s.

This is not a good default, because it does not match what we understand as final-state particles at colliders. I would like to change this and let pi0s decay by default. Final state particles should be those which have a life-time longer than 30 ps, as described in this public ALICE note https://inspirehep.net/literature/1945324, whose definition agrees with the LHC-wide *CERN Minimum Bias and Underlying Event working group*.

Naturally, it is `MCRun.set_stable` should be supported by all models to change this at run-time for each model as desired by the user.

jncots:
This should fix #87 mentioned in #165 and some other problems releated to issue that some event generators cannot decay some particles.


jncots:
About execution time:

`DecayAfterburner.__call__(event)` typically adds the same execution time as it is required for event generation (~ 1 ms), i.e. the time increases twice.  For DpmJet:

image


37% of the time takes decay itself  (`forceHadronLevel()`, `_decay` on picture), 31 % takes extraction of event stack from Pythia8 (`_extract_result` on picture),  10% takes filling of Pythia8 stack (`_fill`), 6% np.copy and np.where,  15 % copying intial stack to final one.

image

jncots:
Noticed problems:

* DpmjetIII193 produces particles with pid = 99999, Dpmjet306 produces particles with pid = 80000, UrQMD34 produces particles with pid = 0. In case of Dpmjet it is not a problem because status != 1, but UrQMD34 produces pid = 0 with status = 1. However, as charge is returned by Pythia8, it is a problem for afterburner, because Pythia8 doesn't know pid with 0, 99999, or 80000. I changed the `charge_from_pid` to overcome this problem.

* UrQMD34 returns neutron with wrong masses, so afterburner cannot decay them. The same is for QGSJet, but in this case neutrons are returned with proton mass.

src/chromo/decay_afterburner.py
```diff
@@ -0,0 +1,96 @@
...
+
+        # Save not_final status
+        not_final = event.status != 1
+        status = np.copy(event.status[not_final])
+
+        # Particles with negative status are not processed by Pythia
+        event.status[not_final] = -1
+
+        # Save original event parameters which are not passed
+        # to Pythia8 stack
```
afedynitch, line 49: Is this a backup of the event stack? Is it an array copy?

src/chromo/decay_afterburner.py
```diff
@@ -0,0 +1,96 @@
...
+        event.en = pevent.en()
+        event.m = pevent.m()
+        event.vx = pevent.vx()
+        event.vy = pevent.vy()
+        event.vz = pevent.vz()
+        event.vt = pevent.vt()
+        event.mothers = np.maximum(pevent.mothers() - 1, -1)
+        event.daughters = np.maximum(pevent.daughters() - 1, -1)
+
+        # Record back original parameters
```
afedynitch, line 85: "Restore from backup"

src/chromo/decay_afterburner.py
```diff
@@ -0,0 +1,96 @@
...
+        vx = event.vx
+        vy = event.vy
+        vz = event.vz
+        vt = event.vt
+
+        mothers = event.mothers if event.mothers is not None else None
+        daughters = event.daughters if event.daughters is not None else None
+        init_len = len(event.pid)
+
+        # Put on stack
```
afedynitch, line 60: I'm wondering if it is safer to call event.reset() beforehand to always start with a clean stack, instead of the backup/restore.

src/chromo/models/pythia8.py
```diff
@@ -95,15 +95,30 @@ def __init__(self, evt_kin, *, seed=None, config=None):
...
 
+        # Common settings
+        self._config += [
+            # use our random seed
+            "Random:setSeed = on",
+            # Pythia's RANMAR PRNG accepts only seeds smaller than 900_000_000,
+            # this may change in the future if they switch to a different PRNG
+            f"Random:seed = {self.seed % 900_000_000}",
+            # reduce verbosity
+            "Print:quiet = on",
```
afedynitch, line 113: Doesn't this always overwrite, these options even if the user doesn't want the quite?
Also, the TODO comment has been abandoned that was mentioning that we want to use our PRNG in future.

tests/test_decay_afterburner.py
```diff
@@ -0,0 +1,81 @@
...
+
+from chromo.decay_afterburner import DecayAfterburner
+from chromo.util import get_all_models
+from .util import run_in_separate_process
+
+if sys.platform == "win32":
+    pytest.skip("Pythia8 does not run on windows", allow_module_level=True)
+
+
+def run_decay_afterburner(model, evt_kin, stable_particles, decaying_particles):
```
afedynitch, line 14: pretty cool!

tests/test_decay_afterburner.py
```diff
@@ -0,0 +1,81 @@
...
+        assert event_dec == event0
+
+
+@pytest.mark.parametrize("Model", get_all_models())
+def test_decay_afterburner(Model):
+    stable_particles = [111]
+    decaying_particles = [-211, 211, 13, -13, 2112]
+
+    if Model.name in ["UrQMD", "QGSJet"]:
+        # UrQMD produce neutrons = 2112 with wrong masses
```
afedynitch, line 66: Wow! Seriously?? All QGSJet have neutrons with proton mass??

tests/test_pythia8_fill.py
```diff
@@ -0,0 +1,98 @@
...
+        pevent.vy(),
+        pevent.vz(),
+        pevent.vt(),
+        np.maximum(pevent.mothers() - 1, -1),
+        np.maximum(pevent.daughters() - 1, -1),
+    )
+    return res.copy()
+
+
+def append_result(pythia, event):
```
afedynitch, line 50: "xxx_result" seems not the right terms. You could name it more like pushback_event or append_eventstacks, or concat_event_stacks, etc..

tests/test_pythia8_fill.py
```diff
@@ -0,0 +1,98 @@
...
+            event.px[i],
+            event.py[i],
+            event.pz[i],
+            event.en[i],
+            event.m[i],
+        )
+    return get_pythia_event(pythia, event.kin)
+
+
+def fill_result(pythia, event):
```
afedynitch, line 67: here same "result" sounds not exactly what it does, but since it is only for tests, it might be ok.

src/chromo/decay_afterburner.py
```diff
@@ -0,0 +1,96 @@
+import chromo
+import numpy as np
+
+
+class DecayAfterburner:
```
afedynitch, line 5: What happens if this is imported on Windows where Pythia8 is not built?

afedynitch (commented):
Nice work, there are a few minor comments and the code looks good to me.

src/cpp/_pythia8.cpp
```diff
@@ -17,9 +17,12 @@ using namespace pybind11::literals;
...
-    assert(pptr); // never nullptr if charge_from_pid is used on particles produced by Pythia
-    // ParticleData returns partice even if anti-particle pid is used
-    return pid == pptr->id() ? pptr->charge() : -pptr->charge();
+    if (pptr) {
+        // ParticleData returns partice even if anti-particle pid is used
+        return pid == pptr->id() ? pptr->charge() : -pptr->charge();
+    } else {
+        return 0;
+    }    
 }
```
HDembinski, line 26: Please revert. You cannot just return anything when the pid is not known to Pythia. This is hiding that the result is arbitrary. How do you know that charge = 0 is correct? If you don't like the assert, then throw an exception and then handle the exception in the client code which may have more information.

src/chromo/decay_afterburner.py
```diff
@@ -0,0 +1,96 @@
...
+        """Decay particles in `event` which are set as `decaying`.
+
+        Particle is `decaying` if its pdg id is passed
+        in `decaying_pids` list of constructor or it is decaing by default
+        in Pythia8. `stable_pids` of constructor is used to set stable
+        particles.
+
+        After the call `event` contains initial particles + produced particles
+        The decayed particles has event.status == 2.
+        """
```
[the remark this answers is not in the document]
jncots: Ok, commnet are passed through chatGPT

src/chromo/decay_afterburner.py
```diff
@@ -0,0 +1,96 @@
...
+        if len(event.pid) == 0:
+            return
+
+        # Save not_final status
```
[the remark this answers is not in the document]
jncots: Actually, I am not sure about this part. Currently the initial statuses are saved if they are not 1. If the value of initial statuses are not important, then they could be all set to 2. It would be faster and easier because it doesn't require np.copy and np.where in the code

src/chromo/decay_afterburner.py
jncots: Here only refs are saved. No copy.

src/chromo/decay_afterburner.py
jncots: Yes.

src/chromo/decay_afterburner.py
jncots: event.reset() is called by C++ code, so only clean stack is used. I copy the parameters that are not passed to Pythia. The first init_len particles returned by Pythia is the original particles of the event. Pythia assignes some values to vx, vy, vz, vt, which are different from original values. Statuses of decayed particles are set to 2. Here I save the statuses given by original event generator. Charges are saved, because some particles have `pid` that are unknown to Pythia and it will return some default value (777) in this case.

tests/test_decay_afterburner.py
jncots: Yes, QGSJet gives 938.27 GeV. UrQMD produces very random numbers around 900 GeV. I guess it is the reason why Pythia doesn't decay neutrons for QGSJet or UrQMD.

src/chromo/decay_afterburner.py
jncots: It will raise an exception that the library is now found. I haven't checked, but I guess so based on errors when Pythia8 tests are not skipped on CI.

src/cpp/_pythia8.cpp
jncots: Ok, I changed `charge_from_pid` so that it throws an exception.

The problem with exceptions is that the function charge stumbles upon unknown pdg id, while it is needed that it will return the values for known pdgs and inform which items are not known. I was trying to pass that info through exceptions, but pybind11 doesn't allow in a simple way to pass custom exception which contain additional information.

So I added new function `charge_nothrow` which returns 2 arrays: one for charge values and one boolean array whether that value is invalid. It is done for efficiency reasons, as one array with sentinel values (777) is enough to calculate boolean numpy array from it.

HDembinski (commented):
I need more time to review this.

src/cpp/_pythia8.cpp
jncots: Yep, `NaN` is more elegant. I changed the function to return NaN.

jncots:
I started to integrate DecayAutoburner with models. I changed interface to provide more flexible settings for decaying/stable particles.  "theonly_decaying_pids" sets all particles as stable, except the one that in the list. "theonly_stable_pids" sets all particles as unstable, except the one that in the list. The arguments are convenient to set an initial state of stable/unstable in Pythia. "extra_decaying/stable_pids" toggle this state further.

I noticed that no model has expicit lists of stable/unstable particles. Also, there is no common interface to inquire event generators what they consider stable/unstable. Currently, I set:

```python
self._afterburner = DecayAfterburner(
                seed=seed, theonly_stable_pids=long_lived
            )
```
,i.e. all other particles that is not in `long_lived` list will decay by `DecayAfterburner`. 

I guess, it is just enough to keep `final_state_particles` list for each instance and implicitly assume that all other should decay. Then `set_stable` should modify this list. `final_state_particles` could be a property which call `set_stable` if a new list is assigned to it.

The easiest way to switch on `DecayAfterburner` for a model is to provide a common method `switch_on_autoburner`, which will call:

```python
self._afterburner = DecayAfterburner(
                seed=seed, theonly_stable_pids=self.final_state_particles
            )
```

HDembinski:
I am unhappy with the naming.

1) I would prefer a more serious name than "DecayAfterburner". I am open for suggestions. Since Pythia8 is doing the work, we should call it Pythia8Decay?
2) The keyword `theonly_stable_pids` breaks rules regarding underscores. Either `the_only_stable_pids` or just `stable_pids`. Keywords shouldn't be too long and don't have to be self-explanatory. You have to write documentation anyway to explain what the keyword does anyway.

Pythia8Decay is primarily component for us the library developers. It should be used automatically internally by some model that cannot internally decay all particles. It should only be run if the user requests that particles should decay which the model cannot decay by itself. All this should happen without any user input. The user just requests which particles should be stable with `set_stable`.

> I noticed that no model has expicit lists of stable/unstable particles. Also, there is no common interface to inquire event generators what they consider stable/unstable. Currently, I set:

We already have an interface to set particles stable with `set_stable`, but as you say we don't have a way to query which particles are currently set stable in the model.

I like the idea of replacing this interface with a property on the model called `final_state_particles` which returns a list of pids that are considered final state by the model. Assigning a list to final_state_particles should be allowed and should set all those particles stable and decay all others. We can then deprecate `set_stable` which is a bit of a misnomer.

jncots:
I added the functionality discussed above. By default DecayAfterburner is not used. To switch on/off one should explicitly call `switch_afterburner(on=True/False)`.

`_final_state_particles` is the list containing list of the pdg ids which are considered as final. They can be obtained, but not directly assigned. `set_stable` is used to do this. The default value is `long_lived`.

```python
gen = chromo.models.DpmjetIII193(evt_kin)
gen.switch_afterburner()
for event in gen(10):
     ...

# switch off afterburner
gen.switch_afterburner(on=False)
```

jncots:
> I am unhappy with the naming.

Ok, I like `Pythia8DecayHandler`. `stable_pids` is fine.

> It should only be run if the user requests that particles should decay which the model cannot decay by itself. 

Ideally, yes. Is there any good way to get those particles?

> Assigning a list to final_state_particles should be allowed and should set all those particles stable and decay all others.

Basically, it is done in `_update_final_state_particles` but currently it is private method. I find it more covenient to be a method because it allows add and remove particles from the list.

src/chromo/decay_afterburner.py
jncots: It should raise "ModuleNotFoundError".

HDembinski:
> Ideally, yes. Is there any good way to get those particles?

It is model specific. To my knowledge, only QGSJet does not decay anything. I learned from your comments that there are also some issues with DPMJet. Most of the other models do not need this functionality.

I don't like the current design.

* Pythia8DecayHandler should not be called by MCRun. It is not a generic component that all models need. It is a special component that should be used in specific model implementations, which do not implement decays properly, like the class QGSJet in the qgsjet module.
* If you find that there is common code which all such models can share that require external decay handling, then make a new subclass of MCRun called MCRunWithExternalDecay of something, which derives from MCRun. The models like QGSJet then inherit from MCRunWithExternalDecay while the other models that does need this functionality derive from MCRun. This is OOP.
* Please add a property called `final_state_particles` to `MCRun`, which all concrete models implement. The property should return the list of particles that the model includes in the final state. For most models, this is the list of long-lived pids that we have in `constants`. When assigned to, the property will set which particles are included in the final state by the model. This can be easily implemented by calling the hidden model-specific method `_set_stable`.
* Remove `switch_afterburner`. This should not be toggled manually. The model decides on its own if needs the Pythia8DecayHandler based on the requests that the user made to the attribute `final_state_particles`.

>  I find it more covenient to be a method instead of property setter because it allows add and remove particles from the list via second parameter (stable/unstable).

Allowing assignment to `final_state_particles` is pure convenience, but it is Pythonic. We need this propery anyway to have an interface to get the list of particles. It is then natural/pythonic to also support assignment to this property. We don't need another method to add/remove individual particles, we have `set_stable` for that.

src/chromo/common.py
```diff
@@ -708,6 +713,30 @@ def kinematics(self, kin):
         self._kinematics = kin
         self._set_kinematics(kin)
 
+    @property
+    def final_state_particles(self):
+        """Returns sorted list"""
+        uids = np.unique(self._final_state_particles)
+        return list(uids[np.argsort(2 * np.abs(uids) - (uids < 0))])
```
HDembinski, line 720: I said it return a list, but after more thought it should be a tuple instead.

src/chromo/common.py
```diff
@@ -708,6 +713,30 @@ def kinematics(self, kin):
         self._kinematics = kin
         self._set_kinematics(kin)
 
+    @property
+    def final_state_particles(self):
+        """Returns sorted list"""
+        uids = np.unique(self._final_state_particles)
```
HDembinski, line 719: better `_get_final_state_particles`

src/chromo/common.py
```diff
@@ -573,6 +574,8 @@ def __init__(self, seed):
         self._rng = np.random.default_rng(seed)
         if hasattr(self._lib, "npy"):
             self._lib.npy.bitgen = self._rng.bit_generator.ctypes.bit_generator.value
+        self._apply_decay_handler = False
+        self._final_state_particles = np.array([], dtype=np.int64)
```
HDembinski, line 578: This is array should not be there. You are duplicating state that exists in the model. You need to query the model implementation with model specific calles which particles it considers stable. Duplicating state is bad, because it tends to get out of sync.

src/chromo/common.py
```diff
@@ -708,6 +713,30 @@ def kinematics(self, kin):
...
+        else:
+            is_stable = np.logical_not(
+                np.isin(self._final_state_particles, unique_pdgid)
+            )
+            self._final_state_particles = self._final_state_particles[is_stable]
+
+        if self._apply_decay_handler:
+            self._decay_handler.set_stable_decaying(
+                theonly_stable_pids=self._final_state_particles
+            )
```
HDembinski, line 738: I don't see why this new method is needed. You can achieve this with calls to `set_stable`.

src/chromo/common.py
```diff
@@ -721,8 +750,10 @@ def set_stable(self, pdgid, stable=True):
         if abs(pdgid) == 311:
             self._set_stable(130, stable)
             self._set_stable(310, stable)
+            self._update_final_state_particles([130, 310], stable)
```
HDembinski, line 753: Here, that's what I mean, you are duplicating functionality. Remove `_update_final_state_particles` and integrate any code that you need from there into `set_stable`.

src/chromo/decay_handler.py
```diff
@@ -0,0 +1,204 @@
...
+            seed (int): Random seed for Pythia8 initialization.
+            extra_stable_pids (list[int]): List of additional PDG IDs
+            for stable particles.
+            extra_decaying_pids (list[int]): List of additional PDG IDs
+            for decaying particles.
+            stable_pids (list[int]): List of PDG IDs
+            for exclusively stable particles (all other are decaying if they can decay).
+            decaying_pids (list[int]): List of PDG IDs
+            for exclusively decaying particles (all other are stable).
+
```
HDembinski, line 27: This is confusing and way too complicated. I don't understand what you mean by "extra" and "exclusive" decay later. This needs to be simplified.

What does this Handler do? I thought it decays particles which the model does not decay. Do you also want to use it to do the opposite, remove daughters of particles which were decayed in the original model? I think this case never happens, or does it?

I assume that we only need to decay particles which have not been decayed by the model. In that case, you only need one argument in addition to seed, the pids of particles which the handler should decay.

src/chromo/decay_handler.py
```diff
@@ -0,0 +1,204 @@
...
+        # Particles with negative status are not processed by Pythia
+        event.status[not_final] = -1
+
+        # Save refs to original event parameters
+        # that are not passed to Pythia8 stack
+        charge = event.charge
+        vx = event.vx
+        vy = event.vy
+        vz = event.vz
+        vt = event.vt
```
HDembinski, line 162: This is not making copies. Is this code correct?

src/cpp/_pythia8.cpp
```diff
@@ -17,9 +18,13 @@ using namespace pybind11::literals;
...
     auto pptr = pd.findParticle(pid);
-    assert(pptr); // never nullptr if charge_from_pid is used on particles produced by Pythia
+    // return NaN if unknown pid is met
+    if (pptr == nullptr) {
+        return std::numeric_limits<float>::quiet_NaN();
+    } else {
     // ParticleData returns partice even if anti-particle pid is used
-    return pid == pptr->id() ? pptr->charge() : -pptr->charge();
+        return pid == pptr->id() ? pptr->charge() : -pptr->charge();
+    }
```
HDembinski, line 27: Remove the else. The code was already correct before and better style. A function which returns something should always have an unconditional return statement at the end for clarity.

HDembinski (commented):
Still a lot of issues to resolve.

src/chromo/decay_handler.py
HDembinski: Turns out that this was correct, since you are replacing the arrays instead over overriding their contents later.

src/chromo/common.py
jncots: > It is model specific.

I checked all models what they decay and what not. It turns out that only EposLHC decays everything that could be passed to it as unstable particles. Models that use Pythia6 (Sibyll, Sophia) and Pythia6 itself cannot decay neutron. Phojet, Dpmjet - just not decay what set as unstable. UrQMD and QGSJet just don't decay particles.

So, I would say we need DecayHandler for pretty much everything.

src/cpp/_pythia8.cpp
jncots: Done

src/chromo/decay_handler.py
```diff
@@ -0,0 +1,204 @@
...
+        event.m = pevent.m()
+        event.vx = pevent.vx()
+        event.vy = pevent.vy()
+        event.vz = pevent.vz()
+        event.vt = pevent.vt()
+        event.mothers = np.maximum(pevent.mothers() - 1, -1)
+        event.daughters = np.maximum(pevent.daughters() - 1, -1)
+
+        # Restore the original parameters to the event
+        event.status[np.where(not_final)[0]] = status
```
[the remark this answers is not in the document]
jncots: It is possible way but it is ~2x longer than np.where.

src/chromo/decay_handler.py
jncots: > Do you also want to use it to do the opposite, remove daughters of particles which were decayed in the original model? 

I am not sure that I understood this. 

I just wanted to have fine-grained control to what is decaying and what is stable. For current needs it is, perhaps, overkill. In current implementation only stable particles should be set. All other will be decaying.

src/chromo/common.py
jncots: Yeap, it is possible. Removed it.

src/chromo/common.py
jncots: `self._final_state_particles` is set know through `final_state_particles` property and `set_stable` method.

jncots:
I notice one unpleasant side effect of using Pythia8 as DecayHandler: it looks like one can set stable/decay only for the pair of particle-antiparticle. I.e. one cannot set pi+ as stable and pi- as unstable.

src/chromo/constants.py
```diff
@@ -53,32 +55,6 @@
...
-    3222,
-    3312,
-    3322,
-    3334,
-    -321,
-    -211,
-    -2112,
-    -13,
-)
-
```
HDembinski, line 81: You cannot remove this, it is public API. Please revert.

src/chromo/constants.py
```diff
@@ -94,11 +70,6 @@
 # # unflavored particles
 # standard_particles = tuple(standard_particles + [111, 130, 310, 221, 223, 333])
 
-# Default definition of final state particle
-# All particles with proper lifetime shorter than this
-# will decay
-tau_stable = 30e-12  # 30 ps, typical value at LHC
-
```
HDembinski, line 101: You cannot remove it, it is public API. Please revert.

src/chromo/constants.py
```diff
@@ -107,3 +78,42 @@
...
+            stable.append(pid)
+    return stable
+
+
+# Default definition of final state particle
+# All particles with proper lifetime shorter than this
+# will decay
+tau_stable = 30e-12  # 30 ps, typical value at LHC
+# standard long-lived particles with life-time > 30 ps
+long_lived = tuple(long_lived_for(tau_stable))
```
HDembinski, line 119: Remove all this. There is only a countable set of long-lived SM particles, which are the ones in the tuple `long_lived`. If you don't trust that constant, you can add a unit test which compares our list with one computed from `particle`.

src/chromo/decay_handler.py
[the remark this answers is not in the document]
HDembinski: Ok, then at least change to `event.status[*np.where(not_final)] = status`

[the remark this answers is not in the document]
HDembinski: Ok then. It looks to me like abuse of np.where, but I cannot prove it.

src/chromo/decay_handler.py
HDembinski: Then please don't design something that is more general than what we need. This feature already adds so much complexity to our code. We need to try to keep the complexity minimal.

src/chromo/constants.py
HDembinski: I see it was just moved down. Nevermind.

src/chromo/constants.py
HDembinski: I see you replaced this further below with a dynamic computation. I don't see the benefit, because I don't think we actually ever want to change the tau_stable threshold. So there is no need to make this dynamic as you did.

src/chromo/constants.py
HDembinski: I see that you want to enforce that the sequence `long_lived` corresponds to the `tau_stable` value. But we don't actually ever want to change this, so there is no need to make this dynamic.

I don't think that `get_all_decaying_ctau`, `all_decaying_ctau`, and `all_decaying_pids` should be part of the public API. These look like implementation details to me that you need to implement the Pythia8Decay handler. Therefore, you should put these into `decay_handler.py`

HDembinski: If you want to test against `all_decaying_pids`, you probably want this to be a numpy array or a set, depending on the use case, not a list. A search in a list costs linear time, but only logarithmic time in a set.

src/chromo/constants.py
jncots: I reverted it.

src/chromo/constants.py
jncots: I moved `long_lived_for` to util. It is still required to get all decaying particles. All generators has their own default settings. To have full control on decaying/stable particles, all known decaying particles should be set unstable, before setting some subset as stable. 

`long_lived_for` can accept `tau` to return only subset of decaying particles that are stable for `tau` sec/mm.

`_decaying_pids` changed to sets for searching of pids in set_stable.

jncots: One can set new tau condition like this:

```python
generator.final_state_particles = long_lived_for(tau)
```

All unstable particles will be set as stable:

```python
generator.final_state_particles = long_lived_for()
```

All unstable particles will decay if:

```python
generator.final_state_particles = []
```

HDembinski: Ok, if you want to have a function that computes the number of long-lived particles depending on tau, then such a function should live in a module called "constants". It can be a in `util.py`. The list in `constants.py` should refer to the LPCC MB+UE working group's definition of primary particles, which uses tau > 30 ps. ALICE wrote a nice document about this https://inspirehep.net/literature/1945324. It is true that ATLAS considers only particles with tau > 300 ps as primary, but the ATLAS definition makes no sense, since it considers prompt hyperons as non-primary.

I want us to use the LPCC MB+UE working group's definition of primary particles by default. If we then provide a function that computes the list of long-lived particles for a different tau in `util.py` everyone is covered, no?

jncots: So `long_lived_for` is in `util.py`?

src/chromo/models/pythia8.py
HDembinski: If you think it is necessary to give users the freedom to override `Print:quiet = on`, go ahead. I thought it is not necessary. Strictly speaking, we only need to make sure that the handling randomness and initial state (beam particles, frame, etc) cannot be overridden by a user.

src/chromo/constants.py
afedynitch: Agree with 30ps as standard!

A function and some readable docstring would be good. And, there should be a TODO in the source code reminding us to write about how to set stable particles in a proper documentation. Since we don't have any config file that can be browsed by a curious used, so we need to rely more and more on docs due to complexity.

jncots:
`_validate_decay` function is added which checks whether all particles that should decay have been decayed. If not, it uses decay_handler if it is switched on, or suggest to do that if decay_handler is switched off.

src/chromo/models/pythia8.py
jncots: Ok, added possibility to Print:quiet = off

jncots:
> @jncots, could you pull from main before preparing for merge?

Because the next merge is for #170, I've done it there.

afedynitch (commented):
OK. Good to go

afedynitch (approved):
OK. Good to go

afedynitch (approved):
OK. Good to go

afedynitch (approved):
OK. Good to go

afedynitch (approved):
OK. Good to go

afedynitch (approved):
OK. Good to go

afedynitch (approved):
OK. Good to go

afedynitch (approved):
OK. Good to go

src/chromo/util.py
```diff
@@ -687,3 +688,37 @@ def __or__(self, other: Set[PDGID]):
...
+def unique_sorted_pids(ids):
+    """np.arrays of unique ids sorted by abs value
+    with negative value first, e.g.
+    -11, 11, -12, 12, ...
+    """
+    uids = np.unique(np.fromiter(ids, dtype=np.int64))
+    return uids[np.argsort(2 * np.abs(uids) - (uids < 0))]
+
+
+def long_lived_for(tau=0, mm=False):
```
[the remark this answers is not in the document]
jncots: Changed to `select_long_lived`

src/chromo/models/epos.py
```diff
@@ -6,10 +6,64 @@
     fortran_array_insert,
     fortran_array_remove,
     Nuclei,
+    long_lived_for,
 )
 from chromo.constants import standard_projectiles
 
 
+def get_epos_decaying_pids():
```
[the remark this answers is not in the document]
jncots: Done

afedynitch (approved):
Pretty cool. You recovered all of my lost comments!

jncots:
After @afedynitch 's review:

* Renamed `long_lived_for` to `select_long_lived`
* Made lists and function related to unstable particles to be private
* Changed `decaying` to `unstable` where possible
* Change an order in `_set_final_state_particles`
* Add comments to `run_decay_handler`

PR #169, opened 2023-08-09
Sibyll*

afedynitch:
This PR combines multiple items (sorry for that). The main novelty is SIBYLL* as shown at ICRC by F. Riehn:

1. Include SIBYLL* (thx @jncots)
2. When using Ninja for compilation, the FORTRAN files were pre-processed twice, once due to the `-cpp` flag in CMake, and a second time by Ninja itself that calls the preprocessor by default. Fixing this restores the quick build times on machines with many cores.
3. The allowed projectiles for all of the models have been reviews and the lists extended where appropriate
4. Some fixes to UrQMD file list (exclude the includes from f2py)

Opinions desired on:

1. If the SIBYLL* subclasses should be called using the underscore or camel case for consistency?
2. If I should put effort in using only one source file for SIBYLL2.3d and SIBYLL*. In principle, one can restore original SIbyll2.3d behaviour from the sibyll* source file.


jncots:
I think that:

* The name of classes should be changed standard camel style, like `Sibyll23dStarNoEnh`
* There should be only 1 file to avoid code duplication.

src/chromo/models/dpmjetIII.py
```diff
@@ -213,10 +213,11 @@ def _generate(self):
 
 class DpmjetIII191(DpmjetIIIRun):
     _version = "19.1"
+    _projectiles = standard_projectiles | Nuclei() | {3322, 3312, 3222, 3122, 3112, 311}
```
HDembinski, line 216: If it is shared between DpmjetIII191 and  DpmjetIII193, it should probably be in DpmjetIIIRun. This also works, but putting common code there is cleaner.

src/chromo/models/dpmjetIII.py
```diff
@@ -213,10 +213,11 @@ def _generate(self):
...
     _library_name = "_dpmjetIII191"
 
 
-class DpmjetIII193(DpmjetIIIRun):
+class DpmjetIII193(DpmjetIII191):
```
HDembinski, line 220: If you put the extra code into DpmjetIIIRun, you can revert this change.

src/chromo/models/sibyll.py
```diff
@@ -137,34 +141,80 @@ class Sibyll23(SIBYLLRun):
...
+        411,
+        421,
+        4232,
+        431,
+        4122,
+        4132,
+        4232,
+        431,
+        4332,
+    }
```
HDembinski, line 159: Likewise, shouldn't this be in SIBYLLRun?

HDembinski:
`SIBYLL*` is an unfortunate naming choice for us. I agree with @jncots that CAP style for the new classes is the way to be PEP8 compliant, see https://peps.python.org/pep-0008/#class-names

src/chromo/models/sibyll.py
afedynitch: Ok

src/chromo/models/dpmjetIII.py
afedynitch: ok

afedynitch:
@HDembinski I fixed the issues that you pointed out.

Further, I updated Sibyll* to patch release 2 that came in today with a bugfix. After your comments are resolved, I would like to merge this with to_release branch instead of main to bundle the numerous open PRs.

cmake_ext.py
```diff
@@ -53,6 +53,7 @@ def get_models():
     # sib23c02
     # sib23c03
     # dev_dpmjetIII193=/full/path/to/dir/dpmjetIII-19.3
+    # dev_sib23d=/full/path/to/dir/sibyll
```
jncots, line 56: This should be removed because sib23d_start is added.

setup.py
```diff
@@ -17,9 +17,9 @@
         subp.check_call(["git", "submodule", "update"])
 
 
-# for convenience, support building extra models via extra.cfg
-# extra.cfg is not tracked by git, so can be freely modified
-# extra.cfg example:
+# for convenience, support building extra models via model.cfg
```
jncots, line 20: It is `models.cfg`. More complete comment is in `get_models` function. Here it is better to write something like "Refer to the comments for 'get_models' on instructions for adding extra models."

tests/test_cli.py
```diff
@@ -195,7 +196,7 @@ def test_model_3():
         returncode=1,
         stderr=(
             "Error: model=sib is ambiguous, matches (SIBYLL-2.1, SIBYLL-2.3, "
-            "SIBYLL-2.3c, SIBYLL-2.3d)"
+            "SIBYLL-2.3Star-mix, SIBYLL-2.3c, SIBYLL-2.3d)"
```
jncots, line 199: Maybe "SIBYLL-2.3Star-Mix"?

jncots (commented):
It is great to have a new model. I have found only a couple of minor suggestions for improvement.

tests/test_cli.py
afedynitch: This is just the name how it will appear in labels. I prefer the dash here

setup.py
afedynitch: OK, changed

afedynitch:
I implemented the comments by @jncots.

Also, I fixed the cross section readout for nuclei. It was necessary to include more categories into `CrossSectionData` to accommodate the special categories for nuclei, such as "production" and "quasielastic". I updated the tests to accomodate these changes.

src/chromo/common.py
```diff
@@ -36,55 +36,72 @@
...
+        Total cross section (elastic + inelastic).
     inelastic : float
-        Inelastic cross-section. Includes diffractive cross-sections.
+        Inelastic cross section. Includes diffractive cross sections.
     elastic : float
-        Cross-section for pure elastic scattering of incoming particle
+        Cross section for pure elastic scattering of incoming particle
         without any new particle generation.
+    quasielastic : float
+        Nuclear cross section defined as total minus quasielastic.
```
jncots, line 59: It is confusing to name `quasielastic` if it is total-quasielastic. Maybe sort of compl2quasiel - complimentary to quasielastic?

src/chromo/common.py
```diff
@@ -36,55 +36,72 @@
...
         pomeron-pomeron interaction.)
+    diffractive_sum : float
+        Sum of diffractive cross sections.
     """
 
     total: float = np.nan
     inelastic: float = np.nan
     elastic: float = np.nan
+    prod: float = np.nan
+    quasielastic: float = np.nan
```
jncots, line 80: Could you add the description for `prod`?

src/chromo/common.py
```diff
@@ -36,55 +36,72 @@
...
-            - self.diffractive_xx
-            - self.diffractive_axb
+        return self.inelastic - self.diffractive
+
+    @property
+    def diffractive(self):
+        if float(self.diffractive_sum) > 0.0:
+            return self.diffractive_sum
+        return sum(
+            [
```
jncots, line 96: Better to use np.nansum

examples/cross_section.ipynb
jncots, line 1: It would be good to add plt.ylabel("$\sigma$, mbarn")

tests/test_dpmjetIII.py
```diff
@@ -39,3 +50,35 @@ def run_cross_section_ntrials():
...
 def test_cross_section_ntrials():
     run_in_separate_process(run_cross_section_ntrials)
+
+
+def test_cross_section_pp():
+    c = run_in_separate_process(run_cross_section, "p", "p")
+    assert_allclose(c.total, 38.9, atol=0.1)
+    assert_allclose(c.inelastic, 32.1, atol=0.1)
+    assert_allclose(c.elastic, 6.9, atol=0.1)
+    assert_allclose(
```
jncots, line 60: If 38.9, 32.1 and 6.9 are some kind of constants, maybe it is better to give them names like `total_at_10GeV` and so on.

jncots: Or leave a comment that those numbers are for specific kinematics

tests/test_dpmjetIII.py
```diff
@@ -39,3 +50,35 @@ def run_cross_section_ntrials():
...
+
+
+def test_cross_section_pA():
+    c = run_in_separate_process(run_cross_section, "p", "O16")
+    assert_allclose(c.total, 446.1, atol=0.1)
+    assert_allclose(c.inelastic, 328.0, atol=0.1)
+    assert_allclose(c.elastic, 118.1, atol=0.1)
+    assert_allclose(c.prod, 298.7, atol=0.1)
+    assert_allclose(c.quasielastic, 144.6, atol=0.1)
+    assert_allclose(
```
jncots, line 77: The same is here

jncots (commented):
Great job, a lot of changes. Just a few fixes are needed.

src/chromo/common.py
afedynitch: I improved the docstring.

src/chromo/common.py
afedynitch: Added.

src/chromo/common.py
afedynitch: Good idea!

tests/test_dpmjetIII.py
afedynitch: I addeed a comment, but it's just trivial. Simply what DPMJET returns for this calls today and not related to any data reference.

jncots (approved):
Ready for merge

PR #170, opened 2023-08-15
Unified treatment of beam particles

### LINKED ISSUE #101: Unify treatment of beam partices in Event records
We need to unify the Event record produced by different generators a bit, so that downstream code can rely on it. Every record should contain the two beam particles as the first two entries. The status code for a beam particle should be 4. The beam particles should be the root of all particles that are produced in the event.

Why do we need the beam particles in every record at all? Couldn't we just remove the beam particles from all records instead? No, we need the beam particles.

- It allows us to identify diffractive events.
- It allows us to apply a uniform simple cut event[2:] to remove beam particles consistently for all models.
- When we use the CompositeTarget, the beam particles are not the same for all events in a run. Putting the beam particles in the record allows us to identify which ions are colliding in each event.

jncots:
Fixes #101. 

The patch prepend to event 2 beam particles if the generator doesn't contain them, or refill 2 first entry if generator doesn't produce correct one (case of nuclei for DpmJet). These 2 beam particles are calculated from EventKinematics and take into account reference frames of generator and EventKinematics to provide correct px, py, pz, en values. 2 beam particles have status == 4.

The event entries for beam particles are calculated only once for EventKinematics object and then applied for each event generated for this EventKinematics. Therefore, the beam particles event entries are kept in EventKinematics as one of the attribute.

For the particles which have mothers = [-1, -1] and some other similar combinations ([0, -1], [1, -1], [1, 1]), the mothers are reset to [0,1], so all such particles are daughters of beam particles. Phojet193 has mothers of the kind [6, 4]. I don't know what it means, but when such event are stored, pyhepmc produces error. I swapped number to [4, 6] to fix this. Svg graphs after these fixes have a single origin vertex where beam particles meet.


jncots:
It seems that all tests pass. It is a part of #165 issue. Together with #167 it is a part of release #164.

src/chromo/kinematics.py
```diff
@@ -272,3 +302,56 @@ def __init__(self, energy, particle1, particle2):
                 f"{energy!r} is neither a number nor one of "
                 "TotalEnergy, KinEnergy, Momentum"
             )
+
+
+class BeamData:
```
HDembinski, line 307: Make this a dataclass. You get the init then for free, ie. you don't have to write it yourself.

src/chromo/kinematics.py
```diff
@@ -230,11 +230,41 @@ def __init__(
...
+            px=np.zeros((2,), dtype=np.float64),
+            py=np.zeros((2,), dtype=np.float64),
+            pz=np.array([kin.beams[0][2], kin.beams[1][2]]),
+            en=np.array([kin.beams[0][3], kin.beams[1][3]]),
+            m=np.array([kin.m1, kin.m2]),
+            vx=np.zeros((2,), dtype=np.float64),
+            vy=np.zeros((2,), dtype=np.float64),
+            vz=np.zeros((2,), dtype=np.float64),
+            vt=np.zeros((2,), dtype=np.float64),
+            mothers=np.array([[-1, -1], [-1, -1]], dtype=np.int32),
```
HDembinski, line 258: Beam particles never have mothers and you cannot set the daugthers before you know the actual event. Neither attribute should be part of the dataclass.

src/chromo/kinematics.py
HDembinski: Remove attributes mothers, daughters, status. These are all constant values, they should not be settable.

HDembinski: Actually, going further, also vx, vy, vz, vt have predictable constant values. So these do not need to be part of the class either.

src/chromo/common.py
```diff
@@ -543,6 +537,79 @@ def __getnewargs__(self):
         # upon unpickling, create EventData object instead of MCEvent object
         return (EventData,)
 
+    @abstractmethod
+    def _add_init_beam_info(self):
+        pass
```
HDembinski, line 542: Remove the abstractmethod decorator, so that this is the default implementation.

src/chromo/common.py
```diff
@@ -543,6 +537,79 @@ def __getnewargs__(self):
...
+
+        if ibeam.ref_frame != self._generator_frame:
+            # Boost current ref frame (ibeam.ref_frame)
+            # to generators ref frame (self._generator_frame)
+            # Use apply_boost use self.kin.frame
+            self.kin.frame = self._generator_frame
+            self.kin.apply_boost(ibeam, ibeam.ref_frame)
+            self.kin.frame = ibeam.ref_frame
+            ibeam.ref_frame = self._generator_frame
+        return ibeam
```
HDembinski, line 558: This code should live in EventKinematics, not here. You can add a method `_get_beam_data` (ibeam is redundant, there is only one kind of beam) which accepts the generator frame as an argument.

src/chromo/common.py
```diff
@@ -543,6 +537,79 @@ def __getnewargs__(self):
...
+            attr_beam = getattr(ibeam, attr_field)
+            attr_event = getattr(self, attr_field)
+            if attr_event is not None:
+                attr_event[0:2] = attr_beam
+                if attr_field in ["mothers"]:
+                    attr_event[
+                        np.isin(attr_event, [[-1, -1], [0, -1], [1, -1], [1, 1]]).all(
+                            axis=1
+                        )
+                    ] = [0, 1]
```
HDembinski, line 572: What does this do? Please add a comment.

src/chromo/common.py
```diff
@@ -543,6 +537,79 @@ def __getnewargs__(self):
...
+                    attr_event[0:2, :] = [-1, -1]
+
+    def _append_initial_beam(self):
+        ibeam = self._get_ibeam()
+        for attr_field in ibeam._data_attr:
+            attr_beam = getattr(ibeam, attr_field)
+            attr_event = getattr(self, attr_field)
+            if attr_field in ["mothers", "daughters"]:
+                if attr_event is None:
+                    res = None
```
HDembinski, line 600: if attr_event is None, you can continue here, which simplifies the code. It makes no sense to set `res` to `None` and then do setattr below, if the field was already None to begin with.

src/chromo/common.py
```diff
@@ -543,6 +537,79 @@ def __getnewargs__(self):
...
+                    res = np.append(attr_beam, attr_event + 2).reshape((-1, 2))
+                    # Change [1, 1] to [0, 1]
+                    if attr_field == "mothers":
+                        res[np.all(res == [1, 1], axis=1)] = [0, 1]
+                    if attr_field == "daughters":
+                        res[np.all(res == [1, 1], axis=1)] = [-1, -1]
+
+            else:
+                res = np.append(attr_beam, attr_event)
+            setattr(self, attr_field, res)
```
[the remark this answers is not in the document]
HDembinski: Even better, don't use BeamData at all and store this stuff in a dict.

src/chromo/models/pythia6.py
```diff
@@ -13,6 +13,9 @@ def _charge_init(self, npart):
         # TODO accelerate by implementing this loop in Fortran
         return np.fromiter((self._lib.pychge(ki) / 3 for ki in k), np.double)
 
+    def _add_init_beam_info(self):
+        pass
+
```
HDembinski, line 18: Remove this and the other empty implementations to declutter the code after remove the `abstractmethod` decorator in the base class. We only want to make a method abstract if it is absolutely required for every derived class to implement it. If a default implementation exists and can be used for a subset of all derived classes, then we only override it for those where it is necessary and use the (empty) default implementation for the others.

tests/test_to_hepmc3.py
```diff
@@ -28,20 +28,8 @@ def test_to_hepmc3(Model):
...
-        hev = event.to_hepmc3()
-        # only final state is stored
-        fs = event.final_state()
-        assert len(hev.particles) == len(fs)
-        for i, p in enumerate(hev.particles):
-            assert p.pid == fs.pid[i]
-            assert p.status == fs.status[i]
-        assert len(hev.vertices) == 0
-        return  # test ends here
     # special case for Pythia8, which does not contain the parton show
```
HDembinski, line 31: please fix the comment to "shower"

HDembinski (commented):
Thank you for the initiative. This looks good, but this implementation needs some simplifications.

src/chromo/kinematics.py
jncots: Ok. It uses dict now.

src/chromo/common.py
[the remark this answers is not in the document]
jncots: dict.item is used for loop but `getattr(self, field)` is still used, because it is more concise than writing all by hand.

src/chromo/kinematics.py
jncots: Anyway, there should be some value assigned in the end. It is convenient to have beam record that has the same structure as event.

src/chromo/common.py
jncots: The method is moved with some modifications to EventKinematics.

src/chromo/common.py
jncots: It attaches all particles which have beam particles as parents to a single vertex.

src/chromo/common.py
```diff
@@ -543,6 +537,79 @@ def __getnewargs__(self):
...
+                        # Additional condition if that both indeces > 0
+                        condition = (attr_event > 0).all(axis=1)
+                        condition[condition] = (
+                            attr_event[condition, 0] > attr_event[condition, 1]
+                        )
+                        (attr_event[condition, 0], attr_event[condition, 1]) = (
+                            attr_event[condition, 1],
+                            attr_event[condition, 0],
+                        )
+
```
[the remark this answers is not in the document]
jncots: Agree. The specific code moved to Phojet193Event class.

src/chromo/common.py
jncots: Yes, it is better to use `continue`.

src/chromo/models/pythia6.py
jncots: Removed.

jncots:
@HDembinski, thanks for comments, the implementation became much simpler. Now it is consist of one additional method for EventKinematics: `get_beam_data` and 2 methods in EventData. I needed an inverse boost, so I added 1 parameter for `apply_boost` and used SimpleNamespace to have en, pz attributes.

src/chromo/common.py
```diff
@@ -543,6 +537,45 @@ def __getnewargs__(self):
...
+                event_field[0:2] = beam_field
+
+    def _append_initial_beam(self):
+        beam = self.kin._get_beam_data(self._generator_frame)
+        for field, beam_field in beam.items():
+            event_field = getattr(self, field)
+            if field in ["mothers", "daughters"]:
+                if event_field is None:
+                    continue
```
afedynitch, line 566: Shouldn't that be before the if?

src/chromo/models/phojet.py
```diff
@@ -59,6 +62,26 @@ def khard(self):
     #                   self._lib.poevt1.phep[0:4, 5]))
 
 
+class Phojet193Event(PhojetEvent):
+    def _fill_initial_beam(self):
+        super()._fill_initial_beam()
+        # Phojet193 produces mothers with first index > second index
```
afedynitch, line 68: Shouldn't that also be the case for Phojet191 and also the old Phojet?

jncots: Yes, this should be added for other Phojets. I checked once again and found that the problem occurs in Photjet191 and Phojet112 as well.

src/chromo/common.py
jncots: Yes, it is also possible. I moved it up.

src/chromo/common.py
```diff
@@ -543,6 +537,43 @@ def __getnewargs__(self):
         # upon unpickling, create EventData object instead of MCEvent object
         return (EventData,)
 
+    def _add_init_beam_info(self):
+        pass
+
+    def _fill_initial_beam(self):
```
afedynitch, line 543: Replace initial beam?

tests/test_common.py
```diff
@@ -91,13 +91,17 @@ def __init__(self):
...
+            _frame=EventFrame.CENTER_OF_MASS,
         )
 
         super().__init__(generator)
 
     def _charge_init(self, npart):
         return np.zeros(npart)
 
+    def _add_init_beam_info(self):
+        self._append_initial_beam()
```
afedynitch, line 103: There is no test that checks that you have successfully added the particles in the correct positon in the event.

tests/test_to_hepmc3.py
```diff
@@ -63,8 +51,8 @@ def test_to_hepmc3(Model):
         # in case of overlapping ranges of incoming particles
         # the earlier vertex keeps them
         for a, b in unique_vertices:
-            if pa != (a, b) and a <= pa[0] < b:
-                pa = b, pa[1]
+            if pa != (a, b) and a <= pa[0] <= b:
+                pa = b + 1, pa[1]
```
afedynitch, line 55: Where does this change come from?

afedynitch (commented):
I left a bunch of comments and some questions

src/chromo/common.py
```diff
@@ -543,6 +537,43 @@ def __getnewargs__(self):
...
+                    # of to beam particles
+                    event_field[
+                        np.isin(event_field, [[-1, -1], [0, -1], [1, -1], [1, 1]]).all(
+                            axis=1
+                        )
+                    ] = [0, 1]
+
+                event_field[0:2] = beam_field
+
+    def _append_initial_beam(self):
```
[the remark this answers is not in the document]
jncots: Yes prepend is better. Renamed

src/chromo/common.py
jncots: Yes, replace is better. Renamed.

src/chromo/common.py
```diff
@@ -543,6 +537,43 @@ def __getnewargs__(self):
         # upon unpickling, create EventData object instead of MCEvent object
         return (EventData,)
 
+    def _add_init_beam_info(self):
```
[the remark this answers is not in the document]
jncots: The name is changed to `_add_initial_beam`. "init" was for "initial".

src/chromo/kinematics.py
```diff
@@ -95,6 +96,11 @@ def apply_boost(self, event, generator_frame):
             raise NotImplementedError(
                 f"Boosts from {generator_frame} to {self.frame} are not yet supported"
             )
+
+        # Inverse transformation
+        if inverse:
+            bg = -bg
```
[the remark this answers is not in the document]
jncots: If I understand correctly this sign would be used always, but the 'inverse' operation is needed quite rare. So I leave it as it is.

src/chromo/kinematics.py
```diff
@@ -230,11 +236,44 @@ def __init__(
...
+        self.m1 = m1
+        self.m2 = m2
 
         super().__init__(
             frame, part1, part2, ecm, plab, elab, ekin, beams, _gamma_cm, _betagamma_cm
         )
 
+        self._beam_data = None
+
+    def _get_beam_data(self, generator_frame):
```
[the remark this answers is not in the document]
jncots: `_beam_data`  is not a constant of `kinematics` object set once and for all. It depends also on event generator (`generator_frame`) that uses that the `kinematics` object. It could not be a property, i.e. getter without argument.

tests/test_to_hepmc3.py
jncots: This fixes the problem with one of the asserts that compared hempc result with the result of the code that imitates (as I understand) one of the hepmc algorithms. The problem was that before, for example, [1, 5] region with [1, 2] existing region was devided to [1, 2] and [2, 5]. It gave an overlapping ranges. The fix devides [1, 5] to [1,2] and [3, 5] ranges.

src/chromo/common.py
```diff
@@ -543,6 +537,43 @@ def __getnewargs__(self):
...
+        pass
+
+    def _fill_initial_beam(self):
+        beam = self.kin._get_beam_data(self._generator_frame)
+        for field, beam_field in beam.items():
+            event_field = getattr(self, field)
+            if event_field is not None:
+                if field == "mothers":
+                    # all mothers that looks like [-1, -1], [0, -1], [1, -1], [1, 1]
```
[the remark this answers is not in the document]
jncots: It was not correct. Indeed, spectators should only attach to only one of the mothers. However, it turned out that the DpmJet and Phojet history is too complex for hepmc. Because of that the history produced by these models is left as it is. Only indicies in mothers and daughters are shifted to satisfy zero indexing.

src/chromo/models/phojet.py
```diff
@@ -58,6 +61,24 @@ def khard(self):
...
+        super()._fill_initial_beam()
+        # Phojet produces mothers with first index > second index
+        # pyhepmc produces an error when trying to restore such events
+        # from the file recorded by Hepmc writer (writing is OK)
+        # Fix:
+        # Swap if first index is smaller than second, i.e.
+        # if [6, 4], then change to [4, 6]
+        # Additional condition if that both indeces > 0
+
+        mothers = getattr(self, "mothers")
```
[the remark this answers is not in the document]
jncots: No, it wasn't. Now it is changed so that the string's status is changed to positive before shifting mothers to  -1.

src/chromo/models/phojet.py
```diff
@@ -47,6 +47,9 @@ def khard(self):
         """Total number of realized hard cuts"""
         return self._lib.podebg.khard
 
+    def _add_init_beam_info(self):
+        self._fill_initial_beam()
```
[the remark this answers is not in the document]
jncots: Ok. Phojet's beam is left as it is, except status is changed to 4. Dpmjet's beam is fixed to have a correct `pid`, `status`, `en`, `pz`.

tests/test_common.py
jncots: Ok. The `test_models_beam` is added. It tests beam values with first 2 entries of event.

jncots:
There have been introduced couple of changes.

* Because models are more different that it was assumed, it is more convenient to introduce  `_history_zero_indexing` and `_repair_initial_beam` for MCEvent and override them in each model separately.
* Sophia history is added
* Phojet and Dpmjet have a complex history that hepmc cannot understand, therefore `to_hepmc3` was overriden to output only final particles attached to beam particles. The original history (mothers and daughters) is left unchanged except shift to -1 to follow zero-indexing.
* Epos-LHC history changes when target is nuclei. It is only possible to add target ahead of the stack to be valid history for hepmc. Not sure if this is correct.
* Test for writing and reading hepmc format is added to tests correctness of event history.
* Test for beam particles is added. Epos-LHC is checked only for pp because pO switches beam particles.

jncots:
I noticed 2 following problems with Phojet and Dpmjet.

* In Phojet all decayed particles disappear from history, i.e. only gammas appear and not pi0, and so on. It is a problem of Phojet, because the printout of original stack has the same problem
* In DpmJet only final particles are boosted to lab frame. I found the place where it could be changed:

```fortran
# in SUBROUTINE DT_LT2LAB

#This

      DO i = NPOint(4) , NHKk
         IF ( (ABS(ISThkk(i)).EQ.1) .OR. (ISThkk(i).EQ.1000) .OR. 
     &        (ISThkk(i).EQ.1001) ) THEN
            CALL DT_LTNUC(PHKk(3,i),PHKk(4,i),pz,pe,-3)
            PHKk(3,i) = pz
            PHKk(4,i) = pe
         END IF
      END DO

# should be changed to

      DO i = 1 , NHKk
            CALL DT_LTNUC(PHKk(3,i),PHKk(4,i),pz,pe,-3)
            PHKk(3,i) = pz
            PHKk(4,i) = pe
      END DO
```

@afedynitch, could you change it?

afedynitch:
Cool! Did you try the proposed source code? It may need to be behind a flag, maybe. Need to have a deeper look

jncots:
> Cool! Did you try the proposed source code? It may need to be behind a flag, maybe. Need to have a deeper look

Yes, I tried it. 

I also tried to fix the energy using only chromo:

```python
self._lib.dtflg1.iframe = 2
self._frame = EventFrame.CENTER_OF_MASS
```
but I've seen events where the energy of decayed particle (rho+ if I remember it correctly) is larger than the sum of its products.

src/chromo/common.py
```diff
@@ -418,6 +418,12 @@ def to_hepmc3(self, genevent=None):
         """
         import pyhepmc  # delay import
 
+        if not pyhepmc.__version__ >= "2.13.2":
```
afedynitch, line 421: Is this right way to compare versions?
There is a typo in pyhepm.

src/chromo/common.py
```diff
@@ -438,14 +444,19 @@ def to_hepmc3(self, genevent=None):
             ev = self[mask]
             mask = (ev.mothers[:, 0] == 0) | (ev.mothers[:, 0] == 1)
             ev.mothers[mask] = (0, 1)
-        elif model in ("UrQMD", "PhoJet", "DPMJET-III"):
-            # can only save final state until history is fixed
-            warnings.warn(
-                f"{model}-{version}: only final state particles "
-                "available in HepMC3 event",
-                RuntimeWarning,
+        elif model == "PhoJet":
```
afedynitch, line 447: The specific of the models should be handled by the model-related classes and not in common. Common should remain model agnostic as much as possible.

src/chromo/kinematics.py
```diff
@@ -230,11 +236,44 @@ def __init__(
 
         _gamma_cm = (elab + m2) / ecm
         _betagamma_cm = plab / ecm
+        self.m1 = m1
```
afedynitch, line 239: Add a comment above as doc, that that may introduce some E/p conservation issue if the internal mass in the model is too different to that from `particle`/PDG/Pythia.

src/chromo/models/dpmjetIII.py
```diff
@@ -225,6 +236,11 @@ def _generate(self):
         self._lib.dtevno.nevent += 1
         return not reject
 
+    def print_original_event(self, mode=1):
+        if hasattr(self._lib, "dtflka"):
+            self._lib.dtflka.lpri = 5
```
afedynitch, line 241: Bug.  Lpri state should be saved and then restored after the fit.

src/chromo/models/epos.py
```diff
@@ -23,6 +23,50 @@ def _get_impact_parameter(self):
...
+        # Set [i, i] to [i, -1]
+        condition = self.mothers[:, 0] == self.mothers[:, 1]
+        self.mothers[condition, 1] = -1
+
+        self.daughters = self.daughters - 1
+        condition = self.daughters[:, 0] == self.daughters[:, 1]
+        self.daughters[condition, 1] = -1
+
+    def _repair_initial_beam(self):
+        # projectile and target are switched!!!
```
afedynitch, line 44: The order of projectile and target can not be switched, since soem configurations are asymmetric. Something else needs to be done here.

src/chromo/models/phojet.py
```diff
@@ -196,6 +206,11 @@ def _set_kinematics(self, k):
     def _generate(self):
         return not self._lib.pho_event(1, self.p1, self.p2)[1]
 
+    def print_original_event(self, mode=2):
+        if hasattr(self._lib, "poinou"):
+            self._lib.poinou.lpri = 5
```
afedynitch, line 211: same as in dpmjet

tests/test_to_hepmc3.py
```diff
@@ -24,24 +27,19 @@ def run(Model):
 
 @pytest.mark.parametrize("Model", models)
 def test_to_hepmc3(Model):
-    if Model == im.UrQMD34:
-        pytest.xfail("UrQMD34 FAILS, should be FIXED!!!")
+    if Model.name in ("PhoJet", "DPMJET-III"):
```
afedynitch, line 30: Is this really the case? Pythia8 should also have an altered event stack.

afedynitch (commented):
Great work! Some items require some fixing but it's almost good to go.

src/chromo/common.py
jncots: Changed it to a more safe way with `packaging.version`

src/chromo/common.py
jncots: Move the code to the corresponding classes

src/chromo/models/dpmjetIII.py
jncots: Fixed

tests/test_to_hepmc3.py
jncots: Below, there were a Pythia8 specific code, therefore the test passed. The code in the test has been changed to use hepmc fix.

src/chromo/kinematics.py
jncots: The comment is added to the code of the beam definition.

src/chromo/models/epos.py
jncots: The code is changed to take into account that both projectile and target could be nucleus. No swaping of projectile and target occurs now. The problem why the swap was introduced is fixed following way.

The problem occured when some particles were attached to the projectile and several wounded nucleons. When target nucleous inserted between them, it created a configuration which hepmc cannot process. I attach all such wounded nucleons only to projectile, whenever such situation occurs.

jncots:
The issues noticed by @afedynitch have been resolved. 

Problems with Phojet (removed history) and Dpmjet (Lorentz boost only for final particles) should be resolved.

One strage behaviour of EposLHC is noticed in connection with `test_final_state`: it produce different particles from run to run for the same seed. Because of that it is not possible to define `known_issues` to pass the test.

afedynitch:
> EposLHC is noticed in connection with `test_final_state`: it produce different particles from run to run for the same seed.

Does it mean that it doesn't conserve the random number sequence? It would be strange because it runs in CORSIKA 7.

HDembinski:
> Does it mean that it doesn't conserve the random number sequence? It would be strange because it runs in CORSIKA 7.

I worked on the rng instrumentation in EPOS-LHC. It is confusing because they use several seeds. Maybe some of them are not properly initialized.

afedynitch (approved):
OK, good to go but instead `print_original_event`, `print_native_event`.

Otherwise great long work. Cheers

src/chromo/common.py
```diff
@@ -403,6 +404,13 @@ def xlab(self):
     #     """I don't remember what this was for..."""
     #     return self.en / self.kin.pcm
 
+    def _repair_for_hepmc(self):
```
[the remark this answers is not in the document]
jncots: Changed

PR #174, opened 2023-09-11
Adding Pythia 8.3.10 to chromo

gaudu:
Attempt to solve https://github.com/impy-project/chromo/issues/171 with @HDembinski ⚡


afedynitch:
Thanks for starting this! What about this cached directory? How big are the files? Does it cleanup/update itself?

gaudu:
> Thanks for starting this! What about this cached directory? How big are the files? Does it cleanup/update itself?

For each (projectile, target, energy) combination, you will produce three files:
- <filename>.sigfit is no bigger than 50 bytes
- <filename>.mpi and <filename>.sasd.mpi increase in size while increasing energy and atomic number of either projectile or target, as would the overall number of interactions from the initial collision. The sizes of the .mpi files, from my personal experience, do not exceed ~150 kB, while the .sasd.mpi files are usually less than ~100 kB.

For example, a proton boosted against a target of oxygen-16 at rest would produce files from the following sizes:
| | pO @ 10^3 GeV| pO @ 10^7 GeV | pO @ 10^11 GeV |
|--------|--------|--------| --------|
| <filename>.sigfit | 45 bytes | 46 bytes | 42 bytes |
| <filename>.mpi | 26 kB | 104 kB | 122 kB |
| <filename>.sasd.mpi | 18 kB | 70 kB | 80 kB |

There is currently nothing in place for the files to cleanup or update itself, since they are supposed to be generated once by the user and kept for future runs using the same (projectile, target, energy) combination. I would need to check if the files are generated in the working directory of the user or in the /examples folder of their Pythia8 installation.

HDembinski:
> Thanks for starting this! What about this cached directory? How big are the files? Does it cleanup/update itself?

We also don't have cleanup routines for the other cached files in place. For example if you update chromo and it needs to download a newer zip file from our server, the old one is not deleted, as far as I am aware.

We should probably wipe the iamdata directory everytime a download of the zip file is triggered, which would do the correct cleanup automatically.

HDembinski:
@afedynitch @jncots This PR is failing on CI because we are don't have the new zip file for this Pythia8 version uploaded yet. Is it possible to add a new file without making a new release?

afedynitch:
> Is it possible to add a new file without making a new release?

Yes, it's possible. Please attach the file here if it's not too big

afedynitch:
> a proton boosted against a target of oxygen-16 at rest would produce files from the following sizes

The files are so small that I would rather prefer precomputing all of the common combinations and distribute these as binaries using our zip downloads. Otherwise testing on public workers may take too long. What do you think about this @gaudu & @HDembinski ?

gaudu:
> The files are so small that I would rather prefer precomputing all of the common combinations and distribute these as binaries using our zip downloads. 

I would not mind generating few files if needed, it really never took that long for me. Could you provide the "common" (projectile, target, center-of-mass energy) combinations you would like to have run?

HDembinski:
> > > Is it possible to add a new file without making a new release?
> > 
> > 
> > Yes, it's possible. @gaudu, please attach the file here if it's not too big
> 
>     1. Here is the zip file, including the pdfdata and xmldoc of Pythia 8.3.10 (size of 13,2 MB which I hope is not too big for this post).
>        [Pythia8_v003.zip](https://github.com/impy-project/chromo/files/12588226/Pythia8_v003.zip)
> 
>     2. One also needs to update the `Pythia8_v003` text file which will appear in /chromo/src/chromo/iamdata/Pythia8 with the link to the .zip file before merging again!

@afedynitch @jncots You can ignore this, I independently created the zip and uploaded it to the right place.

HDembinski:
> > The files are so small that I would rather prefer precomputing all of the common combinations and distribute these as binaries using our zip downloads.
> 
> I would not mind generating few files if needed, it really never took that long for me. @afedynitch Could you provide the "common" (projectile, target, center-of-mass energy) combinations you would like to have run?

@afedynitch We can add some precomputed file to the Pythia8 zip file, for example those used in the unit tests, so that the unit tests run faster. However, we cannot cover all use cases of everyone, because we cannot predict how people use Chromo. Therefore, it is anyway needed to generate these cache files on demand. I see no issue with that and it is quite nice to have this functionality in place, as it speeds up Pythia8 a lot when you use it the second time for the same inputs. In practice that happens a lot.

HDembinski:
Good news, the tests are all green, except an unrelated Sibyll23StarMixed failure.
```
FAILED tests/test_final_state.py::test_generator[Sibyll23StarMixed] - AssertionError: Sibyll23StarMixed zero counts for Omega~+, Xi~0, Xi~+, Xi-, Xi0, Omega-2112 27491 False, 130 27728 False, 211 244736 False, 310 27554 False, 321 28965 False, -3334 0 False, -3322 0 False, -3312 0 False, -3222 1795 False, -3122 5876 False, -3112 1762 False, 3112 1725 False, 3122 8684 False, 3222 3300 False, 3312 0 False, 3322 0 False, 3334 0 False, -321 26585 False, -211 233154 False, -2112 16256 False__: [ True  True  True  True  True False False False  True  True  True  True
```
This test also fails locally. @afedynitch @jncots How could it happen that this test was committed? We should never merge PRs with failing tests.

HDembinski:
For the record, locally on my computer also another test fails in addition to the Sibyll23StarMixed one:
```
FAILED tests/test_final_state.py::test_generator[Sibyll23StarMixed] - AssertionError: Sibyll23StarMixed zero counts for Omega~+, Xi~0, Xi~+, Xi-, Xi0, Omega-2112 27491 False, ...
FAILED tests/test_final_state.py::test_generator[UrQMD34] - AssertionError: UrQMD34 zero counts for 2112 1705 False, 130 1812 False, 211 15775 False, 310 1756 False,...
```

afedynitch:
> How could it happen that this test was committed? We should never merge PRs with failing tests.

The PR testing this was merged into main before SIBYLL* was merged, or the other way around. We fixed it in #170 . @jncots  is about to merge it into main

afedynitch:
Hi @gaudu,

could you merge from main and see if the tests pass? We merged all of the features we wanted. 
Then we go through a more thorough review of your PR.

gaudu:
> Hi @gaudu,
> 
> could you merge from main and see if the tests pass? We merged all of the features we wanted. Then we go through a more thorough review of your PR.

Hi @afedynitch,

I am still pretty new to git, just to make sure I understood you correctly. Does this mean I should merge my working branch (gaudu/chromo/pythia8310) into my main branch (gaudu/chromo/main) to see if the tests pass, and then you will go through the review of this PR?

afedynitch:
Ah, I see.

please merge our main into the main of your fork. Then merge your main into your branch. This PR should then update automatically.

Probably, you can merge our main directly into your branch but I remember having some weird problems doing this. 

If you succeed, then the conflicts message above should disappear.

afedynitch:
Oh, once you’ve done it, ‘install -e .’ locally and run ‘pytest -vv’. It should be better to run pytest locally after merging into your main to make sure that the errors you encounter later are not from your configuration.

src/chromo/models/pythia8.py
```diff
@@ -161,28 +173,61 @@ def _cross_section(self, kin=None):
...
+
     def _set_kinematics(self, kin):
         config = self._config[:]
 
+        # Path prefix for some cache files that Pythia8 generates
+        # to speed up subsequent runs
+        cache_prefix = str(
+            _cache_base_dir()
+            / "Pythia8"
+            / f"cache_{int(kin.p1)}_{int(kin.p2)}_{kin.ecm / MeV:.0f}".replace("-", "m")
```
afedynitch, line 201: what about the `+` or other characters?

One problem I can see is that a file is created for each center of mass energy. How is this going to behave if we use this in a cascade MC (what we’re doing elsewhere)? It means that an infinite number of files will be created.

I heard that Pythia8 has been made compatible somehow with CORSIKA 8? Then there must be some sort of way to get energy dependent tables. Caching projectile/target combinations is ok though since it must be limited.

Also…. why MeV?

src/chromo/models/pythia8.py
```diff
@@ -161,28 +173,61 @@ def _cross_section(self, kin=None):
...
+        )
+
         # TODO use numpy PRNG instead of Pythia's
         config += [
             # use center-of-mass frame
             "Beams:frameType = 1",
             # do not print progress
             "Next:numberCount = 0",
+            "MultipartonInteractions:reuseInit = 3",
+            f"MultipartonInteractions:initFile = {cache_prefix}.mpi",
```
afedynitch, line 211: Is this globally necessary or recommended for all pythia8 users or just nuclei? If this is only relevant for users of nuclei then it needs to be further below.

src/chromo/models/pythia8.py
```diff
@@ -161,28 +173,61 @@ def _cross_section(self, kin=None):
...
-            )
+        # TODO Pythia8 likes to say this:
+        #  To avoid refitting, add the following lines to your configuration file:
+        #     HeavyIon:SigFitNGen = 0
+        #     HeavyIon:SigFitDefPar = 18.39,1.91,0.36
+        # but these numbers of collision-specific, so we cannot just set this globally.
+        # We could generate collision-specific cache files like we do for the
+        #  MPI stuff and set the numbers via those. To get the numbers, call
+        # SubCollisionModel::getParm(), which is complicated to call from the Pythia
+        # object, this is work in progress.
```
afedynitch, line 222: I’ve seen some draft in your edited pybind interface? Do you know what exactly is the problem? Is the function exposed in the original pythia8 python interface?

src/chromo/util.py
```diff
@@ -396,6 +398,13 @@ def _download_file(outfile, url):
         raise ConnectionError(f"{fname} has not been downloaded")
 
 
+# Path to directory to store precomputed data tables
+def _cache_base_dir():
```
afedynitch, line 402: No docstring :)

src/cpp/_pythia8.cpp
```diff
@@ -244,11 +245,18 @@ PYBIND11_MODULE(_pythia8, m)
 
         ;
 
+    py::class_<SubCollisionModel, std::shared_ptr<SubCollisionModel>>(m, "SubCollisionModel")
+        // TODO add getParm
```
afedynitch, line 249: Which function exactly do you want to implement?

@jncots @HDembinski , this could be a good chance to also satisfy this feature request #187 here as well?

afedynitch (commented):
Great. Thanks for the progress. I left a couple of questions about things I’d like to understand better.

src/chromo/models/pythia8.py
gaudu: The initialization of the MPI framework is the most time-consuming part of the whole initialization step, when simulating collisions for heavy ion systems with Angantyr. It is a feature for the users to gain time if they run the same (projectile, target, eCM) combination for more than one time.

gaudu: If you want to look up the details of this MPI initialization, here is the documentation: https://pythia.org/latest-manual/MultipartonInteractions.html (scroll down to "Initialization speed-up").

HDembinski: Short summary: yes, it is useful for everyone and therefore this placement is ok.

src/chromo/util.py
HDembinski: If you insist on a docstring, we can of course add one, but I usually don't add docstrings for private functions (starting with underscore), because these are not to be called by users and therefore don't need user-friendly documentation. Linters should not complain about those.

src/cpp/_pythia8.cpp
HDembinski: I would do this in a separate PR, but yeah, it is probably not complicated to implement access to the weights in a similar way.

I don't like that an info object has to be created each time when these are accessed, this seems wasteful and unnecessary. There may be a way around that if we access private members. I would like to take a look before we implement something more straight-forward but inefficient.

src/chromo/models/pythia8.py
afedynitch: Thanks for the explanations. @gaudu so you mean that the `allowVariableEnergy` is broken but will be fixed in November? Would that be something we could use permanently in Chromo? We also are in November... so you'd prefer not to wait for this latest version?

For now, I think it is ok to generate infinite files. However, I would like to ask you to throw warnings if the number if files in the cache folder exceeds 10k or so. We already have at least one application that changes the "beam" configuration in every event and it would strain the file system immediately.

src/chromo/models/pythia8.py
afedynitch: OK, so I read through the docs and I disagree with that it is safe to use this globally. Changes to any of the parameters except a few would collide with that file. Also there are clear limitations when using variable beam energy etc. such as the `Beams:allowIDAswitch`. Catching all of the potential issues and misuse is *not* the task of `chromo` and has to be redirected to the user, and the Pythia docs.

Therefore, I favor that Pythia8 is initialized with its default values. The version with accelerated initialization should be either explained in an example `example_speedup_pythia8_initialization.ipynb` or in a separate subclass `Pythia8Fast` or so.

src/chromo/util.py
afedynitch: It's ok :)

src/chromo/models/pythia8.py
HDembinski: Ok, we will implement a warning.

I would prefer to merge now and then switch to the other solution later. We have another patch of the cross-sections planned that is urgent fuer Chloe, which is based on this PR. It is a small patch to remove the cache file feature once a better solution is implemented in Pythia8 and works stable.

afedynitch: Why can't she subclass for the time being? I have concerns about caching globally. If I want to preselect and xB or pT cut then the cache files are not valid. Or if a process is switched off, then the files are invalid. Pythia doesn't seem to check for the consistency.

From Pythia8 docs:
>  This can save a significant time, notably for short test runs. A warning is that all relevant settings must be the same in the two runs. It is up to the user to ensure that this is the case; it would be next-to-impossible to catch any possible conflict.

gaudu: @afedynitch This feature (in a bugged state or not) will be a later addition to our toolkit, or even better within the Pythia toolkit. I believe it should not be a reason to further delay this merge request. 

As Hans brought it up, there is further work to be done to bring to the heavy ions runs in chromo correct cross-sections from a Physics point-of-view. I am hoping to deal with this in another PR.

afedynitch: Yes, but what is the reason to merge an unsafe PR? In `chromo`, all models run in default config, or as close as it can get. Any user can tinkle with whatever is exposed. But it is against the philosophy of such a generic code that targets many different user.

Also I don't understand the rationale behind the caching by default. If you need to create multiple pythia instances within the same interpreter session, why not using a tmp folder? Also, I don't get why this is such a critical feature. A user, producing 10-50k events for a single energy in pp can wait a few seconds to generate fresh tables.

For heavy ions, this might be different but since the heavy ion stuff will get bugfixed in the next version of Pythia, I strongly suggest to subclass for the time being and implement the latest version with some more generic treatment of the nuclear cases.

HDembinski: Using the cache is not a dangerous feature and since it is a performance option, it is not going against your wish to run with defaults as much as possible. It is not an unsafe PR in my view.

src/chromo/models/pythia8.py
HDembinski: I forgot to answer this. This is the issue: It is not clear to me how to get the parameters 
`HeavyIon:SigFitDefPar = 18.39,1.91,0.36`
etc in an automatic way directly from Pythia. I don't want to scrape the text output, that's brittle and ugly. This is something to handle in a follow-up PR.

src/chromo/models/pythia8.py
afedynitch: @HDembinski, if you change pT0, or the tune - this is something Pythia users do quite often - then the presence of an old cache file will generate undefined behavior because the cross section table doesn't match. The cache is transparent to the user (since its hidden in a "system" folder and there is no tracking of which option combination has been used to generate it. 

Artificial but not unrealistic example: I download chromo and run Pythia8 with my favorite pT0 setting of 2.567 GeV at a cms energy that's used in one of the tests. Then running the tests afterwards will likely fail because it will pick up a cache file generated with altered settings but apply it to a default run.

It is strange that you are in favor such an implementation.

I'm not against merging the more recent Pythia version with all the downloads and files etc, but I'm against merging the flawed cache implementation and the non-default default settings.

HDembinski:
@gaudu @afedynitch I am trying to summarise the todolist for accepting this patch. Is this ok?

- [ ] Make caching optional via setting in Pythia8 init
- [ ] Generate hash from the config settings of Pythia8 and generate the filename of the cache files from this hash. This ensure that caches are not falsly used when the user changes physics parameters in Pythia8
- [ ] Generate a warning when too many cache files are generated (more than 1000?) and stop to generate new cache files then.

src/chromo/models/pythia8.py
```diff
@@ -161,28 +181,60 @@ def _cross_section(self, kin=None):
...
+            / f"cache_{int(kin.p1)}_{int(kin.p2)}_{kin.ecm / MeV:.0f}".replace("-", "m")
+        )
+
         # TODO use numpy PRNG instead of Pythia's
         config += [
             # use center-of-mass frame
             "Beams:frameType = 1",
             # do not print progress
             "Next:numberCount = 0",
+            "MultipartonInteractions:reuseInit = 0",
```
afedynitch, line 216: Doesn't this disable the cache permanently for this instance of the Pythia8 class? There is nothing to reset it back to the value 3

afedynitch (commented):
The PR is much improved and I'm happy with the featureset. The tests requrie a bit of work and still contain unfinished TODOs.

src/chromo/models/pythia8.py
```diff
@@ -161,28 +181,60 @@ def _cross_section(self, kin=None):
...
+        #  MPI stuff and set the numbers via those. To get the numbers, call
+        # SubCollisionModel::getParm(), which is complicated to call from the Pythia
+        # object, this is work in progress.
 
-            # speed-up initialization by not fitting nuclear cross-section
-            config.append("HeavyIon:SigFitNGen = 0")
-            config.append("HeavyIon:SigFitDefPar = 10.79,1.75,0.30,0.0,0.0,0.0,0.0,0.0")
+        if (kin.p1.A or 0) > 1 or (kin.p2.A or 1) > 1:
+            config += [
+                # similar to MultipartonInteractions:reuseInit
```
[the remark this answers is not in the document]
gaudu: It was a mistake on my side. I have made few changes since and will push them soon!

src/chromo/models/pythia8.py
gaudu: It was a mistake on my side, sorry!

src/chromo/models/pythia8.py
```diff
@@ -161,28 +181,60 @@ def _cross_section(self, kin=None):
             diffractive_axb=st.sigmaAXB,
         )
 
+        # TODO use pythia.info.sigmaGen(i) for cross-sections including nucleus in the
```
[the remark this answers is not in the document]
gaudu: I can try to rephrase it here, or would you rather I try to modify the TODO itself in the code?

src/chromo/models/pythia8.py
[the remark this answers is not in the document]
gaudu: You can find more information about this topic following this website (https://pythia.org/latest-manual/HeavyIons.html) and scrolling to the section for "Subcollision models".

It seems the default Pythia 8 settings, for the SubCollisionModel in Angantyr,  is `"HeavyIon:SigFitDefPar = 17.24,2.15,0.33,0.0,0.0,0.0,0.0,0.0"`. While the one, you highlighted in the code, might have been computed for another (heavy ion) run that I showed to Hans during a previous call (so most likely for negative pion boosted for 158 GeV/c onto a fixed carbon target using Angantyr).

gaudu:
I made few modifications in pythia8.py and test_pythia8.py recently which were unsuccessful at passing the macos and unbuntu tests at the step `pythin -m pytest -vv -n 3` which is most likely pointing at failing one of the new test I implemented. I am trying to understand what I did that could have this outcome at the moment.

On another note, I would like to discuss where exactly we want to have/not have the cache files for (a) the tests, (b) the users? Mostly to not how to answer to @afedynitch comments:
> Can the base directory be an optional parameter? If that would be possible then the cache files could be stored as part of the test suite and then nuclei could be tested much faster.

> Regarding the caching request further up...I think that this test should be the only one generating and using the cache file. All other tests involving nuclei, should come with caches distributed.
These cache files should be in tests/Pythia8_caches/...

HDembinski:
I have been silent here for a while, but I kinda like the idea of making the cache directory a parameter, since we anyway want to allow users to turn this feature off and instead of passing a boolean we might as well pass a string (the path) or `None` to deactivate the caching feature.

On the other hand, I don't see a fundamental difference between these cache files that we generate ourselves and the hardwired ones that we distributed in `iamdata`. So for testing, I would add the cache files needed for tests to iamdata and make it part of the data that is automatically downloaded from github.

tests/test_pythia8.py
```diff
@@ -177,3 +185,22 @@ def test_gp():
...
+    # filename for the cache files.
+
+
+def test_cache_file_heavy_ion():
+    evt_kin = CenterOfMass(2000 * GeV, "p", "He")
+    m = Pythia8(evt_kin, seed=1, cache=True)  # False
+    config_lines = "\n".join(m._config)
+    assert "HeavyIon:SasdMpiReuseInit = 3" in config_lines  # 0
+    assert "HeavyIon:SigFitReuseInit = 3" in config_lines  # 0
+    # TODO Test for the cache file production for pHe@2TeV when we have agreed on the
```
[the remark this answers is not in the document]
HDembinski: I have a slightly different opinion expressed in a comment below, I think the place for the natural place for these hard wired cache files is still in `iamdata`

afedynitch:
I was not aware of the distribution of some default cache files within the binary package... in that case I agree that the natural place for these files to live is within the iamdata directory.

tests/test_pythia8.py
```diff
@@ -98,18 +99,25 @@ def test_mothers(event):
...
 
 
-@pytest.mark.skip(reason="Simulating nuclei in Pythia8 is very time-consuming")
-def test_nuclear_collision():
-    # The test takes ages because the initialization is extremely long,
-    # and Pythia seldom raises the success flag unless Ecm > TeV are used.
-
-    event = run_collision(2000 * GeV, "p", (4, 2))
-    assert event.pid[0] == 2212
+@pytest.mark.parametrize("projectile", ("pi-", "K+", "p", "He"))
```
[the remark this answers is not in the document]
afedynitch: Yes. I meant that this test should use the cache files supplied with the binary data package in iamdata as @HDembinski suggested.

tests/test_pythia8.py
```diff
@@ -177,3 +186,32 @@ def test_gp():
...
+    evt_kin = CenterOfMass(10 * GeV, "p", "p")
+    m = Pythia8(evt_kin, seed=1, cache=True)
+
+    config_lines = "\n".join(m._config)
+    assert "MultipartonInteractions:reuseInit = 3" in config_lines
+
+    cache_file_path = os.path.join(
+        _cache_base_dir(), "Pythia8", "cache_2212_2212_10000.mpi"
+    )
+    assert os.path.exists(cache_file_path)
```
afedynitch, line 201: If a set of default cache files are provided in iamdata, this test would just check if these files are there or not. This has to be made sure that this test generates something that's not in the default files.

tests/test_pythia8.py
```diff
@@ -177,3 +186,32 @@ def test_gp():
...
+def test_cache_file_heavy_ion():
+    evt_kin = CenterOfMass(2000 * GeV, "p", "He")
+    m = Pythia8(evt_kin, seed=1, cache=True)
+
+    config_lines = "\n".join(m._config)
+    assert "HeavyIon:SasdMpiReuseInit = 3" in config_lines
+    assert "HeavyIon:SigFitReuseInit = 3" in config_lines
+
+    cache_file_path = os.path.join(
+        _cache_base_dir(), "Pythia8", "cache_2212_1000020040_2000000.mpi"
```
afedynitch, line 214: Some of these generated files are very simple. It would be better to actually compare the contents of the sigfit file by comparing `open(fname,"r").read()` of the newly generated file with concrete content such as
```
1 2000 2000
18.3948 
1.90724 
0.357502 
```

This would allow to track (partially) if cache files need updating after Pythia version updates or so.

tests/test_pythia8.py
```diff
@@ -98,18 +100,25 @@ def test_mothers(event):
...
-    # and Pythia seldom raises the success flag unless Ecm > TeV are used.
-
-    event = run_collision(2000 * GeV, "p", (4, 2))
-    assert event.pid[0] == 2212
+@pytest.mark.parametrize("projectile", ("pi-", "K+", "p", "He"))
+def test_nuclear_collision(projectile):
+    # This test is very slow the first time when Pythia is computing
+    # some cached results. Afterwards it runs super fast.
+    ecm = 2000 * GeV
+    event = run_collision(ecm, projectile, "He")
```
afedynitch, line 108: Here the cache should be enabled using files distributed via the binary package in iamdata.

src/chromo/models/pythia8.py
```diff
@@ -137,6 +149,34 @@ def __init__(self, evt_kin, *, seed=None, config=None, banner=True):
             f"Random:seed = {self.seed % 900_000_000}",
         ]
 
+        if evt_kin is not None:
+            # production of cache files by Pythia8 to speed up subsequent runs
+            cache_prefix = str(
```
afedynitch, line 154: This should be under `if cache`, since ent_kin = None and `cache = True` would trigger an error or better `if cache and evt_kin is not None` because you need both to determine the path.

src/chromo/models/pythia8.py
afedynitch: Don't the cache files solve this problem? I saw a cross section fit file among the generated ones.

afedynitch (commented):
Only a few items to go. Please also pull from the latest master after #189 is merged.

HDembinski:
@afedynitch 

> I was not aware of the distribution of some default cache files within the binary package... in that case I agree that the natural place for these files to live is within the iamdata directory.

Sorry, I think I was not clear. Currently, we don't have cache files for the new mpi-stuff in the iamdata directory of Pythia8, but I think that's the place where we should put those cache that we want to use for our unit tests. We have to generate them and then add them to a new version of the [iamdata_v001.zip](https://github.com/impy-project/chromo/releases/tag/zipped_data_v1.0).

What we have in there currently is a Pythia8 directory with data files that Pythia8 needs, like the information about particles known to Pythia8, etc. These are not strictly cache files, but data files that Pythia8 needs. I think that the new cache files would nicely fit in there, as they are also "data".

This is analog to QGSJetII, where the corresponding iamdata directory essentially contains a huge binary cache file that QGSJetII can in principle re-create, but we never do that as it takes too long.

PR #189, opened 2023-11-09
Maintenance release

afedynitch:
The PR is a maintenance release with the following changes:

1. From version 1.26.2 we can build the f2py interface files natively with the public `numpy` release with no hacking. This has been reflected in the build chain.
2. A quantity "called `fw` that was need for producing histograms of the invariant cross section `E*dsigma/dp3` has been brought back to the event data. It requires the center of mass momentum `pcm` to be accessible from the `EventKinmatics` class (same as for `ecm`).
3. Repair event history and pre-/appending beam particle info can now be optionally disabled by setting `generator._restore_beam_and_history = False` to save some CPU time when it is not needed. By default it remains on.
4. When subclassing `MCRun`, for example when developing custom interaction models, the model library may not be called `chromo.models.library_name`. To get this flexibility, the `MCRun` base class is searching under the previous path `chromo.models` for library name and if it fails tries just `library_name`.
5. `__hash__` for EventKinematics such that instances can be dictionary keys.
6. Fixed bugs in QGSJET cross section function when it was called with nuclear or meson primary. Fixed major bug in QGSJET-II cross section readout. `qgsect` supports only 3 types of primaries and not all the known particles.
7. Cross section readout fixed for SIBYLL with meson primaries (where p.A is None)
8. Implemented missing cross section readout for SIBYLL 2.1
9. Some cosmetics and some imports sorted


src/chromo/util.py
```diff
@@ -733,3 +810,94 @@ def select_long_lived(tau=0, mm=False):
             long_lived.append(pid)
 
     return long_lived
+
+
+class redirect_std_streams:
```
HDembinski, line 815: 1) This is a brittle solution, because IPython and Jupyter notebooks also mess around with the streams. Have you tested that this works in IPython and in Jupyter notebooks?

2) Does this solution have confusing side-effects? Can it be asserted that only output from the Fortran code is captured? What if a Python process prints at the same time, e.g. when running multi-threaded?

3) Does this really work? Perhaps Fortran bypasses the Python stdout / stderr?

4) Where are the unit tests for this?

src/chromo/util.py
```diff
@@ -733,3 +810,94 @@ def select_long_lived(tau=0, mm=False):
...
+        file.parent.mkdir(parents=True, exist_ok=True)
+        return file
+
+    def __enter__(self):
+        """
+        Enter the context and redirect stdout and stderr
+        to the specified files or /dev/null.
+        """
+        if self.stdout_file:
+            if self.stdout_file == "null":
```
HDembinski, line 864: The magic string "null" is bad design, and it is not clearly undocumented. Please check for equality with `os.devnull` instead. `os.devnull` is also a string, but since it supported by `os`, it is at least part of the Python standard libs.

src/chromo/util.py
afedynitch: Indeed! It's super brittle :((( 

It only works with a model in some standard cases (actually also in Jupyter) but once you start printing from python it looses focus.

I had a Fortran-based mechanism in place to open and close files and maintain output units. That was removed, but in practice it worked quite nicely. So how can I logs, I need them for actual work and also from multiprocessing.

Shall I try reviving the Fortran file management?

HDembinski: I think the proper way is to patch the FORTRAN codes that they don't write to stdout but to a file that we specify. I looked into how that can be achieved, but I am not a FORTRAN expert. The Fortran file management, if I remember correctly, was not doing that, it was also a brittle solution similar to this one.

HDembinski: What I remember is that WRITE is a statement with a special syntax, so that we cannot just replace write statements with our own function calls. 

I think what we should do is to open our own files and then patch the WRITE statements so that they use our file handle instead of stdout. This patch can probably be applied automatically, so that we don't have to modify the source code in the repository. That would be my preferred solution.

HDembinski: If you need this urgently, the redirect_std_streams can be an intermediate solution. But then it should not be public so that we can remove it later (name starts with _).

In any case there need to be some unit tests. The PR title is misleading then. New features are not part of maintenance. And this is not a release PR, or is it?

afedynitch: I removed this functionality and will push it to user code. It doesn't really depend on chromo.

HDembinski: That's also a solution. To clarify, I think it is fine if you want to place this in chromo's utils, if you use it a lot yourself it might be useful for others as well. But it should be documented as a temporary solution that might disappear in the future.

src/chromo/kinematics.py
```diff
@@ -76,6 +76,7 @@ class EventKinematicsBase:
     p1: Union[PDGID, Tuple[int, int]]
     p2: Union[PDGID, Tuple[int, int], CompositeTarget]
     ecm: float  # for ions this is nucleon-nucleon collision system
+    pcm: float
```
HDembinski, line 79: For the record, carrying around redundant state is a maintenance burden, because it is difficult to guarantee that redundant state stays internally consistent. EventKinematics goes against this for the sake of having quick access to all needed variables without recomputation. Adding another redundant variable is therefore ok.

src/chromo/models/qgsjet.py
```diff
@@ -134,11 +136,18 @@ class QGSJet2Run(QGSJetRun):
 
     def _cross_section(self, kin=None):
         kin = self.kinematics if kin is None else kin
-
+        _projectile_id = {
```
HDembinski, line 139: It is not pythonic to start local variables with an underscore. These variables are not visible outside of the scope of the function anyway. We only use the underscore for attributes and methods, where it gets special treatment.

HDembinski:
I left some comments, but this PR is ready to go.

src/chromo/models/qgsjet.py
afedynitch: Cheers. I had removed the underscores in some other places following this idiom but forgot it here. Is now fixed.

PR #196, opened 2024-03-04
Towards release 0.5.0

jncots:
The goal of this PR is to add Python 3.12 and add release 0.4.1 to PyPI.


afedynitch:
Pls don't forget to bump the version to 0.4.2. otherwise there will be some mess with the tags.

Python 3.12 has moved to major release in October last year. However, pyhepmc has no 3.12 binaries. So, there might be an issue here. If so, @jncots please try to make a short PR to pyhepmc, or if @HDembinski could help with this small thing?

afedynitch (approved):
Good to go

PR #198, opened 2024-03-25
Generate cross section information in (py)hepmc files

afedynitch:
Add inelastic cross section to hepmc output.

Since chromo generates only successful events, the generated cross section is equal to the estimated one.


src/chromo/models/qgsjet.py
```diff
@@ -139,7 +139,7 @@ def _cross_section(self, kin=None, full_AA=False):
         """
         kin = self.kinematics if kin is None else kin
 
-        if full_AA and (kin.p1.is_nucleus and kin.p1.A > 1):
+        if max_info and (kin.p1.is_nucleus and kin.p1.A > 1):
```
jncots, line 142: Could target be anything (kin.p2)?

src/chromo/models/qgsjet.py
```diff
@@ -262,7 +262,7 @@ def _cross_section(self, kin=None, full_AA=False):
...
```
jncots, line 265: Could kin.p2 be anything?

jncots (commented):
Phew, that's a lot of changes for a small modification. I've noted some unclear or suspicious places.

src/chromo/models/qgsjet.py
afedynitch: Usually the possible targets are restricted by whatever is listed in `_targets`. So there is no explicit need to check this. This function is the correct one to call if the target is heavier than a nucleon or if the projectile is a nucleus. Save for the case below and for dpmjet above.

afedynitch: But yeah, actually, there is no check if an externally constructed kinematics object is (without calling `model.kinematics = kin`) is passed

src/chromo/common.py
```diff
@@ -196,6 +196,7 @@ class EventData:
     nevent: int
     impact_parameter: float
     n_wounded: Tuple[int, int]
+    production_cross_section: float
```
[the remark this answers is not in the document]
afedynitch: fixed

src/chromo/models/dpmjetIII.py
```diff
@@ -142,12 +142,11 @@ def __init__(self, evt_kin, *, seed=None):
 
         self._set_final_state_particles()
 
-    def _cross_section(self, kin=None, photon_x=0):
+    def _cross_section(self, kin=None, photon_x=0, max_info=False):
```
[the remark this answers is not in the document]
afedynitch: I'm not sure what photon_x means. I guess it is Q^2/s, but not 100% convinced. Anyhow, it's a private parameter and one can call it by calling the hidden `_cross_section` function.

src/chromo/models/dpmjetIII.py
```diff
@@ -184,6 +183,12 @@ def _generate():
                 + glxs.xsqe2[0, 0, 0]
                 + glxs.xsela[0, 0, 0],
             )
+        elif (kin.p1.is_nucleus and kin.p1.A > 1) or kin.p2.A > 1:
```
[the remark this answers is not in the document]
afedynitch: ok

src/chromo/models/qgsjet.py
afedynitch: yes. Everything from _targets

afedynitch:
Thanks! Fixes implemented

jncots (approved):
Ready to merge

PR #204, opened 2024-06-20
Lower ecm_min for EPOS-LHC

HDembinski:
As user Ashley pointed out, EPOS-LHC accepts sqrt(s) values down to 6 GeV, while we impose the limit at 10 GeV. This patch lowers our threshold to 6 GeV. I tried lowering it more, but then you get error messages from EPOS-LHC.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@ requires = [
     "wheel",
     "cmake",
     "numpy==1.26.2",
-    "charset_normalizer"
+    "charset_normalizer",
 ]
 #build-backend = "setuptools.build_meta"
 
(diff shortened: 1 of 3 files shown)


HDembinski:
This fails for an unrelated reason. In the latest versions of Numpy, VisibleDeprecationWarning was removed. We use this to warn about using deprecated API. Henry told me the proper warning to use is `FutureWarning`. I am going to replace the affected warnings to make the tests pass.

jncots:
@HDembinski, VisibleDeprecationWarning has been moved to numpy.exceptions.VisibleDeprecationWarning. This is how I tried to fix it in https://github.com/impy-project/chromo/pull/203. But your solution relies on core Python, so as I understand, is more general. 

Anyway, it is the least of problems introduced by numpy 2.0. Some problems would probably require f2py people again, because Epos LHC doesn't build for Windows.

HDembinski:
For now, we need to restrict the numpy version as a workaround for this problem. I am currently testing a change to setup.cfg to that effect. Please do not merge yet.

jncots (approved):
@HDembinski, everything builds,  tests pass. It can be merged. One last thing, should we set ecm_min for UrQMD as well?

HDembinski:
It is already set for UrQMD, no?

jncots:
You are right, _ecm_min is set for UrQMD. I confused it with some other model.

afedynitch (approved):
While technically using EPOS LHC is ok at these energies, it is not very recommended from the physics point of view. This is, however, not the task of chrome to check this.

Good to go!

PR #213, opened 2024-12-20
Fix pypi upload

jncots:
This is a small branch to fix problem with uploading of wheels on PyPI.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -200,16 +200,10 @@ jobs:
           merge-multiple: true
           path: dist
 
-      # - name: Publish package to PyPI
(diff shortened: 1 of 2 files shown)


afedynitch (approved):
Thanks

jncots:
The problem was with new version of `actions/upload-artifact@v4` that cannot change existing artifact. Each job in the matrix should have its own named artifact now. To download all artifacts to `dist` directory, we should omit "name" option in `actions/download-artifact`. To download all artifacts to the `dist` and not subfolders as

```bash
dist/
  artifact-1/
  artifact-2/
  .... 
```
we use `merge-multiple: true` .

PR #214, opened 2024-12-27
Integration of QGSJetIII-04

jncots:
PR integrates new model `QGSJetIII-04`


jncots:
Class `QGSJetIII04` is added. The compilation, loading and running are working now. 

Noticed problems:
- Table `chromo/iamdata/qgsjet/qgsjetIII.dat` is large (~1.4 G) and loaded quite slowly (~18 sec) compared to QGSJetII-04 (~4 sec) 
- It runs much slower than `QGSJetII04`: 13 sec for `QGSJetIII04` vs 1.8 sec for `QGSJetII04` for the code

```python
nevents = 0
average_pt = 0

# Generate 10000 events
for event in generator(1000):
    # Filter event
    event = event.final_state_charged()
    # do something with event.pid, event.eta, event.en, event.pt, etc.
    # these variables are numpy arrays, that can be histogrammed or counted like
    pt = event.pt[np.abs(event.pid) == 211]
    # The list could be empty
    if len(pt) > 0:
        nevents += 1
        average_pt += np.mean(pt)

average_pt = average_pt / nevents
print("Average pT for charged pions {0:4.3f}".format(average_pt))
```


Next step: fix tables for changed particle IDs. Probably it is the reason for slow running.

jncots:
The only change in `CQGSHH_HA_CS` is removing `call qggau1(gz1) `, because its code has been moved to `qggau`

afedynitch:
Thanks for starting this. 

Why is it called "04"? There is no number in the model's name. 

Particle ids are different:

'''
c the following notations for the particles types are used:
c 0 - pi0, 1 - pi+, -1 - pi-, 2 - p, -2 - pbar, 3 - n,
c -3 - nbar, 4 - K+, -4 - K-, 5 - Ks, -5 - Kl, 6 - lambda, -6 - lambdabar,
c 7 - delta++, -7 - delta++bar, 8 - delta-, -8 - delta-bar, 10 - eta,
c 16 - rho+, -16 - rho-, 17 - rho0, 18 - K*+, -18 - K*-, 19 - K*0, -19 - K*0bar,
c       20 - sigma+, -20 - sigma+bar, 21 - sigma-, -21 - sigma-bar
'''

afedynitch (approved):
Looks fine to me

PR #219, opened 2025-03-23
Update SIBYLL*

afedynitch:
Update SIBYLL* based on SIBYLL-2.3d to the latest version (p04).


afedynitch:
Hi @HaoZeke,

the update from numpy 2.2.3 to 2.2.5 broke f2py builds. For some reason the definition of multiple commons in 1 line doesn't work anymore. I'm wondering if this change is intentional or whether the tests have not covered thissuite.

Have a look at [this commit](https://github.com/impy-project/chromo/pull/219/commits/21d83711fbf8129c3660283fee7e155757c6de86) for the changes necessary to fix the build. The code I modified is not ours, so I would prefer avoiding this fix to the external source.

PR #221, opened 2025-04-17
Add EPOS LHC-R

afedynitch:
This PR adds EPOS LHC-R event generator to the list of supported models.

There are two flavors, the full (and slow) `class EposLHCR` and the faster `EposLHCRNoHadrRescattering`. 

It's up to debate if any whether the version without hadronic rescattering should become the baseline or the one with rescattering.


jncots (approved):
Everything is ready to merge.

PR #227, opened 2025-07-31
BUGFIX: SIBYLL does not raise warning for projectiles without cross section definition

StFroese:
Sibyll checks if the projectiles is part of the accepted projectiles list. If this is not the case it sets empty `CrossSectionData()`. This fails for some projectiles since they might be part of the projectiles list but not of the cross_section_projectiles where cs data is taken from later.

https://github.com/impy-project/chromo/blob/6f81a99b1dce37c1f04e63f723f3190c4aa60e39/src/chromo/models/sibyll.py#L203-L209

This is a simple fix by replacing `self._projectiles` with `self._cross_section_projectiles`.


afedynitch:
This fix needs to handle the case of p1 nuclei :)

PR #230, opened 2025-09-01
🖌️  Remove NaN prod CS from SIBYLL wrapper

StFroese:
For kaons-air sibyll returns nan production cross sections. Chromo is overwriting this, which should not happen. It should expose the nans of the generator to the user.

https://github.com/impy-project/chromo/blob/c2d88b4cc0dc843b19ff59596224aa635e419b66/src/chromo/models/sibyll.py#L236-L239

Small fix 😊

### DIFF
--- a/tests/test_cross_sections.py
+++ b/tests/test_cross_sections.py
@@ -37,7 +37,12 @@ def test_generator(projectile, target, Model):
         # cannot use Argon in SIBYLL, so make air from N, O only
         p2 = CompositeTarget((("N", 0.78), ("O", 0.22)))
 
-    kin = CenterOfMass(100 * GeV, p1, p2)
+    # increase energy for K+ and SIBYLL23c to avoid prod cs = nan region
+    # this is specifig to SIBYLL23C
+    if Model == im.Sibyll23c and projectile == "K+" and target == "air":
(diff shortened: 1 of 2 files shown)


tests/test_cross_sections.py
```diff
@@ -37,7 +37,12 @@ def test_generator(projectile, target, Model):
         # cannot use Argon in SIBYLL, so make air from N, O only
         p2 = CompositeTarget((("N", 0.78), ("O", 0.22)))
 
-    kin = CenterOfMass(100 * GeV, p1, p2)
+    # increase energy for K+ and SIBYLL23c to avoid prod cs = nan region
+    # this is specifig to SIBYLL23C
+    if Model == im.Sibyll23c and projectile == "K+" and target == "air":
+        kin = CenterOfMass(1e3 * GeV, p1, p2)
```
StFroese, line 43: needed to bump energy here since it otherwise returns `NaN` prod cs

StFroese:
@afedynitch can you review?

afedynitch (approved):
I agree with the changes. But since, strictly speaking, the physics changes you need to bump the minor version and add comments in the release notes and expand the PR text a bit.

PR #269, opened 2026-08-10
Fix the macOS wheel build on main, gated behind a fast pre-flight stage

afedynitch:
`Build` has never been green on `main` — it only triggers on `push`, so no PR
ever exercised it, and it drifted away from the (green) macOS setup in
`test.yml`. Two independent macOS failures in
[run 31360572063](https://github.com/impy-project/chromo/actions/runs/31360572063):

**`macos-15-intel`:** `setup-fortran@v1.6.1` runs
`ln -fs /opt/homebrew/bin/gfortran-13 /usr/local/bin/gfortran` for every
macOS ≥ 14, but Homebrew's prefix on Intel *is* `/usr/local`, so the symlink
dangles → `meson.build:1:0: ERROR: Unknown compiler(s): [['gfortran']]`.

**`macos-latest`:** that label is macOS 26 now, so the Homebrew gfortran dylibs
have `minos 16.0` while the workflow hardcoded
`MACOSX_DEPLOYMENT_TARGET=15.0` → `delocate-wheel` rejects the wheel with
`Library dependencies do not satisfy target MacOS version 15.0`.

## Changes

- Pin `fortran-lang/setup-fortran@v1.9.2` in `build.yml` and `test.yml`
  (was `@v1.6.1` / unpinned `@main`). v1.9.x derives the prefix via
  `brew --prefix`.
- Derive `MACOSX_DEPLOYMENT_TARGET` from the dylibs delocate actually bundles,
  in `scripts/ci/macos_wheel_preflight.sh`. Generalises the inline `vtool` step
  `test.yml` already had; both workflows now call it, so they can't drift again.
- Build arm64 release wheels on `macos-15`, not `macos-latest` — raising the
  target to 16.0 would make published wheels require macOS 26. `test.yml` stays
  on `macos-latest` with `EXPECTED_MACOS_TARGET=15.0` so it warns on the next
  image bump.
- `preflight` is its own stage that `wheels` and `sdist` depend on: 4 jobs,
  ~30 s, and it catches both failures above. It also runs on PRs touching
  `build.yml` or `scripts/ci/**`.
- `timeout-minutes` everywhere (150 for wheels; slowest successful job was
  85 min).

CI on this PR: preflight 4/4 in 23–34 s, `cibw macos-15-intel` →
`macosx_15_0_x86_64` and `cibw macos-latest` → `macosx_16_0_arm64`, both
delocated and tested clean.

https://claude.ai/code/session_014arqpGfDHtJek6cJXF2gd9

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,9 +4,7 @@ on:
   push:
     branches:
       - main
-  # Changes to the wheel build itself never reached a PR before, which is how
-  # this workflow drifted away from the (green) macOS setup in test.yml. Run the
-  # cheap pre-flight stage on such PRs so the drift is caught before the merge.
+  # Runs the preflight stage only; the wheel matrix stays push-only.
   pull_request:
     paths:
       - '.github/workflows/build.yml'
@@ -14,9 +12,9 @@ on:
   workflow_dispatch:
 
 jobs:
-  # Stage 1: seconds, no compilation. Catches a broken Fortran toolchain and a
-  # deployment target that the bundled gfortran runtime cannot satisfy -- the
-  # two things that have taken this workflow down after ~1.5 h of doomed matrix.
+  # Gate for the wheel matrix: no compilation, ~30 s. Fails on an unusable
+  # Fortran toolchain or a deployment target the bundled gfortran runtime does
+  # not satisfy, both of which cibuildwheel only reports an hour in.
   preflight:
     name: preflight ${{ matrix.os }}
     strategy:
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -59,17 +111,24 @@ jobs:
...
-      - name: Set macOS deployment target
-        if: runner.os == 'macOS'
-        run: echo "MACOSX_DEPLOYMENT_TARGET=15.0" >> $GITHUB_ENV
-
       - name: Setup gfortran
-        uses: awvwgk/setup-fortran@v1.6.1
+        # v1.6.1 symlinked /opt/homebrew/bin/gfortran-13 on every macOS >= 14,
+        # which dangles on the Intel runners (Homebrew prefix /usr/local) and
+        # made meson report "Unknown compiler(s): [['gfortran']]". v1.9.x
+        # derives the prefix with `brew --prefix`.
```
afedynitch: You don't need to report the history here, we won't go back to the old versions ever

.github/workflows/build.yml
```diff
@@ -4,17 +4,61 @@ on:
...
+  pull_request:
+    paths:
+      - '.github/workflows/build.yml'
+      - 'scripts/ci/**'
   workflow_dispatch:
 
 jobs:
+  # Stage 1: seconds, no compilation. Catches a broken Fortran toolchain and a
+  # deployment target that the bundled gfortran runtime cannot satisfy -- the
+  # two things that have taken this workflow down after ~1.5 h of doomed matrix.
```
afedynitch: what's this story about. Omit

.github/workflows/build.yml
afedynitch: Done in 324a22b — dropped the history, kept only the constraint (`>= v1.9` because earlier releases hardcode the `/opt/homebrew` prefix). Applied the same trim to the other comments in the diff and to the script header.

.github/workflows/build.yml
afedynitch: Done in 324a22b — now just `# Runs the preflight stage only; the wheel matrix stays push-only.` Applied the same trim to the rest of the diff and to the script header.

PR #275, opened 2026-08-21
Fix SIBYLL nucleon-projectile routing (neutron isospin)

afedynitch:
## Summary

- route protons and neutrons through sibyll() so their distinct production IDs are preserved
- reserve sibnuc() for real nuclei with A > 1
- restore the neutron isospin mirror behavior without changing real-nucleus handling

## Validation

- rebased cleanly onto current main
- git diff --check passed
- Python syntax compilation passed
- targeted SIBYLL tests could not collect locally because the temporary environment lacks the particle dependency; CI will run the project test matrix

### DIFF
--- a/src/chromo/models/sibyll.py
+++ b/src/chromo/models/sibyll.py
@@ -289,7 +289,11 @@ class SIBYLLRun(MCRun):
         self._glauber_trials = ntrials
 
     def _set_kinematics(self, kin):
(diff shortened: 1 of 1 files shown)


afedynitch:
Closing — superseded by #278 (merged), which contains the identical is_real_nucleus guard.

PR #232, opened 2025-09-09
External optional link to FLUKA 2025

afedynitch:
## FLUKA backend

Adds an optional, license-restricted FLUKA 2025.1 backend. FLUKA runs as the
event generator; chromo drives it through F2PY helpers in
`src/fortran/fluka/chromo_fluka.f` linked against the official archives
(`libflukahp.a`, `libdpmmvax.a`, `librqmdmvax.a`, `latestRQMD/librqmd.a`,
`interface/libdpmjet*.a`).

### Features

- `Fluka`: hN, hA, AA, photohadronic (γp), photonuclear (γA) and electromagnetic
  dissociation (EMD), with nuclear remnants in HEPEVT. Interaction channels are
  selected via `InteractionType` (IFLXYZ encoding).
- `FlukaDecay`: kinematics-free interface to FLUKA's decay tables — catalog
  query, isotope lookup, inclusive sampling, and recursive decay chains.
  Usable standalone or as `Fluka(post_event=DecayChainHandler(...).expand)`.
- `CrossSectionData` gains an `emd` field.

### Build & availability

Build from source only; not distributed on PyPI. `"fluka"` is in
`[tool.chromo] enabled-models`, but the meson block validates `$FLUPRO` and the
required archives and, when absent, warns and skips the `_fluka` extension while
the rest of chromo builds normally. `Fluka` always imports; constructing it on a
build without the extension raises `ModuleNotFoundError` with an install hint.

Setup: `scripts/install_fluka.sh` (see `AGENTS.md` for the full procedure).

### Limitations

- **Build from source only.** Never shipped as a wheel (license-restricted).
- **RNG is FLUKA's own.** `set_rng_state`/`get_rng_state` round-trip FLUKA's
  Ranmar state; the default `Pythia8DecayHandler` uses an independent,
  cross-process-unseeded RNG, so for bit-reproducible records disable it with
  `_activate_decay_handler(on=False)`.
- **`set_stable` is a no-op.** FLUKA's decay model is global and not
  runtime-configurable; short-lived particles are decayed by chromo's
  `Pythia8DecayHandler` by default.
- **Material cap of 10 targets** (FLUKA's compile-time `MEDFLK` bound). Default
  list keeps 9 slots, leaving 1 for a kinematics target; exceeding 10 raises
  `ValueError`.
- **Energy ceiling 300 TeV CMS** (`_ecm_max`).
- **Event generation:** no heavy-ion projectiles (A > 4) — EVTXYZ aborts with
[...]

PR #278, opened 2026-09-04
fix(sibyll): bare nucleons are not nuclei — A>1 guard for the sibnuc route

afedynitch:
SIBYLL's `sibnuc()` has no beam-isospin identity: routing bare nucleons through it silently simulates every nucleon beam (n, pbar, nbar) as a proton. `PDGID.is_nucleus` is True for A==1 species, so the existing check was insufficient.

Adds the same `A > 1` guard used by dpmjetIII/qgsjet/pythia8 in `_set_kinematics` and `_generate`.

Split out of #232.

### DIFF
--- a/src/chromo/models/sibyll.py
+++ b/src/chromo/models/sibyll.py
@@ -289,11 +289,9 @@ class SIBYLLRun(MCRun):
         self._glauber_trials = ntrials
 
     def _set_kinematics(self, kin):
-        # NB: PDGID.is_nucleus is True also for p/n/pbar/nbar (A == 1), and
-        # sibnuc() has no beam-isospin identity — routing bare nucleons
-        # through it silently simulates every nucleon beam as a proton.
-        # Same A > 1 guard as dpmjetIII/qgsjet/pythia8.
-        if not (kin.p1.is_nucleus and kin.p1.A > 1):
+        # sibnuc() has no beam-isospin identity — bare nucleons must not
+        # route through it (they would silently run as protons).
+        if not is_real_nucleus(kin.p1):
             self._production_id = self._lib.isib_pdg2pid(kin.p1)
             if self._production_id == 0:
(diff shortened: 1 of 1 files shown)


afedynitch (commented):
Should us is_real_nucleus

src/chromo/models/sibyll.py
```diff
@@ -289,7 +289,11 @@ def glauber_trials(self, ntrials):
         self._glauber_trials = ntrials
 
     def _set_kinematics(self, kin):
-        if not kin.p1.is_nucleus:
+        # NB: PDGID.is_nucleus is True also for p/n/pbar/nbar (A == 1), and
+        # sibnuc() has no beam-isospin identity — routing bare nucleons
+        # through it silently simulates every nucleon beam as a proton.
+        # Same A > 1 guard as dpmjetIII/qgsjet/pythia8.
+        if not (kin.p1.is_nucleus and kin.p1.A > 1):
```
afedynitch: Use is_real_nucleus

src/chromo/models/sibyll.py
```diff
@@ -315,7 +319,7 @@ def _set_stable(self, pdgid, stable):
 
     def _generate(self):
         kin = self.kinematics
-        if kin.p1.is_nucleus:
+        if kin.p1.is_nucleus and kin.p1.A > 1:
```
afedynitch: Same here

src/chromo/models/sibyll.py
afedynitch: Done — switched to `util.is_real_nucleus` in both places (src/chromo/models/sibyll.py:296). Pushed.

src/chromo/models/sibyll.py
afedynitch: Done — switched to `util.is_real_nucleus` in both places (src/chromo/models/sibyll.py:322). Pushed.

afedynitch (commented):
Done — switched to `util.is_real_nucleus`.

PR #283, opened 2026-09-09
fix(pythia6): grow PYJETS particle stack 4000 -> 12000 in legacy embedded copies

afedynitch:
## Problem

EPOS-LHC-R at the LHC energy frontier can produce events with more than 4000 final-state particles. When that happens, PYHEPC reaches the Pythia-6.4 error path

```fortran
IF(NHEP.GT.MSTU(4)) CALL PYERRM(8, '(PYHEPC:) no more space in /PYJETS/')
```

`PYERRM` with error type 8 terminates the calling program, so the entire host process is lost: no Python exception, no partial output, no cleanup. The energy scale at which it occurs depends on generator state, so the abort position within a run varies.

The HEPEVT increase in eea27ea (`nmxhep 9990 → 100000`) left this limitation in place: PYJETS is filled during hadronisation and reaches its limit before the HEPEVT stage is reached.

Reproduction (unpatched `main`):

```python
from chromo.kinematics import CenterOfMass
from chromo.models.epos import EposLHCR
for _ in EposLHCR(CenterOfMass(17783.0, 2212, 2212))(50000):
    pass
```

The process terminates during the first ~15000 events. With this change applied the same configuration produces 20000 events with no error output.

## Fix

Increase the `PYJETS` particle-stack limit to 12000 in the legacy embedded Pythia-6 copies, the value the FLUKA-flavoured DPMJET-III 19.1/19.3 trees already use (`include/pythia/inc/pyjets`: `PARAMETER (MAXLND=12000)`):

* `pythia6115dpm3v1.f`, `upyth.f`, `pythia6409.f`, `pythia-6.4.28.f`: the `COMMON/PYJETS` declarations and the 4th position of `DATA MSTU` (`MSTU(4)`).
* The `NMXHEP` declarations in the same compilation units, raised to the same value. PYHEPC copies `N=MIN(NHEP,MSTU(4))` particles into `/HEPEVT/`, so an `/HEPEVT/` view smaller than `MSTU(4)` turns a mid-sized event (between the view size and `MSTU(4)`) into an out-of-bounds write; and two translation units declaring the same common block with different extents have inconsistent layouts at link time. Both effects were observed locally while developing this change: raising `MSTU(4)` with `/HEPEVT/` views left at their previous size corrupted memory in every DPMJET-III 3.07 test, with the failure signature `GETPJE: IDX1.GT.IDX2`.
* `dpmjet3.0-7.f`: `MAXLND` (6 sites), `NMXHEP`, and the companion bookkeeping arrays of the same module, `IFLG(4000)`→12000 and `IHISMO(8000)`→24000.
* `phojet1.12-36c4.f` (39 `NMXHEP` declarations, `/POEVT1/`, `/POEVT2/`), `upmerge.f` glue: matching declarations in the same block.

Every declaration of a given common block within an extension module now uses the same extent, and `MSTU(4)` equals that extent. SOPHIA uses the Jetset-7.4 `/LUJETS/` block, which is separate and unchanged.

## Verification

* `_eposlhcr`, `_dpmjet307`, `_urqmd34`, `_pythia6` compile and link (meson/ninja, gcc-toolset-14).
* `EposLHCR` at 17.78 TeV pp completes 20000 events cleanly, where the unpatched build terminates around 15000.
* `DpmjetIII307` at 27.4 TeV pp generates events cleanly.
* For events below the previous limit the change alters only the abort threshold, so cross sections and particle compositions are unaffected.

## Version

`0.11.0 → 0.11.1` (bugfix).

🤖 Generated with [opencode](https://opencode.ai)


src/fortran/epos-lhc-r/sources/urqmd34/upyth.f
```diff
@@ -352,7 +352,7 @@ BLOCK DATA PYDATA
  
 C...PYDAT1, containing status codes and most parameters.
       DATA MSTU/
-     &   0,    0,    0, 4000,10000,  500, 8000,    0,    0,    2,
+     &   0,    0,    0, 12000,10000,  500, 8000,    0,    0,    2,
```
afedynitch, line 355: How do you know it belongs to pyjets? any proof?

src/fortran/dpmjet3.0/sources/pythia6115dpm3v1.f
```diff
@@ -247,7 +247,7 @@ BLOCK DATA PYDATA
 
 C...PYDAT1, containing status codes and most parameters.
       DATA MSTU/
-     &   0,    0,    0, 4000,10000,  500, 4000,    0,    0,    2,
+     &   0,    0,    0, 12000,10000,  500, 4000,    0,    0,    2,
```
afedynitch, line 250: Any proof this is a pyjets value?

src/fortran/epos-lhc-r/sources/urqmd34/upyth.f
afedynitch: Yes. The fourth entry of the DATA MSTU initializer is MSTU(4), and in the Pythia 6.4 source MSTU(4) is the number of particles that fit in /PYJETS/. Three independent uses in this same file, unchanged by this PR: (1) upyth.f:2622 in PYHEPC: IF(NHEP.GT.MSTU(4)) CALL PYERRM(8,'(PYHEPC:) no more space in /PYJETS/') — the abort that motivated the PR, and the message names the block; (2) upyth.f:12690 in PYEVOL: IF(N.GT.MSTU(4)-MSTU(32)-10) CALL PYERRM(11,'(PYEVOL:) no more memory left in PYJETS'); (3) upyth.f:2065 in PYINIT, which rejects MSTU(4) below 10 or above 900000, i.e. treats it as a settable PYJETS size. The matching COMMON declaration is bumped to the same extent in this PR, so parameter and array stay consistent. The DATA MSTU statement is the Pythia 6.4 default initialization of the PYDAT1 block (declared at upyth.f:322), so the value is set where the upstream distribution declares it.

src/fortran/dpmjet3.0/sources/pythia6115dpm3v1.f
afedynitch: Same evidence chain, one file down. In pythia6115dpm3v1.f the DATA MSTU is the default initialization of PYDAT1, and the checks that make MSTU(4) the PYJETS bound are in the file: PYHEPC aborts with '(PYHEPC:) no more space in /PYJETS/' when NHEP exceeds MSTU(4) (line ~2036), and PYEVOL emits '(PYEVOL:) no more memory left in PYJETS' at IF(N.GT.MSTU(4)-MSTU(32)-10) (line ~12689). The 7th position (4000 here, 8000 in the other copies) is MSTU(7), the saved-decay table limit, and is left as found.

PR #296, opened 2026-09-15
feat(#157): native printout interface for SIBYLL/Pythia6/8 (promoted to MCRun)

### LINKED ISSUE #157: Printout functions interface
All event generators have some sort of printout functions for an event, e.g. `dt_evtout` in DpmJet or `sib_list` in Sibyll. Since `EventData` class may contain transformed momenta and energy, it would be good to have a direct access to raw data produced by event generator for debugging. 

I have in mind to write a method for "MCRun" which would return the output of the printout functions as a string. Here I would like to discuss whether this functionality is needed and in what form.

A related issue: probably `EventData` class also needs a listing similar to  `dt_evtout` output?

afedynitch:
Closes #157

Adds a native event printout interface (`print_native_event`) for SIBYLL, Pythia6 and Pythia8, following the API already used by DPMJET/Phojet/EPOS, and promotes the generic mechanism to `MCRun`.

- `MCRun.print_native_event()` — default implementation raising `NotImplementedError` with the model name; docstring documents the mechanism. Existing DPMJET/Phojet/EPOS overrides are untouched.
- **SIBYLL** (`SIBYLLRun`): calls the native `SIB_LIST` routine. SIBYLL 2.1 takes no argument while 2.3 takes the LUN, so a `try/except TypeError` covers both without build-system changes.
- **Pythia6**: calls native `PYLIST(mode)` (mode 1–4 = summary/standard/vertices/no-momenta); output unit set to stdout by chromo.
- **Pythia8**: new `printEvent` binding in `_pythia8.cpp` wrapping `Event::list()` with `scoped_ostream_redirect` (same pattern as the existing `Settings.listChanged/listAll`); Pythia 8.317 has no `Pythia::printEvent` member anymore, `Event::list` is its mechanism.
- Tests: per-model tests assert the native banner appears (`SIBYLL EVENT SUMMARY` / `Event record` for 2.1 / `Event listing` / `PYTHIA Event Listing`), plus a `test_common` test for the `NotImplementedError` default. Fortran/C++ printouts are fd-level and fully buffered off a tty, so a `capture_native_printout` helper in `tests/util.py` runs model+printout in a fresh subprocess and captures stdout at exit — this also respects the once-per-process model initialization constraint.
- `pyproject.toml`: drops the duplicated `sib23c01` entry.

Printout goes to stdout (like the existing DPMJET/EPOS/Phojet implementations) rather than being returned as a string, since the backend writes on the Fortran/C++ level.

Opus-reviewed: CHANGES-applied (Sibyll21 banner assertion, unused test-helper kwarg removed). Relevant suites: 506 passed / 25 skipped / 3 xfailed; pre-commit green.


afedynitch (commented):
What about QGSJet and EPOS models? UrQMD?

